PR #9, opened 2023-04-01
feat: add oracle library

marktoda:
This commit adds the oracle library from core-next and tests


piesrtasty:
I see this PR was merged, however I'm not seeing the Oracle lib on the main branch. Was it removed?

PR #22, opened 2023-06-12
Full range hook

emmaguo13:


contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,492 @@
...
+        IPoolManager.ModifyPositionParams params;
+        bool rebalance;
+    }
+
+    struct HookPosition {
+        // the nonce for permits
+        uint96 nonce;
+        // the address that is approved for spending this token
+        address operator;
+        PoolId poolId;
```
ewilz, line 57: if `poolId` is the key in the mapping, it shouldn't need to be in the struct?

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,492 @@
...
+    struct CallbackData {
+        address sender;
+        IPoolManager.PoolKey key;
+        IPoolManager.ModifyPositionParams params;
+        bool rebalance;
+    }
+
+    struct HookPosition {
+        // the nonce for permits
+        uint96 nonce;
```
ewilz, line 54: this was for ERC-721 functionality of liquidity tokens, which isn't relevant anymore..

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,492 @@
...
+        // the address that is approved for spending this token
+        address operator;
+        PoolId poolId;
+        // the tick range of the position
+        int24 tickLower;
```
ewilz, line 59: might need later, but rn for MVP it's hardcoded?

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,492 @@
...
+    function beforeModifyPosition(
+        address sender,
+        IPoolManager.PoolKey calldata key,
+        IPoolManager.ModifyPositionParams calldata params
+    ) external override returns (bytes4) {
+        // check msg.sender
+        require(sender == address(this), "sender must be hook");
+
+        // check full range
+        require(
```
ewilz, line 475: if we're checking that the sender is address this, isn't verifying min/max tick redundant?

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,492 @@
...
+                            feeGrowthInside0LastX128 - position.feeGrowthInside0LastX128,
+                            position.liquidity,
+                            FixedPoint128.Q128
+                        )
+                    );
+
+                uint256 balBefore = balanceOf(data.key.currency0, data.sender);
+                poolManager.take(data.key.currency0, data.sender, position.tokensOwed0);
+                uint256 balAfter = balanceOf(data.key.currency0, data.sender);
+                require(balAfter - balBefore == position.tokensOwed0);
```
ewilz, line 156: check seems redundant since take will most definitely return the the amount it's given? (unless it reverts)

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,492 @@
...
+                tickLower: MIN_TICK,
+                tickUpper: MAX_TICK,
+                liquidityDelta: -int256(int128(hookLiquidity))
+            });
+
+            BalanceDelta balanceDelta = hookModifyPosition(key, params);
+
+            (uint160 sqrtPriceX96,,,,,) = poolManager.getSlot0(key.toId());
+
+            uint160 sqrtRatioAX96 = TickMath.getSqrtRatioAtTick(MIN_TICK);
```
ewilz, line 433: if this doesn't get used more than once, saves gas not to define it in memory (just stick it right in the params on line 438) having some gas tests will help see what's going on as you change things

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,469 @@
...
+        // how many uncollected tokens are owed to the position, as of the last computation
+        uint128 tokensOwed0;
+        uint128 tokensOwed1;
+    }
+
+    mapping(PoolId => address) public poolToERC20;
+    mapping(PoolId => HookPosition) public poolToHookPosition;
+
+    constructor(IPoolManager _poolManager) BaseHook(_poolManager) {
+        blockNumber = block.number; // TODO: potentially set this on before initialization?
```
ewilz, line 70: yes, `block.number` will have to be per pool

lib/forge-std
ewilz, line 1: any reason for updating all these libraries??? If it's not crucial, I'd just leave it out of this PR and do a big library upgrade in isolation for cleanliness.

test/FullRange.t.sol
ewilz, line 1: I think we should definitely get some gas testing going. It's extremely helpful to have that all setup before beginning to refactor/cleanup.

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,469 @@
...
+        // update position
+        Position.Info memory posInfo = poolManager.getPosition(key.toId(), address(this), MIN_TICK, MAX_TICK);
+
+        position.feeGrowthInside0LastX128 = posInfo.feeGrowthInside0LastX128;
+        position.feeGrowthInside1LastX128 = posInfo.feeGrowthInside1LastX128;
+        position.tokensOwed0 = 0;
+        position.tokensOwed1 = 0;
+    }
+
+    function beforeModifyPosition(
```
ewilz, line 427: nit, I'd stick these towards the top to follow the conventional organization of the rest of the hooks in the repo

contracts/hooks/FullRange.sol
ewilz: might be worth a test testing that two diff pools get updated on first pool action of the respective pool. having one block.number for all pools would make that test fail.

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    function _rebalance(IPoolManager.PoolKey calldata key) internal {
+        console.log("we're rebalancing");
+        // retrieve all liquidity -- i think i should already have the liquidity kept track of
+        // uint128 hookLiquidity = poolManager.getLiquidity(key.toId(), address(this), MIN_TICK, MAX_TICK);
+
+        // position.liquidity = hookLiquidity;
+
+        HookPosition storage position = poolToHookPosition[key.toId()];
```
ewilz, line 386: I think you'll save gas by making this `memory`, then just writing it to storage when it's ready to be stored.

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,469 @@
...
+        //     params.tickLower == MIN_TICK && params.tickUpper == MAX_TICK, "Tick range out of range or not full range"
+        // );
+
+        if (block.number > blockNumber) {
+            console.log("block number increased");
+            blockNumber = block.number;
+            HookPosition storage position = poolToHookPosition[key.toId()];
+
+            if (position.tokensOwed1 > 0 || position.tokensOwed0 > 0) {
+                _rebalance(key);
```
ewilz, line 446: you can DRY up the code by putting _all_ the checks here in `_rebalance()` since they just repeat in `beforeSwap` 

so this would just be 
```
require(sender == address(this), "sender must be hook");
_rebalance(key);
```

contracts/hooks/Math.sol
ewilz, line 1: all generic libs can go in libraries

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,470 @@
...
+            IPoolManager.ModifyPositionParams({
+                tickLower: MIN_TICK,
+                tickUpper: MAX_TICK,
+                liquidityDelta: -int256(int128(position.liquidity))
+            })
+        );
+
+        (uint160 sqrtPriceX96,,,,,) = poolManager.getSlot0(key.toId());
+
+        uint128 liquidity = LiquidityAmounts.getLiquidityForAmounts(
```
aadams, line 380: i dont think you need to do this. liquidity for a full-range position should be min(y / sqrtPrice, x * sqrtPrice). you also can never be out of range in this position, so you dont need to pass sqrtPrice ratios. should save you some expensive calls.

this is because p_b and p_a are functionally infinity and 0, so the liquidity formulas resolve to the one above!

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,470 @@
...
+    ) external override returns (bytes4) {
+        // check msg.sender
+        require(sender == address(this), "sender must be hook");
+
+        if (block.number > blockNumber) {
+            blockNumber = block.number;
+            HookPosition storage position = poolToHookPosition[key.toId()];
+
+            if (position.tokensOwed1 > 0 || position.tokensOwed0 > 0) {
+                _rebalance(key);
```
aadams, line 420: my hot take is that you should rebase the erc20 instead of constantly rebasing the lp position. its far too expensive to rebalance it every single block. i think if you rebase the erc20 then apply the new position then there isnt any concerns (could be wrong).

some divergence will happen because of fee returns but if you rebase them on mint/burns it should be fine

ewilz: Yah, was talking to Hayden a bit and was thinking maybe it's not necessary to rebalance every block, (crucial on every burn tho). The one thing that rebasing the lp position does that the ERC20 doesn't, is that it auto-compounds the fees which was a really nice V2 benefit.

contracts/hooks/FullRange.sol
emmaguo13: ok hmm I did this because the min tick and max tick in this hook are actually a little smaller than the actually set min tick and max tick borders for all v4 pools, since we're using a tick spacing of 60. so I thought it might not be completely exact to the v2 formula

contracts/hooks/FullRange.sol
emmaguo13: I guess rebasing on every burn will just means the LP will get less benefit from auto compounding. like if they mint a position, and then burn before any other LP burns their position, then they will not get any autocompounding from their position, whereas having it done every block kind of guarantees that you will get the autocompounding benefits as long as u keep you position in for more than a block, but maybe it is not worth the gas cost

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,430 @@
...
+        );
+        poolInfo.tokensOwed1 += uint128(
+            FullMath.mulDiv(
+                posInfo.feeGrowthInside1LastX128 - poolInfo.feeGrowthInside1LastX128,
+                poolInfo.liquidity,
+                FixedPoint128.Q128
+            )
+        );
+
+        poolInfo.feeGrowthInside0LastX128 = posInfo.feeGrowthInside0LastX128;
```
ewilz, line 250: maybe we discussed,  but, why are we recording this? In v3-periphery we need to keep track for each individual, but if this sits in the pool since it's at the pool level, I think it's just redundant to keep it here.

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,430 @@
...
+            CurrencyLibrary.NATIVE.transfer(msg.sender, ethBalance);
+        }
+    }
+
+    function hookModifyPosition(IPoolManager.PoolKey memory key, IPoolManager.ModifyPositionParams memory params)
+        internal
+        returns (BalanceDelta delta)
+    {
+        IERC20Minimal(Currency.unwrap(key.currency0)).approve(address(this), type(uint256).max);
+        IERC20Minimal(Currency.unwrap(key.currency1)).approve(address(this), type(uint256).max);
```
ewilz, line 95: what are these two lines doing?? approving the tokens to itself? I'm confused lol...

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,430 @@
...
+
+    function lockAcquired(uint256, bytes calldata rawData) external override returns (bytes memory) {
+        require(msg.sender == address(poolManager));
+
+        CallbackData memory data = abi.decode(rawData, (CallbackData));
+
+        BalanceDelta delta = poolManager.modifyPosition(data.key, data.params);
+
+        // check if we are inputting liquidity for token0
+        if (delta.amount0() > 0) {
```
ewilz, line 113: I think instead of having this big if-statement, we have two functions (or maybe just two paths here) one for addingLiquidity, and one for removingLiquidity. We do not need to know from here whether it is a rebasing task, (note: rebasing logic is exactly the same as non-rebasing logic (i.e. lines 126/132).

If `data.params.liquidityDelta` is positive, we know all deltas are the same and they can fit under one it-statement, likewise for negative, and we don't need any conditionals for rebalancing.

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,430 @@
...
+                // since we need to reinvest into the pool
+                // after reinvestment, we should set the tokens owed to 0
+            } else {
+                poolManager.take(data.key.currency0, data.sender, uint256(uint128(-delta.amount0())));
+
+                // NOTE: commented out because we are never adding the liquidity being taken out to the tokensOwed.
+                // position.tokensOwed0 -= delta.amount0();
+            }
+
+            if (data.key.currency0.isNative()) {
```
ewilz, line 138: logic starting at lines 138 and 167 should just be a function that takes a currency and an amount to DRY code. (kind of like balanceOf)

contracts/hooks/UniswapV4ERC20.sol
```diff
@@ -0,0 +1,21 @@
...
+
+    bytes32 public constant PERMIT_TYPEHASH = 0x6e71edae12b1b97f4d1f60370fef10105fa2faae0126114a169c64845d6126c9;
+
+    constructor() public ERC20Permit("V4ERC20") ERC20("V4ERC20", "V4ERC20") Ownable(msg.sender) {}
+
+    function mint(address account, uint256 amount) external onlyOwner {
+        _mint(account, amount);
+    }
+
+    function burn(address account, uint256 amount) external {
```
ewilz, line 18: This should be onlyOwner too, right?

contracts/hooks/FullRange.sol
ewilz: also why are we settling down here. if the delta is negative, we're taking....not sure why we have to settle too?

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,430 @@
...
+    function hookModifyPosition(IPoolManager.PoolKey memory key, IPoolManager.ModifyPositionParams memory params)
+        internal
+        returns (BalanceDelta delta)
+    {
+        IERC20Minimal(Currency.unwrap(key.currency0)).approve(address(this), type(uint256).max);
+        IERC20Minimal(Currency.unwrap(key.currency1)).approve(address(this), type(uint256).max);
+
+        delta = abi.decode(poolManager.lock(abi.encode(CallbackData(address(this), key, params, true))), (BalanceDelta));
+
+        uint256 ethBalance = address(this).balance;
```
ewilz, line 99: why are we sending ETH to msg.sender on pool rebalances?? Thinking `hookModifyPosition` and `modifyPosition` are too short to merit their own functions....?

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,430 @@
...
+        uint256 feeGrowthInside0LastX128;
+        uint256 feeGrowthInside1LastX128;
+        // how many uncollected tokens are owed to the position, as of the last computation
+        uint128 tokensOwed0;
+        uint128 tokensOwed1;
+        uint256 blockNumber;
+        address liquidityToken;
+    }
+
+    mapping(PoolId => PoolInfo) public poolToInfo;
```
ewilz, line 59: just calling this `poolInfo` is more conventional with how we name params in the org

test/FullRange.t.sol
```diff
@@ -0,0 +1,873 @@
...
+        (, bytes32[] memory writes) = vm.accesses(address(impl));
+        vm.etch(address(fullRange), address(impl).code);
+        // for each storage key that was written during the hook implementation, copy the value over
+        unchecked {
+            for (uint256 i = 0; i < writes.length; i++) {
+                bytes32 slot = writes[i];
+                vm.store(address(fullRange), slot, vm.load(address(impl), slot));
+            }
+        }
+        key = IPoolManager.PoolKey(
```
ewilz, line 92: I think I had mentioned maybe starting a test with no fee for easiness, but I'm not sure this merits so much testing?? Feels like an edgecase

.forge-snapshots/swap with fee and rebalance.snap
```diff
@@ -0,0 +1 @@
+187820
```
ewilz, line 1: hold up, this is only 1872 more gas than `swap with fee`. Does this mean rebalancing only costs 1872 rn??? If so that's great......... But I feel like too good to be true

test/FullRange.t.sol
```diff
@@ -0,0 +1,873 @@
...
+    }
+
+    function testInitialAddLiquidityWithFeeSucceeds() public {
+        manager.initialize(feeKey, SQRT_RATIO_1_1);
+
+        uint256 currBalance0 = TestERC20(token0).balanceOf(address(this));
+        uint256 currBalance1 = TestERC20(token1).balanceOf(address(this));
+
+        snapStart("add liquidity with fee");
+        fullRange.addLiquidity(address(token0), address(token1), 3000, 100, 100, address(this), 12329839823);
```
ewilz, line 172: lol where did `12329839823` come from? maybe worth considering just having a constant called MAX_DEADLINE or something

test/FullRange.t.sol
```diff
@@ -0,0 +1,873 @@
...
+        // TODO: make sure 0 is correct
+        assertEq(feeGrowthInside0LastX128, 0);
+        assertEq(feeGrowthInside1LastX128, 0);
+        assertEq(tokensOwed0, 0);
+        assertEq(tokensOwed1, 0);
+    }
+
+    function testAddLiquidityFailsIfNoPool() public {
+        // PoolNotInitialized()
+        vm.expectRevert(0x486aa307);
```
ewilz, line 202: should be able to have something like `FullRange. PoolNotInitialized.selector`

test/FullRange.t.sol
```diff
@@ -0,0 +1,873 @@
...
+        fullRange.addLiquidity(address(token0), address(token1), 0, 100, 100, address(this), 12329839823);
+    }
+
+    function testAddLiquiditySucceedsWithNoFee() public {
```
ewilz, line 206: as mentioned above, thinking these `WithNoFee` and `WithFee` tests are redundant. Just test WithFee (not needed in test case) and only test noFee for when it's necessary to make sure this edgecase works in certain scenarios. Unless you have a reason for testing it so much already??

test/FullRange.t.sol
```diff
@@ -0,0 +1,873 @@
...
+        assertEq(TestERC20(token1).balanceOf(address(this)), currBalance1 - 125);
+
+        assertEq(UniswapV4ERC20(liquidityToken).balanceOf(address(this)), 125);
+    }
+
+    function testAddLiquidityWithDiffRatiosAndFee() public {
+        manager.initialize(feeKey, SQRT_RATIO_1_1);
+
+        uint256 currBalance0 = TestERC20(token0).balanceOf(address(this));
+        uint256 currBalance1 = TestERC20(token1).balanceOf(address(this));
```
ewilz, line 287: probably more aptly named `previousBalance` since by the time we're referencing it, it's not current.

test/FullRange.t.sol
```diff
@@ -0,0 +1,873 @@
...
+    }
+
+    function testSwapAddLiquiditySucceedsWithFeeNoRebalance() public {
+        manager.initialize(feeKey, SQRT_RATIO_1_1);
+
+        uint256 currBalance0 = TestERC20(token0).balanceOf(address(this));
+        uint256 currBalance1 = TestERC20(token1).balanceOf(address(this));
+
+        fullRange.addLiquidity(
+            address(token0), address(token1), 3000, 1000000000000000000, 1000000000000000000, address(this), 12329839823
```
ewilz, line 370: `1000000000000000000` --> `1 ether`

test/FullRange.t.sol
```diff
@@ -0,0 +1,873 @@
...
+
+        (,,,,,, address liquidityToken) = fullRange.poolToInfo(feeId);
+
+        assertEq(UniswapV4ERC20(liquidityToken).balanceOf(address(this)), 1000000000000000000);
+        assertEq(TestERC20(token0).balanceOf(address(this)), currBalance0 - 1000000000000000000);
+        assertEq(TestERC20(token0).balanceOf(address(this)), currBalance0 - 1000000000000000000);
+
+        // only get 98 back because of fees
+        vm.expectEmit(true, true, true, true);
+        emit Swap(feeId, address(swapRouter), 100, -98, 79228162514264329749955861424, 1000000000000000000, -1, 3000); // TODO: modify this emit
```
ewilz, line 381: probably want to be swapping with larger swapAmount

test/FullRange.t.sol
```diff
@@ -0,0 +1,873 @@
...
+
+        snapStart("swap with fee");
+        swapRouter.swap(
+            feeKey,
+            IPoolManager.SwapParams({zeroForOne: true, amountSpecified: 100, sqrtPriceLimitX96: SQRT_RATIO_1_2}),
```
ewilz, line 386: probably want to be swapping with larger swapAmount

.forge-snapshots/swap with fee and rebalance.snap
emmaguo13: no, the rebalancing is happening after the swap in this test (its happening before a modifyPosition). i can change the name of this snap haha

contracts/hooks/FullRange.sol
emmaguo13: I get an allowance insufficient error from the TestERC20 contract if I don't have these two lines... not sure why

contracts/hooks/FullRange.sol
emmaguo13: addressed in new commit, created functions that take in key and delta and sender address!

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,407 @@
...
+        return FullRange.beforeModifyPosition.selector;
+    }
+
+    function beforeSwap(address, PoolKey calldata key, IPoolManager.SwapParams calldata)
+        external
+        override
+        returns (bytes4)
+    {
+        // TODO: maybe don't sload the entire struct
+        PoolInfo storage position = poolInfo[key.toId()];
```
ewilz, line 132: this is unused

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,407 @@
...
+        _rebalance(key);
+        return IHooks.beforeSwap.selector;
+    }
+
+    function balanceOf(Currency currency, address user) internal view returns (uint256) {
```
ewilz, line 137: nit: organization, this function is kind of "random" (and should probably be in CurrencyLib), but maybe put it at the bottom instead of in the middle of relevant functions (the external hooks)

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,407 @@
...
+        } else {
+            return IERC20Minimal(Currency.unwrap(currency)).balanceOf(user);
+        }
+    }
+
+    function modifyPosition(PoolKey memory key, IPoolManager.ModifyPositionParams memory params)
+        internal
+        returns (BalanceDelta delta)
+    {
+        // msg.sender is the test contract (aka whoever called addLiquidity/removeLiquidity)
```
ewilz, line 149: nit: haha in prod it's not the test contract, I think this is self explanatory and comment not needed

contracts/hooks/FullRange.sol
ewilz: hmm if there's code in this contract where it's sending funds to itself, that might be a code smell...

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,407 @@
...
+
+    /// @notice Thrown when trying to interact with a non-initialized pool
+    error PoolNotInitialized();
+
+    /// @dev Min tick for full range with tick spacing of 60
+    int24 internal constant MIN_TICK = -887220;
+    /// @dev Max tick for full range with tick spacing of 60
+    int24 internal constant MAX_TICK = -MIN_TICK;
+
+    uint256 public constant MINIMUM_LIQUIDITY = 10 ** 3;
```
ewilz, line 42: looks like we haven't integrated this yet

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,407 @@
...
+    int256 internal constant MAX_INT = 2 ^ 256 / 2 - 1;
```
ewilz, line 43: `type(int256).max`?

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,407 @@
...
+        }
+    }
+
+    function _removingLiquidity(address sender, PoolKey memory key, BalanceDelta delta) internal {
+        poolManager.take(key.currency0, sender, uint256(uint128(-delta.amount0())));
+        poolManager.take(key.currency1, sender, uint256(uint128(-delta.amount1())));
+    }
+
+    function lockAcquired(bytes calldata rawData) external override(ILockCallback, BaseHook) returns (bytes memory) {
+        require(msg.sender == address(poolManager));
```
ewilz, line 189: should take advantage of modifier in `BaseHook.sol`

contracts/hooks/FullRange.sol
```diff
@@ -140,85 +142,61 @@ contract FullRange is BaseHook {
...
+        IERC20Minimal(Currency.unwrap(key.currency0)).approve(address(this), type(uint256).max);
+        IERC20Minimal(Currency.unwrap(key.currency1)).approve(address(this), type(uint256).max);
 
         delta = abi.decode(poolManager.lock(abi.encode(CallbackData(address(this), key, params))), (BalanceDelta));
+    }
 
-        uint256 ethBalance = address(this).balance;
-        if (ethBalance > 0) {
-            CurrencyLibrary.NATIVE.transfer(msg.sender, ethBalance);
+    function _addingLiquidity(address sender, PoolKey memory key, BalanceDelta delta) internal {
```
ewilz, line 164: I think a better name for these functions would be `_settleDeltas` and `_takeDeltas` respectively.

contracts/hooks/FullRange.sol
```diff
@@ -0,0 +1,407 @@
...
+        internal
+        returns (BalanceDelta delta)
+    {
+        IERC20Minimal(Currency.unwrap(key.currency0)).approve(address(this), type(uint256).max);
+        IERC20Minimal(Currency.unwrap(key.currency1)).approve(address(this), type(uint256).max);
+
+        delta = abi.decode(poolManager.lock(abi.encode(CallbackData(address(this), key, params))), (BalanceDelta));
+    }
+
+    function _addingLiquidity(address sender, PoolKey memory key, BalanceDelta delta) internal {
```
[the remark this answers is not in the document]
ewilz: Im' going to recommend some changes to CurrencyLibrary though in core... to get transferFrom in there..

contracts/hooks/FullRange.sol
emmaguo13: dk why this was there i just deleted it!

contracts/hooks/FullRange.sol
emmaguo13: i don't think we need this constant anymore!

ewilz:
noticing we're not enforcing minimumLiquidity locked up permanently for each pool, which I think opens us up to the same bugs V2 had without it

test/FullRange.t.sol
```diff
@@ -0,0 +1,413 @@
...
+        token1 = new MockERC20("token1", "1", 18);
+        token2 = new MockERC20("token2", "2", 18);
+
+        token0.mint(address(this), 2 ** 128);
+        token1.mint(address(this), 2 ** 128);
+        token2.mint(address(this), 2 ** 128);
+
+        manager = new PoolManager(500000);
+
+        vm.record();
```
ewilz, line 88: why are we recording here?

test/FullRange.t.sol
```diff
@@ -0,0 +1,413 @@
...
+    }
+
+    function testInitialAddLiquiditySucceeds() public {
+        manager.initialize(key, SQRT_RATIO_1_1);
+
+        uint256 prevBalance0 = MockERC20(token0).balanceOf(address(this));
+        uint256 prevBalance1 = MockERC20(token1).balanceOf(address(this));
+
+        snapStart("add liquidity");
+        fullRange.addLiquidity(address(token0), address(token1), 3000, 10 ether, 10 ether, address(this), MAX_DEADLINE);
```
ewilz, line 142: if you're snapshotting gas, you'll want to cache token0/token1 in memory to save on those sloads!

test/FullRange.t.sol
```diff
@@ -0,0 +1,413 @@
...
+        snapEnd();
+
+        assertEq(MockERC20(token0).balanceOf(address(this)), prevBalance0 - 10 ether);
+        assertEq(MockERC20(token1).balanceOf(address(this)), prevBalance1 - 10 ether);
+
+        (, address liquidityToken) = fullRange.poolInfo(id);
```
ewilz, line 148: personal preference, but I'd gather all variables first (lines 148 + 152) above and then all assertions together underneath

```
```
        (bool owed,) = fullRange.poolInfo(id);
       (, address liquidityToken) = fullRange.poolInfo(id);

        assertEq(MockERC20(token0).balanceOf(address(this)), prevBalance0 - 10 ether);
        assertEq(MockERC20(token1).balanceOf(address(this)), prevBalance1 - 10 ether);
        assertEq(UniswapV4ERC20(liquidityToken).balanceOf(address(this)), 10 ether - 1000);
        assertEq(owed, false);

emmaguo13: sometimes i need to get a newly updated bool owed after i use the liquidityToken, so i can't put them together all the time, but i will put the owed variable before the asserts

ewilz: just saying that asserts are grouped together after all actions + getters (so nothing should change). sometimes a test might have multiple stages, but `assign beforeParams, action, assign afterParams, assertions` for each stage is a nice way to consistently organize 🤷

.forge-snapshots/add liquidity.snap
```diff
@@ -0,0 +1 @@
+466644
```
ewilz, line 1: maybe follow similar naming conventions in repo for gas tests. i.e. `FullRangeAddLiquidity`. (I Know it's diff than core, I guess were still figuring this out but might be nice to have periphery consistent)

contracts/hooks/examples/FullRange.sol
```diff
@@ -0,0 +1,326 @@
...
+    uint16 internal constant MINIMUM_LIQUIDITY = 10 ** 3;
+
+    struct CallbackData {
+        address sender;
+        PoolKey key;
+        IPoolManager.ModifyPositionParams params;
+    }
+
+    struct PoolInfo {
+        bool owed;
```
ewilz, line 54: `hasAccruedFees`?

contracts/hooks/examples/FullRange.sol
```diff
@@ -0,0 +1,326 @@
...
+        return FullRange.beforeModifyPosition.selector;
+    }
+
+    function beforeSwap(address, PoolKey calldata key, IPoolManager.SwapParams calldata)
+        external
+        override
+        returns (bytes4)
+    {
+        PoolId poolId = key.toId();
+        bool tokensOwed = poolInfo[poolId].owed;
```
ewilz, line 208: might save tiny gas just inlining `poolInfo[poolId].owed` on line 210 ?

contracts/hooks/examples/FullRange.sol
```diff
@@ -0,0 +1,326 @@
...
+
+        liquidity = LiquidityAmounts.getLiquidityForAmounts(
+            sqrtPriceX96,
+            TickMath.getSqrtRatioAtTick(MIN_TICK),
+            TickMath.getSqrtRatioAtTick(MAX_TICK),
+            amountADesired,
+            amountBDesired
+        );
+
+        // require(amountADesired > 1000 && amountBDesired > 1000, "Input amount does not minimum liquidity");
```
ewilz, line 117: leftover comment?

contracts/hooks/examples/FullRange.sol
```diff
@@ -0,0 +1,326 @@
...
+            beforeModifyPosition: true,
+            afterModifyPosition: false,
+            beforeSwap: true,
+            afterSwap: false,
+            beforeDonate: false,
+            afterDonate: false
+        });
+    }
+
+    function addLiquidity(
```
ewilz, line 80: this function might want to take `amountAMin`, `amountbMin` so users can be wary against slippage
https://github.com/Uniswap/v3-periphery/blob/main/contracts/NonfungiblePositionManager.sol#L151

ewilz:
you'll have some conflicts when you merge main, I think esp since tokenA/B must be in a specific order now, so might want to think about doing that!

contracts/hooks/examples/FullRange.sol
```diff
@@ -0,0 +1,326 @@
...
+        UniswapV4ERC20 erc20 = UniswapV4ERC20(poolInfo[poolId].liquidityToken);
+
+        erc20.burn(msg.sender, liquidity);
+
+        delta = modifyPosition(
+            key,
+            IPoolManager.ModifyPositionParams({
+                tickLower: MIN_TICK,
+                tickUpper: MAX_TICK,
+                liquidityDelta: -int256(liquidity)
```
ewilz, line 165: when we call the actual poolManager liquidity will have increased bc of accrued fees, right? So when we burn we need to get `(liquidity / totalSupply) * manager.getLiquidity(pool)` basically a % of all liquidity owned, but take that percentage out of the actual pool's liquidity.

ewilz: (maybe this is why liquidity is getting weird in some of your fuzz tests??)

test/FullRange.t.sol
```diff
@@ -0,0 +1,453 @@
...
+
+        assertEq(UniswapV4ERC20(liquidityToken).balanceOf(address(this)), 10 ether - 1000);
+        assertEq(owed, false);
+    }
+
+    function testInitialAddLiquidityFuzz(uint256 amount) public {
+        manager.initialize(key, SQRT_RATIO_1_1);
+
+        if (amount <= 1000) {
+            vm.expectRevert("Input amount does not minimum liquidity");
```
ewilz, line 162: custom error?

test/FullRange.t.sol
```diff
@@ -0,0 +1,453 @@
...
+            fullRange.addLiquidity(address(token0), address(token1), 3000, amount, amount, address(this), MAX_DEADLINE);
+        } else if (amount >= 10 ** 34) {
```
ewilz, line 164: magic number could use var name, not sure what this is

test/FullRange.t.sol
```diff
@@ -0,0 +1,453 @@
...
+            vm.expectRevert();
+            fullRange.addLiquidity(address(token0), address(token1), 3000, amount, amount, address(this), MAX_DEADLINE);
+        } else {
+            fullRange.addLiquidity(address(token0), address(token1), 3000, amount, amount, address(this), MAX_DEADLINE);
+
+            (bool owed,) = fullRange.poolInfo(id);
+            assertEq(owed, false);
```
ewilz, line 171: wouldn't this be false for nil pools? could check pool manager has the pool initialized, otherwise this just checks no revert happened (which I guess is fine)

test/FullRange.t.sol
```diff
@@ -0,0 +1,453 @@
...
+            id, address(swapRouter), 1 ether, -906610893880149131, 72045250990510446115798809072, 10 ether, -1901, 3000
+        );
+
+        IPoolManager.SwapParams memory params =
+            IPoolManager.SwapParams({zeroForOne: true, amountSpecified: 1 ether, sqrtPriceLimitX96: SQRT_RATIO_1_2});
+        PoolSwapTest.TestSettings memory settings =
+            PoolSwapTest.TestSettings({withdrawTokens: true, settleUsingTransfer: true});
+
+        snapStart("swap");
+        swapRouter.swap(key, params, settings);
```
ewilz, line 221: want to try to avoid referencing storage in any gas tests

test/FullRange.t.sol
```diff
@@ -0,0 +1,453 @@
...
+        snapEnd();
+
+        (bool owed,) = fullRange.poolInfo(id);
+
+        assertEq(MockERC20(token0).balanceOf(address(this)), prevBalance0 - 10 ether - 1 ether);
+        assertEq(MockERC20(token1).balanceOf(address(this)), prevBalance1 - 9093389106119850869);
+
```
ewilz, line 228: nit: unnecessary newline?

test/FullRange.t.sol
```diff
@@ -0,0 +1,453 @@
...
+        fullRange.addLiquidity(address(token0), address(token1), 3000, 10 ether, 10 ether, address(this), MAX_DEADLINE);
+
+        IPoolManager.SwapParams memory params =
+            IPoolManager.SwapParams({zeroForOne: true, amountSpecified: 1 ether, sqrtPriceLimitX96: SQRT_RATIO_1_2});
+        PoolSwapTest.TestSettings memory settings =
+            PoolSwapTest.TestSettings({withdrawTokens: true, settleUsingTransfer: true});
+
+        snapStart("swap first");
+        swapRouter.swap(key, params, settings);
```
ewilz, line 250: once again, a storage reference for gas... would go down the file and cache in memory when taking gas estimates!

test/FullRange.t.sol
```diff
@@ -0,0 +1,453 @@
...
+        uint256 prevBalance1 = MockERC20(token1).balanceOf(address(this));
+
+        address token0Addr = address(token0);
+        address token1Addr = address(token1);
+
+        fullRange.addLiquidity(token0Addr, token1Addr, 3000, 10 ether, 10 ether, address(this), MAX_DEADLINE);
+
+        (, address liquidityToken) = fullRange.poolInfo(id);
+
+        assertEq(UniswapV4ERC20(liquidityToken).balanceOf(address(this)), 10 ether - 1000);
```
ewilz, line 300: would constant the 1000 as something like `LOCKED_LIQUIDITY` or something

test/FullRange.t.sol
```diff
@@ -0,0 +1,453 @@
...
+
+        assertEq(MockERC20(token0).balanceOf(address(this)), prevBalance0 - 10 ether);
+        assertEq(MockERC20(token1).balanceOf(address(this)), prevBalance1 - 10 ether);
+
+        UniswapV4ERC20(liquidityToken).approve(address(fullRange), type(uint256).max);
+
+        snapStart("remove liquidity");
```
ewilz, line 307: fyi since remove liquidity is technically in the same transaction these are all dirty storage slots so unfortunately is likely downplaying actual gas...Im not sure how to get around this other than adding liquidity in some pool that's used specifically for remove liquidty gas tests 🥱. Not super important to address for now... did something similar here... https://github.com/Uniswap/v4-core/blob/main/test/foundry-tests/TickBitmap.t.sol#L19

test/FullRange.t.sol
```diff
@@ -0,0 +1,453 @@
...
+        manager.initialize(key, SQRT_RATIO_1_1);
+
+        (, address liquidityToken) = fullRange.poolInfo(id);
+        UniswapV4ERC20(liquidityToken).approve(address(fullRange), type(uint256).max);
+
+        vm.expectRevert(); // Insufficient balance error from ERC20 contract
+        fullRange.removeLiquidity(address(token0), address(token1), 3000, 10 ether, MAX_DEADLINE);
+    }
+
+    function testRemoveLiquiditySucceedsWithPartialAndFee() public {
```
ewilz, line 356: not sure I understand "withPartialAndFee"

ewilz: would think that "withFee" means they're taking out their liquidity + owed swap fees?

test/FullRange.t.sol
```diff
@@ -0,0 +1,453 @@
...
+
+        (bool owed,) = fullRange.poolInfo(id);
+
+        assertEq(UniswapV4ERC20(liquidityToken).balanceOf(address(this)), 5 ether - 1000);
+        assertEq(MockERC20(token0).balanceOf(address(this)), prevBalance0 - 5 ether - 1);
+        assertEq(MockERC20(token1).balanceOf(address(this)), prevBalance1 - 5 ether - 1);
+        assertEq(owed, false);
+    }
+
+    function testRemoveLiquidityWithDiffRatiosAndFee() public {
```
ewilz, line 383: this test seems to be testing AddLiquidityWithDiffRatios?? then just removing partial part of the fee??

test/FullRange.t.sol
ewilz, line 1: Would be really nice to have a test where multiple (3?) parties add liquidity, earn some significant fees through a few big swaps, and then each pull liquidity and each end up with the right proportion of principal/fees.

Would be good to test beforeSwap() sets owed accurately

test/FullRange.t.sol
```diff
@@ -0,0 +1,453 @@
...
+
+        IPoolManager.SwapParams memory params =
+            IPoolManager.SwapParams({zeroForOne: true, amountSpecified: 1 ether, sqrtPriceLimitX96: SQRT_RATIO_1_2});
+
+        PoolSwapTest.TestSettings memory testSettings =
+            PoolSwapTest.TestSettings({withdrawTokens: true, settleUsingTransfer: true});
+
+        swapRouter.swap(key, params, testSettings);
+
+        fullRange.addLiquidity(address(token0), address(token1), 3000, 5 ether, 5 ether, address(this), MAX_DEADLINE);
```
ewilz, line 432: this step seems kinda extra?? or does it serve a specific purpose

test/FullRange.t.sol
ewilz, line 1: I've been pushing this format of foundry test names 😈  for readability, adding contraact name (TWAMM) at begining for easy grab with `--mt` for all tests in one contract and in console tests will be ordered nicely if function name is always prioritized etc. not a blocker, just a shill

contracts/hooks/examples/FullRange.sol
```diff
@@ -0,0 +1,357 @@
...
+            IPoolManager.ModifyPositionParams({
+                tickLower: MIN_TICK,
+                tickUpper: MAX_TICK,
+                liquidityDelta: liquidity.toInt256()
+            })
+        );
+
+        if (poolLiquidity == 0) {
+            liquidity -= MINIMUM_LIQUIDITY;
+            UniswapV4ERC20(pool.liquidityToken).mint(address(0), MINIMUM_LIQUIDITY); // permanently lock the first MINIMUM_LIQUIDITY tokens
```
ewilz, line 143: nit: I feel like this comment is long enough to just have it's own line on L42

test/FullRange.t.sol
```diff
@@ -0,0 +1,767 @@
...
+                fullRange.removeLiquidity(
+                    FullRange.RemoveLiquidityParams(address(token0), address(token1), 3000, amount, MAX_DEADLINE)
+                );
+            }
+
+            assertTrue(manager.getLiquidity(id) >= LOCKED_LIQUIDITY);
+        }
+    }
+
+    function testFullRange_removeLiquidity_ThreeLPsRemovePrincipalAndFees() public {
```
ewilz, line 663: so for this test, can we test that all LP's remove their liquidity, and that the only liquidity left in the pool is the MINIMUM liquidity?? I think that's the big important thing, bc to me it still looks like we're not removing the extra fees in the hook function?

test/FullRange.t.sol
```diff
@@ -0,0 +1,767 @@
...
+        fullRange.removeLiquidity(
+            FullRange.RemoveLiquidityParams(
+                address(token0), address(token1), 3000, 300 ether - LOCKED_LIQUIDITY, MAX_DEADLINE
+            )
+        );
+        (hasAccruedFees,) = fullRange.poolInfo(id);
+
+        // PoolManager does not have any liquidity left over
+        assertTrue(manager.getLiquidity(id) >= LOCKED_LIQUIDITY);
+        assertTrue(manager.getLiquidity(id) < LOCKED_LIQUIDITY + 30);
```
ewilz, line 753: can we assign 30 to a parameter called dust? just to be more explicit here?

test/FullRange.t.sol
```diff
@@ -0,0 +1,767 @@
...
+        fullRange.addLiquidity(
+            FullRange.AddLiquidityParams(
+                address(token0),
+                address(token1),
+                3000,
+                100 ether,
+                100 ether,
+                99 ether,
+                99 ether,
+                address(this),
```
ewilz, line 726: ohhh i see this all ends up being one position that's taken out

test/FullRange.t.sol
```diff
@@ -0,0 +1,767 @@
...
+    function testFullRange_addLiquidity_InitialAddSucceeds() public {
+        manager.initialize(key, SQRT_RATIO_1_1);
+
+        uint256 prevBalance0 = MockERC20(token0).balanceOf(address(this));
+        uint256 prevBalance1 = MockERC20(token1).balanceOf(address(this));
+
+        address token0Addr = address(token0);
+        address token1Addr = address(token1);
+
+        snapStart("FullRangeAddLiquidity");
```
ewilz, line 160: can we name this `FullRangeAddInitialLiquidity` and have another gas test `FullRangeAddLiquidity` that adds to the keyWithLiq since that's a more accurate gas snapshot for average LPs?

contracts/hooks/examples/FullRange.sol
```diff
@@ -0,0 +1,358 @@
...
+        delta = modifyPosition(
+            key,
+            IPoolManager.ModifyPositionParams({
+                tickLower: MIN_TICK,
+                tickUpper: MAX_TICK,
+                liquidityDelta: -(params.liquidity.toInt256())
+            })
+        );
+
+        erc20.burn(msg.sender, params.liquidity);
```
ewilz, line 186: this is an interesting line. There's the mantra "checks, effects, interactions" to avoid reentrancy bugs by putting code in that order. burning the token I would say is an "effect" (even tho it's kind of an interaction with the token) bc we're changing the balance the LP is owed by this contract. I'm trying to think if there's a way they could reenter and remove liquidity again having the same token balance.

But I don't think there's a good place where the user can reenter, and we'd always hit this line multiple times and require all the liquidity so I don't think reentrancy matters. But it stood out since we're paying them what they're owed before we nil out the balance.

ewilz: either way tho, having this on line 176 might be preferable and it will cause an earlier revert if they try and burn more than they own

contracts/hooks/examples/FullRange.sol
```diff
@@ -0,0 +1,360 @@
...
+import {FullMath} from "@uniswap/v4-core/contracts/libraries/FullMath.sol";
+import {UniswapV4ERC20} from "../../libraries/UniswapV4ERC20.sol";
+import {FixedPoint96} from "@uniswap/v4-core/contracts/libraries/FixedPoint96.sol";
+import {FixedPointMathLib} from "solmate/utils/FixedPointMathLib.sol";
+import {IERC20Metadata} from "@openzeppelin/contracts/interfaces/IERC20Metadata.sol";
+import {Strings} from "@openzeppelin/contracts/utils/Strings.sol";
+
+import "../../libraries/LiquidityAmounts.sol";
+
+contract FullRange is BaseHook, ILockCallback {
```
ewilz, line 26: can we call this FullRangePool ?

contracts/hooks/examples/FullRange.sol
```diff
@@ -0,0 +1,360 @@
...
+
+        if (poolLiquidity == 0) {
+            // permanently lock the first MINIMUM_LIQUIDITY tokens
+            liquidity -= MINIMUM_LIQUIDITY;
+            UniswapV4ERC20(pool.liquidityToken).mint(address(0), MINIMUM_LIQUIDITY);
+        }
+
+        UniswapV4ERC20(pool.liquidityToken).mint(params.to, liquidity);
+
+        require(
```
ewilz, line 149: nit: seems more appropriate on line 140?

contracts/hooks/examples/FullRange.sol
```diff
@@ -0,0 +1,360 @@
...
+        return FullRange.beforeInitialize.selector;
+    }
+
+    function beforeModifyPosition(address sender, PoolKey calldata, IPoolManager.ModifyPositionParams calldata)
+        external
+        view
+        override
+        returns (bytes4)
+    {
+        require(sender == address(this), "Sender must be hook");
```
ewilz, line 218: custom error?

contracts/hooks/examples/FullRange.sol
```diff
@@ -0,0 +1,360 @@
...
+        uint256 liquidity;
+        uint256 deadline;
+    }
+
+    mapping(PoolId => PoolInfo) public poolInfo;
+
+    constructor(IPoolManager _poolManager) BaseHook(_poolManager) {}
+
+    modifier ensure(uint256 deadline) {
+        require(deadline >= block.timestamp, "Expired");
```
ewilz, line 81: custom error?

ewilz (approved):

contracts/hooks/examples/FullRange.sol
emmaguo13: the issue is, if it's the first LP, then i'll mint them the liquidity without subtracting the amount of liquidity they are supposed to lock

ewilz (approved):

PR #49, opened 2023-08-08
Optimize v2 hook

emmaguo13:


contracts/hooks/examples/FullRange.sol
```diff
@@ -47,13 +48,7 @@ contract FullRange is BaseHook, ILockCallback {
 
     struct PoolInfo {
         uint128 liquidity;
```
ewilz, line 50: is this liquidity param redundant with the erc20 `totalSupply`??

contracts/hooks/examples/FullRange.sol
```diff
@@ -298,67 +252,63 @@ contract FullRange is BaseHook, ILockCallback {
...
+        if (position.owed && paramsLiquidity < 0) {
+            position.owed = false;
+
+            BalanceDelta balanceDelta = poolManager.modifyPosition(
+                key,
+                IPoolManager.ModifyPositionParams({
+                    tickLower: MIN_TICK,
+                    tickUpper: MAX_TICK,
+                    liquidityDelta: -int256(int128(position.liquidity))
+                })
```
ewilz, line 267: confused here....should we be taking out the liquidity as stated by `poolManager.getLiquidity()` which gets out of sync with our internal position.liquidity here as fees get accumulated? If what I said was true...we woudln't be swapping against a 0 liquidity pool....

ewilz:
codes a lot cleaner!

contracts/hooks/examples/FullRange.sol
emmaguo13: ah it is! i am thinking of getting rid of this liquidity param anyways.

contracts/hooks/examples/FullRange.sol
emmaguo13: although it adds 300 gas to rebalancing, i assume syncing position.liquidity will add ~200 gas so maybe not a big deal

ewilz (approved):

PR #50, opened 2023-08-10
Update to core-next latest (currency order + arb hook data)

ewilz:
## Related Issue

We are going to merge FullRnage hook tomorrow, then get this merged with the latest core-next changes

## Description of changes


contracts/hooks/examples/FullRange.sol
```diff
@@ -37,6 +37,8 @@ contract FullRange is BaseHook, ILockCallback {
     error ExpiredPastDeadline();
     error TooMuchSlippage();
 
+    bytes internal constant ZERO_BYTES = new bytes(0);
```
hensha256, line 40: why not empty bytes instead of `bytes(0)`?

test/GeomeanOracle.t.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.19;
 
+import "forge-std/console.sol";
```
hensha256, line 4: 👀

hensha256 (dismissed):
left a couple of comments but lgtmmm

PR #58, opened 2023-09-01
Routing

ewilz:
Probably of interest to note that foundry gas snapshots are not accounting for storage clear refunds, so before we get transient storage in here, gas is incredibly exaggerated.

Could cut down on bytecode by removing `Single` options


contracts/V4Router.sol
```diff
@@ -0,0 +1,174 @@
...
+        zeroForOne = currencyIn == currency0;
+        poolKey = PoolKey(currency0, currency1, params.fee, params.tickSpacing, params.hooks);
+    }
+
+    function _payAndSettle(Currency currency, address msgSender, int128 settleAmount) private {
+        _pay(Currency.unwrap(currency), msgSender, address(poolManager), uint256(uint128(settleAmount)));
+        poolManager.settle(currency);
+    }
+
+    function _pay(address token, address payer, address recipient, uint256 amount) internal virtual;
```
ewilz, line 173: @hensha256 wanted to outsource this to UniversalRouter...wondering what you think of adding a generically named "pay" method which internals can call abstractly.... 🤔

contracts/V4Router.sol
```diff
@@ -0,0 +1,174 @@
...
+    modifier poolManagerOnly() {
+        if (msg.sender != address(poolManager)) revert NotPoolManager();
+        _;
+    }
+
+    constructor(IPoolManager _poolManager) {
+        poolManager = _poolManager;
+    }
+
+    function v4Swap(SwapType swapType, bytes memory params) internal {
```
hensha256, line 30: internal so I'd add a `_`

contracts/V4Router.sol
```diff
@@ -0,0 +1,174 @@
...
+            params.sqrtPriceLimitX96,
+            msgSender,
+            true,
+            true
+        );
+    }
+
+    function _swapExactInput(ExactInputParams memory params, address msgSender) private {
+        unchecked {
+            for (uint256 i = 0; i < params.path.length; i++) {
```
hensha256, line 66: probably worth caching `params.path.length` given its accessed multiple times on every loop

contracts/interfaces/IV4Router.sol
```diff
@@ -0,0 +1,76 @@
...
+    error TooMuchRequested();
+
+    struct SwapInfo {
+        SwapType swapType;
+        address msgSender;
+        bytes params;
+    }
+
+    struct PathKey {
+        Currency tradeCurrency;
```
hensha256, line 28: I dont like the name `tradeCurrency`, its taken me a while to understand it. I guess I would call it like `intermediateCurrency`? Open to other ideas. just `tradeCurrency` wasn't immediately apparent to me

contracts/interfaces/IV4Router.sol
```diff
@@ -0,0 +1,76 @@
...
+        IHooks hooks;
+    }
+
+    struct ExactInputSingleParams {
+        PoolKey poolKey;
+        bool zeroForOne;
+        address recipient;
+        uint128 amountIn;
+        uint128 amountOutMinimum;
+        uint160 sqrtPriceLimitX96;
```
hensha256, line 40: Why are both `sqrtPriceLimitX96` and `amountOutMinimum` needed? Dont they both achieve slippage protection?

hensha256: In fact I'm pretty sure `amountOutMinimum` is unused?

contracts/V4Router.sol
hensha256: So you mean like the UR would have a `pay` function with `override` in it? And then any of the contracts that it overrides can have an abstract `pay` function too?

hensha256: The `V3SwapRouter` inherits from `Permit2Payments` so it can then call `payOrPermit2Transfer` where it needs to do payments. So I guess we could change that to a `pay` function too and implement the permit2 functionality? 

I'd be interested to see how it looks in the UR repo given that currently the hierarchy is
`Permit2Payments` -> `V3SwapRouter` -> `Dispatcher`
and the `Dispatcher` doesnt have to know anything about handling payments

contracts/V4Router.sol
```diff
@@ -0,0 +1,174 @@
...
+            params.sqrtPriceLimitX96,
+            msgSender,
+            true,
+            true
+        );
+    }
+
+    function _swapExactOutput(ExactOutputParams memory params, address msgSender) private {
+        unchecked {
+            for (uint256 i = params.path.length; i > 0; i--) {
```
hensha256, line 102: ugh i dislike that this is from `n` to `1` not `n-1` to `0`... but I see the issue.
I guess it could be 
`for (uint256 i = params.path.length-1; i < params.path.length; i--) {`
but thats also ugly so i guess theres no good way 😂

contracts/interfaces/IV4Router.sol
ewilz: great name, yah I just threw tradeCurrency in as a placeholder. I was hoping for something short to avoid annoying linting, but ultimately intermediateCurrency is a good option

contracts/V4Router.sol
ewilz: dude yah, I actually searched all Uniswap for `i--` to see how ppl deal with this type of thing. And I found ANOTHER POC v4 router that Pote wrote ages ago and he did the same thing baha

contracts/interfaces/IV4Router.sol
ewilz: ohh it was actually on my TODO to see how UR/router-sdk deals with slippage. amountOutMinimum is typically used from our UI. The only time Ive used `sqrtPriceLimitX96` is with TWAMM which needs it (but not through a router).`sqrtPriceLimitX96` is less about slippage and more about halting a swap midway through. Noah mentioned he was glad TWAMM needed it bc they were on the fence about adding the feature in. Anyway it's there for people if they need it, but it's not used much...

contracts/V4Router.sol
snreynolds: I feel like this is typical for for loops iterating backwards through the path though.. like doesn't seem too weird to me. Is there a reason we force it to be backwards though? I guess compatibility with how other routers work but there is no reason we couldn't just encode the path the other way also

contracts/V4Router.sol
ewilz: haha 20 gas per hop

contracts/V4Router.sol
ewilz: yah we could leave it out of dispatcher, easiest case, have a V4SwapRouter contract in UR that inherits this and implements pay??

contracts/interfaces/IV4Router.sol
ewilz: I think it could be cool to keep it for others to use if they want (we wouldn't be using any of the swapSingle in UR anyway, which is the only place it's exposed) But if unused internal functions contribute to UR bytecode (can experiment with that) happy to take out

test/shared/implementation/V4RouterImplementation.sol
```diff
@@ -0,0 +1,18 @@
...
+pragma solidity ^0.8.19;
+
+import {V4Router} from "../../../contracts/V4Router.sol";
+import {IPoolManager} from "@uniswap/v4-core/contracts/interfaces/IPoolManager.sol";
+import {IERC20Minimal} from "@uniswap/v4-core/contracts/interfaces/external/IERC20Minimal.sol";
+
+contract V4RouterImplementation is V4Router {
+    constructor(IPoolManager _poolManager) V4Router(_poolManager) {}
+
+    function swap(SwapType swapType, bytes memory params) external {
```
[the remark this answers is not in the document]
ewilz: wonder if making this a virtual function on the actual router contract is too big of a security vulnerability? In UR, if we don't override it to do nothing, we have a potential reentrancy vulnerability that could swap on v4 with funds sitting in the router. If it's a vuln for us, could potentially be for other ppl inheriting into an aggregator-like contract.... 🤔

contracts/interfaces/IV4Router.sol
```diff
@@ -0,0 +1,70 @@
...
+import {BalanceDelta} from "@uniswap/v4-core/src/types/BalanceDelta.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+import {Currency, CurrencyLibrary} from "@uniswap/v4-core/src/types/Currency.sol";
+import {TickMath} from "@uniswap/v4-core/src/libraries/TickMath.sol";
+import {IHooks} from "@uniswap/v4-core/src/interfaces/IHooks.sol";
+import {PathKey} from "../libraries/PathKey.sol";
+
+/// @title UniswapV4Routing
+/// @notice Abstract contract that contains all internal logic needed for routing through Uniswap V4 pools
+interface IV4Router is IUnlockCallback {
```
hensha256, line 16: theres no functions in this interface... feels like maybe it shouldnt be an interface?

contracts/V4Router.sol
hensha256: > yah we could leave it out of dispatcher, easiest case, have a V4SwapRouter contract in UR that inherits this and implements pay??

yeah i think thats going to be the best solution

contracts/interfaces/IV4Router.sol
hensha256: yeah im pro keeping it for now, other people could well use it even if we dont

contracts/V4Router.sol
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    /// @inheritdoc IUnlockCallback
+    function unlockCallback(bytes calldata encodedSwapInfo) external override returns (bytes memory) {
+        if (msg.sender != address(poolManager)) revert NotPoolManager();
+
+        SwapInfo memory swapInfo = abi.decode(encodedSwapInfo, (SwapInfo));
+
+        if (swapInfo.swapType == SwapType.ExactInput) {
+            _swapExactInput(abi.decode(swapInfo.params, (IV4Router.ExactInputParams)), swapInfo.msgSender);
```
hensha256, line 37: i wonder if we should be trying to do gas-efficient abi decoding like we do in universal router given it saved so much gas? and then always using calldata and not memory if we never alter the parameters?
i could do that in like a gas-golfing PR later though to see if it works and how much it helps

contracts/V4Router.sol
```diff
@@ -0,0 +1,186 @@
...
+            uint256 pathLength = params.path.length;
+            uint128 amountOut;
+
+            for (uint256 i = 0; i < pathLength; i++) {
+                (PoolKey memory poolKey, bool zeroForOne) = _getPoolAndSwapDirection(params.path[i], params.currencyIn);
+                amountOut = uint128(
+                    _swap(
+                        poolKey,
+                        zeroForOne,
+                        int256(-int128(params.amountIn)),
```
hensha256, line 75: is this safe? it casts uint128 to int128 (unsafe), then flips, then goes to int256. Might be better to do
```
int256(-uint256(params.amountIn))
```
that feels safer?

contracts/V4Router.sol
```diff
@@ -0,0 +1,186 @@
...
+
+            if (amountOut < params.amountOutMinimum) revert TooLittleReceived();
+        }
+    }
+
+    function _swapExactOutputSingle(IV4Router.ExactOutputSingleParams memory params, address msgSender) private {
+        _swap(
+            params.poolKey,
+            params.zeroForOne,
+            int256(int128(params.amountOut)),
```
hensha256, line 96: again feels more "correct" to do `int256(uint256(params.amountOut))`?

contracts/V4Router.sol
```diff
@@ -0,0 +1,186 @@
...
+        );
+
+        if (zeroForOne) {
+            reciprocalAmount = amountSpecified < 0 ? delta.amount1() : delta.amount0();
+            if (settle) _payAndSettle(poolKey.currency0, msgSender, delta.amount0());
+            if (take) poolManager.take(poolKey.currency1, msgSender, uint128(delta.amount1()));
+        } else {
+            reciprocalAmount = amountSpecified < 0 ? delta.amount0() : delta.amount1();
+            if (settle) _payAndSettle(poolKey.currency1, msgSender, delta.amount1());
+            if (take) poolManager.take(poolKey.currency0, msgSender, uint128(delta.amount0()));
```
hensha256, line 162: so thoughts on this `msgSender` parameter
- what if we want to set a different recipient? sometimes in UR we want the recipient to be the UR itself, say if the next trade is then a V2 trade so we want the output of a V4 trade to be the UR so it can be used as an input to a new protocol. So i think we need a recipient parameter?
e.g. for v3 we do this
```
v3SwapExactInput(map(recipient), amountIn, amountOutMin, path, payer);
```
and that `map` function then figures out the recipient's address and passes it in. I think we need that here too.

- what if we want to set a different payer? again in UR sometimes the payer of the transaction should be the UR itself if this is in the second half of a trade that started on another protocol version. In that code snippet above, the `payer` parameter is the address of who is paying, which UR defines as
```
address payer = payerIsUser ? lockedBy : address(this);
```
I think we again need that possibility here? So I think we need to pass in `payer` and `recipient` instead of `msgSender`?

hensha256 (commented):
Left some thoughts!

src/V4Router.sol
```diff
@@ -0,0 +1,179 @@
...
+    function _swapExactOutput(IV4Router.ExactOutputParams memory params) private {
+        unchecked {
+            uint256 pathLength = params.path.length;
+            uint128 amountIn;
+
+            for (uint256 i = pathLength; i > 0; i--) {
+                (PoolKey memory poolKey, bool oneForZero) =
+                    _getPoolAndSwapDirection(params.path[i - 1], params.currencyOut);
+                amountIn = uint128(
+                    -_swap(poolKey, !oneForZero, int256(int128(params.amountOut)), 0, params.path[i - 1].hookData)
```
snreynolds, line 115: why is this safe to flip the sign? is it guaranteed to be negative? its just not super obvious

src/V4Router.sol
```diff
@@ -0,0 +1,179 @@
...
+            ? (currencyIn, params.intermediateCurrency)
+            : (params.intermediateCurrency, currencyIn);
+
+        zeroForOne = currencyIn == currency0;
+        poolKey = PoolKey(currency0, currency1, params.fee, params.tickSpacing, params.hooks);
+    }
+
+    function _take(Currency currency, address recipient) private {
+        int256 delta = poolManager.currencyDelta(address(this), currency);
+        if (delta < 0) revert();
```
snreynolds, line 162: custom error?

snreynolds: also I feel like TAKE should accept a parameter? not just use the open delta? what if you want to take then swap then pay back or something?

src/V4Router.sol
```diff
@@ -0,0 +1,179 @@
...
+
+        reciprocalAmount = (zeroForOne == amountSpecified < 0) ? delta.amount1() : delta.amount0();
+    }
+
+    function _getPoolAndSwapDirection(PathKey memory params, Currency currencyIn)
+        private
+        pure
+        returns (PoolKey memory poolKey, bool zeroForOne)
+    {
+        (Currency currency0, Currency currency1) = currencyIn < params.intermediateCurrency
```
snreynolds, line 152: cache params.intermediateCurrency? see if that saves gas? it also might be easier to read each time haha

src/V4Router.sol
snreynolds: also todo.. use CurrencySettle library

src/V4Router.sol
```diff
@@ -0,0 +1,179 @@
...
+    using TransientStateLibrary for IPoolManager;
+    using CalldataBytesLib for bytes;
+
+    constructor(IPoolManager poolManager) BaseActionsRouter(poolManager) {}
+
+    function _handleAction(uint256 action, bytes calldata params) internal override {
+        // swap actions and payment actions in different blocks for gas efficiency
+        if (action < 0x10) {
+            if (action == Actions.SWAP_EXACT_IN) {
+                _swapExactInput(abi.decode(params, (IV4Router.ExactInputParams)));
```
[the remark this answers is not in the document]
hensha256: #199

src/V4Router.sol
hensha256: Theres actually a lib to do this. Have imported the lib, and then cached this in the lib

src/V4Router.sol
```diff
@@ -0,0 +1,179 @@
...
+            } else if (action == Actions.TAKE) {
+                // equivalent: abi.decode(params, (Currency, address))
+                Currency currency;
+                address recipient;
+                assembly ("memory-safe") {
+                    currency := calldataload(params.offset)
+                    recipient := calldataload(add(params.offset, 0x20))
+                }
+
+                // TODO add min amount??
```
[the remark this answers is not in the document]
hensha256: yeah sorry thats exactly what it is, making it clearer now

src/V4Router.sol
hensha256: > what if you want to take then swap then pay back or something?

what do you mean by that?
but yes, either way, when it inherits from DeltaResolver it will take an amount field

src/V4Router.sol
```diff
@@ -0,0 +1,179 @@
...
+            params.hookData
+        );
+    }
+
+    function _swapExactInput(IV4Router.ExactInputParams memory params) private {
+        unchecked {
+            uint256 pathLength = params.path.length;
+            uint128 amountOut;
+
+            for (uint256 i = 0; i < pathLength; i++) {
```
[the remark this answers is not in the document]
hensha256: We batch here because the input to the second step is the output from the first step.
Like if you have a split-route like this:

1) 100 UNI -> DAI -> USDC
2) 50 UNI -> USDT -> USDC

How much DAI should be traded for USDC? everything that was produced by the UNI->DAI trade. So it just does this loop to pass values through. 
It also saves having so many parameters because the trade is passed in as `100 UNI->DAI->USDC` instead of `100 UNI -> DAI`, `X DAI -> USDC`.

So the example split route above would be passed in as 2 swap commands one for each route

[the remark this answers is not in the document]
hensha256: This is how our V2 and V3 swap routers have worked too - passing in an entire path beginning to end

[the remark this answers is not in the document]
hensha256: I can try to play around with passing values between commands like you suggest and see what happens

src/V4Router.sol
hensha256: >what do you mean by that?

i understand this comment now. Yeah this will be done when `DeltaResolver` is inherited 👯

test/V4Router.t.sol
```diff
@@ -0,0 +1,455 @@
...
+
+        uint256 prevBalance0 = key0.currency0.balanceOf(address(this));
+        uint256 prevBalance1 = key0.currency1.balanceOf(address(this));
+
+        plan = plan.add(Actions.SWAP_EXACT_IN_SINGLE, abi.encode(params));
+        _finalizePlan(key0.currency0, key0.currency1, address(this));
+        bytes memory data = plan.encode();
+
+        router.executeActions(data);
+        snapLastCall("RouterExactInputSingle");
```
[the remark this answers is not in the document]
hensha256: The thing i dont like about separating them is you then have to do double the number of tests? Like you have to do
`test_swapExactInputSingle_zeroForOne` AND
`test_gas_swapExactInputSingle_zeroForOne`

and they both do the same thing? but one does `assertEq(...)` to check things and the other does gas.

Do you have a way of removing all that duplicated test code?

src/V4Router.sol
hensha256: so amountSpecified is positive (because params.amountOut is positive), which makes it exactOut. Which means itll be swapped and generate an input delta, which will be negative. WIll leave a comment

test/V4Router.t.sol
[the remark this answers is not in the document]
snreynolds: I actually like it separated out.. like in posm we have a Gas.t.sol file with no asserts and i think its so much easier to read what exactly is being tested gas wise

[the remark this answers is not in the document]
snreynolds: but theres not really a way afaik to deduplicate that.. but IMO they are soo much easier to write/add new gas tests for that I honestly think its easier and quicker to do it this way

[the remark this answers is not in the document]
snreynolds: With the other pattern too, not every assert test has a gas snapshot so its really confusing & the name of the test with asserts doesnt tell you/inform you well about the gas sometimes 

like test_exactInputZeroForOne
but we should have many test_gas_exactInputZeroForOne like where the pool is warm vs. when its not

src/V4Router.sol
```diff
@@ -0,0 +1,179 @@
...
+                _swapExactOutputSingle(abi.decode(params, (IV4Router.ExactOutputSingleParams)));
+            } else {
+                revert UnsupportedAction(action);
+            }
+        } else {
+            if (action == Actions.SETTLE) {
+                // equivalent: abi.decode(params, (Currency))
+                Currency currency;
+                assembly ("memory-safe") {
+                    currency := calldataload(params.offset)
```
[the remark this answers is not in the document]
snreynolds: hahah nice

src/V4Router.sol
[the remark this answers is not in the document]
snreynolds: hmmm yeah I guess this is just structurally different from POSM that doesn't do subbatching even when it "could" roll its positions. i just find it odd to write a batched actions router where the command below are basically another batch. it just means you could have 2 less commands bc you can just encode many swapExactInputSingles together to get a full swapExactInput trade

in posm, we dont have anything like this but we COULD.. ie decrease (USDC, ETH) then increase (ETH, DAI) is technically like a "path" but it requires calling decrease then increase with the full params

src/V4Router.sol
snreynolds: great! yeah comment would be helpful its hard to remember all the signs for specific swap types

src/base/DeltaResolver.sol
```diff
@@ -1,17 +1,33 @@
...
+
+    /// @notice Emitted trying to settle a positive delta, or take a negative delta
+    error InvalidDeltaForAction();
+
     /// @notice Take an amount of currency out of the PoolManager
     /// @param currency Currency to take
     /// @param recipient Address to receive the currency
     /// @param amount Amount to take
     function _take(Currency currency, address recipient, uint256 amount) internal {
+        if (amount == ENTIRE_OPEN_DELTA) {
```
snreynolds, line 25: I do think in general I'd prefer this logic to be handled outside of this library. I view DeltaResolver as a pretty agnostic library that is really just used to share code for the direct poolManager calls to `take`, `settle`, `mint`, and `burn`.

snreynolds: Also I think you and I have been in general MAXing not having to use booleans to do certain behavior which is why I like "close" it just knows you want to settle the whole delta. If we move to this we're checking booleans/parameters in commands. id just rather have more commands that do specific behaviors

src/libraries/CalldataBytesLib.sol
```diff
@@ -2,12 +2,47 @@
...
     bytes4 constant SLICE_ERROR_SELECTOR = 0x3b99b53d;
 
+    /// @notice Decode the `_arg`-th element in `_bytes` as a dynamic array
+    /// @dev The decoding of `length` and `offset` is universal,
+    /// whereas the type declaration of `res` instructs the compiler how to read it.
+    /// @param _bytes The input bytes string to slice
+    /// @param _arg The index of the argument to extract
+    /// @return length Length of the array
+    /// @return offset Pointer to the data part of the array
+    function toLengthOffset(bytes calldata _bytes, uint256 _arg)
```
snreynolds, line 18: why do we need toLengthOffset and toBytes? Not sure where they are being used?

src/libraries/CalldataBytesLib.sol
```diff
@@ -2,12 +2,47 @@
...
+        (uint256 length, uint256 offset) = toLengthOffset(_bytes, _arg);
+        assembly {
+            res.length := length
+            res.offset := offset
+        }
+    }
+
     /// @notice Performs the equivalent of `abi.decode(data, (uint256[], bytes[]))` in calldata
     /// @param _bytes The input bytes string to extract input arrays from
     /// @return actions The uint256 calldata array of actions
```
snreynolds, line 48: I think decodeInCalldata is not memory-safe fyi

src/libraries/CalldataBytesLib.sol
hensha256: removed

src/base/DeltaResolver.sol
hensha256: yeah thats fair, i've added a new command. I think routing will want generic TAKE and also TAKE_PORTION. But for now implemented TAKE_ALL for the generic swap case

src/V4Router.sol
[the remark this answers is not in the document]
hensha256: yeah thats fair... i think it will cost way more gas because we will have to pass outputs between commands, or load deltas from pool manager, and pass in each token twice so way more calldata.
Its like how posm will have a command CLOSE_PAIR which could really just be CLOSE done twice, to be optimal.
I'm down to open an issue and explore it after - and we can get emily's opinion too given shes done a lot of routing work for both v4 and previous versions.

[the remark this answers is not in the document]
hensha256: opened #211

src/libraries/Actions.sol
```diff
@@ -5,17 +5,25 @@ pragma solidity ^0.8.24;
...
-    uint256 constant CLEAR = 0x14;
+    uint256 constant SETTLE_ALL = 0x11;
+
+    uint256 constant TAKE = 0x12;
+    uint256 constant TAKE_ALL = 0x13;
+    uint256 constant TAKE_PORTION = 0x14;
+
+    uint256 constant CLOSE_CURRENCY = 0x15;
+    uint256 constant CLOSE_PAIR = 0x16;
+    uint256 constant CLEAR = 0x17;
```
snreynolds, line 26: I think we should discuss offline which of these we can share between router/posm!

src/V4Router.sol
```diff
@@ -0,0 +1,168 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {BalanceDelta} from "@uniswap/v4-core/src/types/BalanceDelta.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+import {Currency, CurrencyLibrary} from "@uniswap/v4-core/src/types/Currency.sol";
+import {TickMath} from "@uniswap/v4-core/src/libraries/TickMath.sol";
+import {TransientStateLibrary} from "@uniswap/v4-core/src/libraries/TransientStateLibrary.sol";
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
```
snreynolds, line 10: double import

src/V4Router.sol
```diff
@@ -0,0 +1,168 @@
...
+/// @title UniswapV4Router
+/// @notice Abstract contract that contains all internal logic needed for routing through Uniswap V4 pools
+/// @dev the entry point to executing actions in this contract is calling `BaseActionsRouter._executeActions`
+/// An inheriting contract should call _executeActions at the point that they wish actions to be executed
+abstract contract V4Router is IV4Router, BaseActionsRouter, DeltaResolver {
+    using PathKeyLib for PathKey;
+    using CalldataDecoder for bytes;
+    using TransientStateLibrary for IPoolManager;
+
+    constructor(IPoolManager poolManager) BaseActionsRouter(poolManager) {}
```
snreynolds, line 28:
```suggestion
    constructor(IPoolManager _poolManager) BaseActionsRouter(_poolManager) {}
```

src/V4Router.sol
```diff
@@ -0,0 +1,168 @@
...
+                // equivalent: abi.decode(params, (Currency))
+                Currency currency;
+                assembly ("memory-safe") {
+                    currency := calldataload(params.offset)
+                }
+
+                int256 delta = poolManager.currencyDelta(address(this), currency);
+                if (delta > 0) revert InvalidDeltaForAction();
+
+                // TODO support address(this) paying too
```
snreynolds, line 56: yep - lets discuss this too! posm needs similar things!!

src/V4Router.sol
```diff
@@ -0,0 +1,168 @@
...
+                // TODO should it have a maxAmountOut added slippage protection?
```
snreynolds, line 57: my thought is that slippage protection should be with the swap commands? why do we also need slippage protection with settle commands?

i think we can make settle just take an amount, and if you're ok with the full settle amount then you pass in a constant/sentinel

src/V4Router.sol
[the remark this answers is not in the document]
snreynolds: sounds good, I think we may run out of time to explore this though. It makes sense as an optimization for router and the design similarities with prev routers makes sense

src/V4Router.sol
hensha256: Its to support split route trades. Lets say youre trading 1000 DAI for USDC, where you want at least 998 USDC
- 500 DAI -> USDT -> USDC
- 500 DAI -> USDC

Then you would do a command thats like "take all USDC, and check we're receiving at least 998".

hensha256: We implement the same style of logic in UR using the sweep command e.g. [here](https://github.com/Uniswap/universal-router/blob/4cf96077ba21dbbbbbc8ffa8c549648f69623fbc/contracts/modules/Payments.sol#L80)

hensha256: And we do it with the unwrapETH command to do the same check for ETH output trades e.g. [here](https://github.com/Uniswap/universal-router/blob/4cf96077ba21dbbbbbc8ffa8c549648f69623fbc/contracts/modules/Payments.sol#L130)

snreynolds: but surely you're also checking slippage from USDT to USDC and from DAI to USDC? or are you ignoring those checks until the end? My concern is that we're doing lots of slippage checks and the ones in the middle then dont even matter?

src/V4Router.sol
```diff
@@ -0,0 +1,165 @@
...
+                _settle(currency, _msgSender(), uint256(-delta));
+            } else if (action == Actions.TAKE_ALL) {
+                // equivalent: abi.decode(params, (Currency, address))
+                Currency currency;
+                address recipient;
+                assembly ("memory-safe") {
```
snreynolds, line 62: [#199](https://github.com/Uniswap/v4-periphery/issues/199) just tagging that 199 should also put these in a lib so we can use in both contracts

src/V4Router.sol
```diff
@@ -0,0 +1,165 @@
...
+            uint256 pathLength = params.path.length;
+            uint128 amountIn;
+            uint128 amountOut = params.amountOut;
+            Currency currencyOut = params.currencyOut;
+            PathKey memory pathKey;
+
+            for (uint256 i = pathLength; i > 0; i--) {
+                pathKey = params.path[i - 1];
+                (PoolKey memory poolKey, bool oneForZero) = pathKey.getPoolAndSwapDirection(currencyOut);
+                amountIn = uint128(-_swap(poolKey, !oneForZero, int256(uint256(amountOut)), 0, pathKey.hookData));
```
snreynolds, line 133: I feel like I commented at some point about documenting/adding a little comment about the return signs but can't find it anywhere.. Anyways I do think it would be nice explaining that the output is guaranteed to be negative for exactOutput in the reciprocal token.

src/base/DeltaResolver.sol
```diff
@@ -7,6 +7,9 @@ import {ImmutableState} from "./ImmutableState.sol";
 /// @notice Abstract contract used to sync, send, and settle funds to the pool manager
 /// @dev Note that sync() is called before any erc-20 transfer in `settle`.
 abstract contract DeltaResolver is ImmutableState {
+    /// @notice Emitted trying to settle a positive delta, or take a negative delta
```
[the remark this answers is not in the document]
snreynolds: Oh weird.. on codespaces the code I was reviewing had the changes where you kept the flag for checking to use the contract balance.. I guess we can do that in a different PR!

test/shared/RoutingTestHelpers.sol
```diff
@@ -0,0 +1,115 @@
...
+            path[i - 1] = PathKey(_tokenPath[i - 1], 3000, 60, IHooks(address(0)), bytes(""));
+        }
+
+        params.currencyOut = _tokenPath[_tokenPath.length - 1];
+        params.path = path;
+        params.amountOut = uint128(amountOut);
+        params.amountInMaximum = type(uint128).max;
+    }
+
+    function _finalizePlan(Currency inputCurrency, Currency outputCurrency, address recipient) internal {
```
snreynolds, line 111: wanna add this to Planner.sol? maybe just overload planner.finalize? or you could call it like planner.finalizeSwap()?

test/shared/RoutingTestHelpers.sol
```diff
@@ -0,0 +1,115 @@
...
+import {Currency} from "@uniswap/v4-core/src/types/Currency.sol";
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+import {Deployers} from "@uniswap/v4-core/test/utils/Deployers.sol";
+import {PositionManager} from "../../src/PositionManager.sol";
+import {IERC20} from "forge-std/interfaces/IERC20.sol";
+import {LiquidityOperations} from "./LiquidityOperations.sol";
+import {IV4Router} from "../../src/interfaces/IV4Router.sol";
+
+/// @notice A shared test contract that wraps the v4-core deployers contract and exposes basic liquidity operations on posm.
```
snreynolds, line 22:
```suggestion
/// @notice A shared test contract that wraps the v4-core deployers contract and exposes basic helpers for swapping with the router.
```

test/router/V4Router.t.sol
```diff
@@ -0,0 +1,327 @@
...
+        uint256 newBalance0 = key0.currency0.balanceOf(address(this));
+        uint256 newBalance1 = key0.currency1.balanceOf(address(this));
+
+        assertEq(prevBalance0 - newBalance0, amountIn);
+        assertEq(newBalance1 - prevBalance1, expectedAmountOut);
+    }
+
+    function test_swapExactInputSingle_oneForZero() public {
+        uint256 amountIn = 1 ether;
+        uint256 expectedAmountOut = 992054607780215625;
```
snreynolds, line 44: where is this number from? I think you could use deltas instead of hardcoding?

snreynolds (commented):
ok looks good! I do think we should be adding some more tests (like split routes) but we can followup?

test/shared/RoutingTestHelpers.sol
hensha256: LOL ty

src/V4Router.sol
hensha256: > but surely you're also checking slippage from USDT to USDC and from DAI to USDC? or are you ignoring those checks until the end? My concern is that we're doing lots of slippage checks and the ones in the middle then dont even matter?

yeah the ones in the middle dont matter. the user doesnt care if one half of the trade gets hit with more slippage than the other - theres just an output amount that they want from the trade in total and thats it

hensha256: >My other thought process then also is just if posm & routers take/settle commands looks different now

yeah i could do it as a different cmmand or something depending how the other commands end up

src/V4Router.sol
hensha256: have actually added a safecast on this - i think with custom accounting hooks on low liquidity pools its technically not guaranteed

test/router/V4Router.t.sol
hensha256: gonna do this in the next PR - #221

src/libraries/SafeCast.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+import {CustomRevert} from "@uniswap/v4-core/src/libraries/CustomRevert.sol";
+
+/// @title Safe casting methods
+/// @notice Contains methods for safely casting between types
+library SafeCast {
```
snreynolds, line 8: why not just import SafeCast from core?

hensha256: this function doesnt exist in core rip

snreynolds (approved):
approving, but will need to move that func to core! (alice opening issue)

PR #60, opened 2023-09-14
Replace Implementation+etch Pattern with CREATE2

### LINKED ISSUE #59: Replace implementation+etch pattern with Create2Deployer
### Component

General design optimization (improving efficiency, cleanliness, or developer experience)

Arachnid's [create2 deployer proxy](https://github.com/Arachnid/deterministic-deployment-proxy) is now available on anvil and forge test by default

This allows us to deterministically mine + deploy arbitrary contracts without the need for a factory or *Implementation* wrapper that overrides `validateHookAddress`.

A minor change but it reduces repo clutter+boilerplate by adding consistent abstractions around hook development and testing

Currently you need to define an *Implementation* which overrides the flag checks, and then etch the bytecode to a correctly-flagged address:

```solidity
       LimitOrder limitOrder = LimitOrder(address(uint160(Hooks.AFTER_INITIALIZE_FLAG | Hooks.AFTER_SWAP_FLAG)));
        vm.record();
        LimitOrderImplementation impl = new LimitOrderImplementation(manager, limitOrder);
        (, bytes32[] memory writes) = vm.accesses(address(impl));
        vm.etch(address(limitOrder), address(impl).code);
        // for each storage key that was written during the hook implementation, copy the value over
        unchecked {
            for (uint256 i = 0; i < writes.length; i++) {
                bytes32 slot = writes[i];
                vm.store(address(limitOrder), slot, vm.load(address(impl), slot));
            }
        }
```

Abstracting the create2 deployer proxy, we get a much cleaner devex:
```solidity
        // Deploy the hook to an address with the correct flags
        uint160 flags = uint160(Hooks.AFTER_INITIALIZE_FLAG | Hooks.AFTER_SWAP_FLAG);
        bytes memory hookBytecode = abi.encodePacked(type(LimitOrder).creationCode, abi.encode(address(manager)));
        (, uint256 salt) = HookDeployer.mineSalt(CREATE2_DEPLOYER, flags, hookBytecode);

        LimitOrder limitOrder = new LimitOrder{salt: salt}(address(manager));
```

### Describe alternatives.

N/A

### Additional context.

Already works and is available here:

https://github.com/saucepoint/v4-template/blob/main/test/Counter.t.sol#L30-L37

and

https://github.com/saucepoint/v4-template/blob/main/test/utils/HookDeployer.sol

---

I'm happy to PR and update all the tests to the new pattern!

saucepoint:
## Related Issue

Closes #59 

By using CREATE2 natively, we reduce the testing complexity. Prior to this change, testing a hook required a wrapper contract ('Implementation') that overrides `validateHookAddress` as a no-op. Within tests, users need to define both contracts, and etch the 'Implementation' bytecode to a desired address

This pattern is noticeably verbose, with little reusable patterns that leads to copy + paste boilerplate

## Description of changes

* Created a helper library `test/utils/HookMiner.sol`, which finds a salt to deploy a hook with a specific flag pattern

* Updated the tests to deploy hooks with CREATE2 instead of the implementation and vm.etch pattern

Example:

Before, where `TWAMMImplementation` is a boilerplate wrapper contract around the actual `TWAMM`
```solidity
        TWAMM twamm = TWAMM(
            address(uint160(Hooks.BEFORE_INITIALIZE_FLAG | Hooks.BEFORE_SWAP_FLAG | Hooks.BEFORE_MODIFY_POSITION_FLAG))
        );
        TWAMMImplementation impl = new TWAMMImplementation(manager, 10_000, twamm);
        (, bytes32[] memory writes) = vm.accesses(address(impl));
        vm.etch(address(twamm), address(impl).code);
        // for each storage key that was written during the hook implementation, copy the value over
        unchecked {
            for (uint256 i = 0; i < writes.length; i++) {
                bytes32 slot = writes[i];
                vm.store(address(twamm), slot, vm.load(address(impl), slot));
            }
        }
```

After:
```solidity
        // Find a salt that produces a hook address with the desired `flags`
        uint160 flags =
            uint160(Hooks.BEFORE_INITIALIZE_FLAG | Hooks.BEFORE_SWAP_FLAG | Hooks.BEFORE_MODIFY_POSITION_FLAG);
        (address hookAddress, bytes32 salt) =
            HookMiner.find(address(this), flags, 0, type(TWAMM).creationCode, abi.encode(manager, 10_000));

        // Deploy hook to the precomputed address using the salt
        twamm = new TWAMM{salt: salt}(manager, 10_000);
```


test/utils/HookMiner.sol
```diff
@@ -0,0 +1,59 @@
...
+library HookMiner {
+    // mask to slice out the top 8 bit of the address
+    uint160 constant FLAG_MASK = 0xFF << 152;
+
+    // Maximum number of iterations to find a salt, avoid infinite loops
+    uint256 constant MAX_LOOP = 10_000;
+
+    /// @notice Find a salt that produces a hook address with the desired `flags`
+    /// @param deployer The address that will deploy the hook. In `forge test`, this will be the test contract `address(this)` or the pranking address
+    ///                 In `forge script`, this should be `0x4e59b44847b379578588920cA78FbF26c0B4956C` (CREATE2 Deployer Proxy)
```
ewilz, line 15: conventionally we don't have this kind of whitespace in our natspec, I have found some 4 space indentations for multiline natspec if you don't mind changing to that?

test/FullRange.t.sol
```diff
@@ -90,8 +88,12 @@ contract TestFullRange is Test, Deployers, GasSnapshot {
...
         manager = new PoolManager(500000);
 
-        FullRangeImplementation impl = new FullRangeImplementation(manager, fullRange);
-        vm.etch(address(fullRange), address(impl).code);
+        uint160 flags =
+            uint160(Hooks.BEFORE_INITIALIZE_FLAG | Hooks.BEFORE_MODIFY_POSITION_FLAG | Hooks.BEFORE_SWAP_FLAG);
+        (address hookAddress, bytes32 salt) =
+            HookMiner.find(address(this), flags, 0, type(FullRange).creationCode, abi.encode(manager));
+        fullRange = new FullRange{salt: salt}(manager);
+        require(address(fullRange) == hookAddress, "TestFullRange: hook address mismatch");
```
ewilz, line 96: confused about this check. Is there a scenario where this would be false and we'd want to halt the test? Doesn't the check on line 32 in HookMiner take care of this? Feels extra, or like we don't trust our hook miner...in which case it could maybe have its own tests 🤔

test/GeomeanOracle.t.sol
```diff
@@ -166,7 +157,7 @@ contract TestGeomeanOracle is Test, Deployers, TokenFixture {
 
     function testBeforeModifyPositionObservationAndCardinality() public {
         manager.initialize(key, SQRT_RATIO_2_1, ZERO_BYTES);
-        geomeanOracle.setTime(3); // advance 2 seconds
+        skip(2); // advance 2 seconds
```
ewilz, line 160: nice

ewilz:
lgtm other than the nits. impressed with the lack of latency on the mining!?

test/FullRange.t.sol
saucepoint: It wouldnt fail and we can trust the miner. It's really a safety check since there's a minor footgun with the `deployer` argument in `HookMiner.find(address deployer, ...)`

In tests, `deployer` can be `address(this)` or the pranking address from `vm.prank`. There's a probability that the mined salt creates a valid address with the wrong deployer. i.e. `HookMiner.find(alice, ...)` but forgetting to prank on `new FullRange`. This would cause a mismatch between `hookAddress` and `address(fullRange)`

The hook would still deploy but a dev might use `hookAddress` in a way that causes false positive bugs/errors.

Happy to remove it or change it to `assertEq` (to imply a safety check)

saucepoint: after thinking about it some more, decided to remove it entirely. while it is good to encourage safety checks, i dont think its something that had to be showcased in the codebase

saucepoint:
replaced by https://github.com/Uniswap/v4-periphery/pull/122

mining addresses in tests has been reported to be annoyingly slow

PR #62, opened 2023-09-23
Update TWAMM.sol

0xScratch:
Well, this PR made 2 types of changes and these changes feels like valid as far as I am concerned. If not then do let me know and tell the reason which will be another great learning experience. Anyways let's dive into the changes made:

**1. Struct Packing:** As we know, that solidity use slots and the usage of it really affects the gas (means more the slots more gas used by the contract). I have noticed that rest of the `structs` were perfectly packed but not this one. Earlier it was using 4 slots and now decreased to 3.

**2. Changed `bool [variable-name] = false;` to `bool [variable-name];`:** Not really sure about this change (looks like it's been done on purpose) but while I was going through the repository, there were several examples of unsigned integers not initialized to any value, thus leaving their value as 0 by default. And, same concept applies to this boolean variable. When not initialized, it's default value gonna be false


ewilz:
PR is incomplete and adds small gas increase when run with `forge test`. Not necessary, so closing.

PR #70, opened 2023-11-01
Update to v4-core latest: solmate MockERC20

saucepoint:
## Related Issue
Syncing with the latest v4-core

## Description of changes

Breaking changes that are fixed in this PR:

* `IDynamicFeeManager` interface was changed
* `v4-core's MockERC20` was removed in favor of solmate/MockERC20


saucepoint:
stale

PR #73, opened 2023-11-15
feat: Revert style quoter

ConjunctiveNormalForm:
## Which issue does this pull request resolve?
Quoter contract for off-chain use
## Description of changes
Full implementation of revert-style quoter, and migrated over all test cases for V3's QuoterV2


contracts/libraries/SwapIntention.sol
ConjunctiveNormalForm, line 1: This is probably terrible naming but I can't come up with anything better 

Also reason this is not in quoter interface is that Router would probably share the same structs

contracts/interfaces/IQuoter.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.20;
+
+import "../libraries/SwapIntention.sol";
+import {BalanceDelta} from "@uniswap/v4-core/contracts/types/BalanceDelta.sol";
```
marktoda, line 5: nit: unused?

marktoda: same as IpoolManager and PoolKey

contracts/interfaces/IQuoter.sol
```diff
@@ -0,0 +1,31 @@
...
+    error InvalidQuoteType();
+    error InvalidQuoteTypeInRevert();
+    error UnexpectedRevertBytes();
+
+    struct NonZeroDeltaCurrency {
+        Currency currency;
+        int128 deltaAmount;
+    }
+
+    function quoteExactInputSingle(ExactInputSingleParams calldata params)
```
marktoda, line 20: nit: natspec here, then you can reference it from the implementation with @inheritdoc

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+import {PoolId, PoolIdLibrary} from "@uniswap/v4-core/contracts/types/PoolId.sol";
+import {PoolManager} from "@uniswap/v4-core/contracts/PoolManager.sol";
+
+contract Quoter is IQuoter {
+    using PoolIdLibrary for PoolKey;
+    using SafeCast for *;
+    using Hooks for IHooks;
+
+    // v4 Singleton contract
+    IPoolManager poolManager;
```
marktoda, line 26: since it is defined in constructor and never changes, you can specify it as `immutable` which has big gas benefits as it's written into the contract code rather than sloaded every time its read

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+import {IPoolManager} from "@uniswap/v4-core/contracts/interfaces/IPoolManager.sol";
+import {BalanceDelta} from "@uniswap/v4-core/contracts/types/BalanceDelta.sol";
+import {Currency} from "@uniswap/v4-core/contracts/types/Currency.sol";
+import {PoolKey} from "@uniswap/v4-core/contracts/types/PoolKey.sol";
+import {PoolId, PoolIdLibrary} from "@uniswap/v4-core/contracts/types/PoolId.sol";
+import {PoolManager} from "@uniswap/v4-core/contracts/PoolManager.sol";
+
+contract Quoter is IQuoter {
+    using PoolIdLibrary for PoolKey;
+    using SafeCast for *;
```
marktoda, line 22: I don't see any actual usages of this?

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+    using Hooks for IHooks;
+
+    // v4 Singleton contract
+    IPoolManager poolManager;
+
+    constructor(address _poolManager) {
+        poolManager = IPoolManager(_poolManager);
+    }
+
+    function fillSlot0(PoolId id) private view returns (Pool.Slot0 memory slot0) {
```
marktoda, line 32: why `fill`? seems like `getSlot0`

marktoda, line 32: nit: generally put privates and internals below publics and externals

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+        //TODO: extsload when storage is stable?
+        (slot0.sqrtPriceX96, slot0.tick,,) = poolManager.getSlot0(id);
+
+        return slot0;
+    }
+
+    function validateRevertReason(bytes memory reason) private pure returns (bytes memory) {
+        if (reason.length < 96) {
```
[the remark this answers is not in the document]
marktoda: also seems like this validation should differ between exactInput and exactInputSingle since they have different sizes of return values, no? i would assume we need separate validation for each

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+            abi.decode(reason, (int128[], uint160[], uint32[]));
+    }
+
+    function lockAcquired(bytes calldata encodedSwapIntention) external returns (bytes memory) {
+        require(msg.sender == address(poolManager));
+
+        SwapInfo memory swapInfo = abi.decode(encodedSwapIntention, (SwapInfo));
+
+        if (swapInfo.swapType == SwapType.ExactInputSingle) {
+            (BalanceDelta deltas, uint160 sqrtPriceX96After, int24 tickAfter) =
```
marktoda, line 92: think each branch here deserves its own helper function

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+        int256 amountSpecified,
+        uint160 sqrtPriceLimitX96,
+        bytes memory hookData
+    ) private returns (BalanceDelta deltas, uint160 sqrtPriceX96After, int24 tickAfter) {
+        deltas = poolManager.swap(
+            poolKey,
+            IPoolManager.SwapParams({
+                zeroForOne: zeroForOne,
+                amountSpecified: amountSpecified,
+                sqrtPriceLimitX96: sqrtPriceLimitX96 == 0
```
marktoda, line 262: double ternary is tough to read, would pull this out

contracts/libraries/SwapIntention.sol
```diff
@@ -0,0 +1,75 @@
...
+import {PoolKey} from "@uniswap/v4-core/contracts/types/PoolKey.sol";
+
+enum SwapType {
+    ExactInput,
+    ExactInputSingle,
+    ExactOutput,
+    ExactOutputSingle
+}
+
+struct SwapInfo {
```
marktoda, line 16: This one is pretty specific to Quoter, right? so should probably live in Quoter.sol

marktoda: same with SwapType enum

contracts/lens/Quoter.sol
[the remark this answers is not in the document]
marktoda: oh i guess 96 is from exactInputSingle revert which is 3 words long, yea comment would be good :D

contracts/lens/Quoter.sol
marktoda: one nice pattern that we've used in the past when lock has multiple branches is to encode a sub-call in the lock and self-call in lockAcquired
as defined here https://github.com/Uniswap/v4-periphery/blob/581d96dfd8b281cafe351205bf6d2d65efb4df90/contracts/BaseHook.sol#L52-L61

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+        private
+        returns (
+            int128[] memory deltaAmounts,
+            uint160[] memory sqrtPriceX96AfterList,
+            uint32[] memory initializedTicksCrossedList
+        )
+    {
+        uint256 pathLength = params.path.length;
+
+        //mapping(Currency currency => int128) memory currencyDeltas;
```
marktoda, line 205: ? rm

contracts/libraries/SwapIntention.sol
```diff
@@ -0,0 +1,75 @@
...
+
+struct ExactOutputParams {
+    Currency currencyOut;
+    PathKey[] path;
+    address recipient;
+    uint128 amountOut;
+    uint160 sqrtPriceLimitX96;
+}
+
+library SwapIntention {
```
marktoda, line 62: General pattern is to have struct objects and libraries supporting them to be named after the struct, so here I would have a new library file called `PathKey` and define the struct in there and any helper functions associated with the struct in a library there as well

marktoda: similar to https://github.com/Uniswap/v3-periphery/blob/main/contracts/libraries/Path.sol

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+        BalanceDelta deltas;
+        deltaAmounts = new int128[](2);
+        (, tickBefore,,) = poolManager.getSlot0(poolKey.toId());
+        reason = validateRevertReason(reason);
+        (deltas, sqrtPriceX96After, tickAfter) = abi.decode(reason, (BalanceDelta, uint160, int24));
+        deltaAmounts[0] = deltas.amount0();
+        deltaAmounts[1] = deltas.amount1();
+
+        initializedTicksCrossed =
+            PoolTicksCounter.countInitializedTicksCrossed(poolManager, poolKey, tickBefore, tickAfter);
```
marktoda, line 69: why do we do the counting _post revert_ here but _pre revert_ for exactInput? think you could refactor this into `_quoteExactInputSingle` to deduplicate code across the paths

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+                poolKey: poolKey,
+                zeroForOne: zeroForOne,
+                recipient: params.recipient,
+                amountIn: params.amountIn,
+                sqrtPriceLimitX96: 0,
+                hookData: params.path[i].hookData
+            });
+            (BalanceDelta curDeltas, uint160 sqrtPriceX96After, int24 tickAfter) = _quoteExactInputSingle(singleParams);
+
+            deltaAmounts[i] += zeroForOne ? curDeltas.amount0() : curDeltas.amount1();
```
marktoda, line 226: nit: more readable i think to do 
```
(int256 deltaIn, int256 deltaOut) = zeroForOne? (amount0, amount1) : (amount1, amount0);
deltaAmounts[i] = deltaIn
deltaAmounts[i + 1] = deltaOut;
```

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+            deltaAmounts[i + 1] += zeroForOne ? curDeltas.amount1() : curDeltas.amount0();
+
+            params.amountIn = zeroForOne ? uint128(-curDeltas.amount1()) : uint128(-curDeltas.amount0());
+            params.currencyIn = params.path[i].intermediateCurrency;
```
marktoda, line 230: editing parameters is a bit of an antipattern, would do locals instead

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+                    ptr := add(ptr, 0x20)
+                }
+
+                mstore(ptr, initializedTicksLength)
+                ptr := add(ptr, 0x20)
+                for { let i := 0 } lt(i, initializedTicksLength) { i := add(i, 1) } {
+                    let value := mload(add(initializedTicksCrossedList, add(mul(i, 0x20), 0x20)))
+                    mstore(ptr, value)
+                    ptr := add(ptr, 0x20)
+                }
```
marktoda, line 161: sheeesshh nice assembly ahah. Though I'm a bit surprised you can't do something more like

```
bytes memory result = abi.encode(deltaAmounts, sqrtPrice, initializedTicksLength);
assembly {
   revert(result, result.length)
}
```
to get the same result

foundry.toml
```diff
@@ -2,11 +2,13 @@
 src = 'contracts'
 out = 'foundry-out'
 solc_version = '0.8.20'
-optimizer_runs = 800
+via_ir = true
```
marktoda, line 5: oof is it super slow now

foundry.toml
```diff
@@ -2,11 +2,13 @@
...
+optimizer_runs = 1000000
 ffi = true
 fs_permissions = [{ access = "read-write", path = ".forge-snapshots/"}]
 
 [profile.ci]
 fuzz_runs = 100000
+via_ir = true
```
marktoda, line 12: this inherits from default so shouldnt be needed here

test/Quoter.t.sol
```diff
@@ -0,0 +1,441 @@
...
+import {IPoolManager} from "@uniswap/v4-core/contracts/interfaces/IPoolManager.sol";
+import {Currency, CurrencyLibrary} from "@uniswap/v4-core/contracts/types/Currency.sol";
+import {TickMath} from "@uniswap/v4-core/contracts/libraries/TickMath.sol";
+
+contract QuoterTest is Test, Deployers {
+    using SafeCast for *;
+    using PoolIdLibrary for PoolKey;
+
+    // Min tick for full range with tick spacing of 60
+    int24 internal constant MIN_TICK = -887220;
```
marktoda, line 28: nit: import from TickMath

marktoda (commented):
awesome work!!! Some first pass comments, apologies for the volume hopefully its helpful!

contracts/lens/Quoter.sol
marktoda: ik gas not super important here since its a quoter, but does help us squeeze more quotes into one multicall

foundry.toml
ConjunctiveNormalForm: kinda but not super? sub minute compile time 
had to do this otherwise stack too deep

contracts/libraries/SwapIntention.sol
```diff
@@ -0,0 +1,68 @@
...
+    ExactOutput,
+    ExactOutputSingle
+}
+
+struct SwapInfo {
+    SwapType swapType;
+    bytes params;
+}
+
+struct ExactInputSingleParams {
```
marktoda, line 22: oo also i just remembered we have these defined here https://github.com/Uniswap/swap-router-contracts-next/blob/main/contracts/interfaces/IV4SwapRouter.sol so you can just import them!

.solhint.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "solhint:recommended",
```
zhongeric, line 2: don't think this should be checked in?

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,210 @@
...
+import {Currency} from "@uniswap/v4-core/contracts/types/Currency.sol";
+import {PoolKey} from "@uniswap/v4-core/contracts/types/PoolKey.sol";
+import {PoolIdLibrary} from "@uniswap/v4-core/contracts/types/PoolId.sol";
+
+contract Quoter is IQuoter {
+    using PoolIdLibrary for PoolKey;
+    using Hooks for IHooks;
+
+    // v4 Singleton contract
+    IPoolManager private immutable MANAGER;
```
zhongeric, line 21: nit: lowercase here instead of caps? Caps usually for constants

zhongeric: also can do public instead of private

test/Quoter.t.sol
```diff
@@ -0,0 +1,441 @@
...
+    MockERC20 token2;
+
+    PoolKey key01;
+    PoolKey key02;
+    PoolKey key12;
+
+    MockERC20[] tokenPath;
+
+    function setUp() public {
+        manager = new PoolManager(500000);
```
zhongeric, line 48: nit: magic number for fee controller gas limit

test/Quoter.t.sol
```diff
@@ -0,0 +1,441 @@
...
+        params.currencyIn = Currency.wrap(address(_tokenPath[0]));
+        params.path = path;
+        params.recipient = address(this);
+        params.amountIn = uint128(amountIn);
+    }
+
+    function logTicksLoaded(uint32[] memory num) private view {
+        console.logString("=== Num Ticks Crossed ===");
+        for (uint256 i = 0; i < num.length; i++) {
+            console.logUint(num[i]);
```
zhongeric, line 424: forge console2? not sure what the difference is lol

contracts/libraries/SwapIntention.sol
ConjunctiveNormalForm: yeah, I pulled them from the Router interface to their own file given they are not either router or quoter specific now

test/Quoter.t.sol
ConjunctiveNormalForm: tickMath only defines constants for tick spacing of 1 I think

.solhint.json
ConjunctiveNormalForm: I saw v4-core also checked in solhint config? happy to remove tho

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+            (int128 deltaIn, int128 deltaOut) =
+                !oneForZero ? (curDeltas.amount0(), curDeltas.amount1()) : (curDeltas.amount1(), curDeltas.amount0());
+            deltaAmounts[i - 1] += deltaIn;
+            deltaAmounts[i] += deltaOut;
+
+            prevAmountIn = !oneForZero ? uint128(curDeltas.amount0()) : uint128(curDeltas.amount1());
+            prevCurrencyIn = params.path[i - 1].intermediateCurrency;
+            sqrtPriceX96AfterList[i - 1] = sqrtPriceX96After;
+            initializedTicksLoadedList[i - 1] =
+                PoolTicksCounter.countInitializedTicksLoaded(manager, poolKey, tickBefore, tickAfter);
```
ConjunctiveNormalForm, line 355: can someone double check this is correct? I'm like 99% sure it is (since we work backwards through the path, but store those values in the same order as what would happen for an actual swap)

test/Quoter.t.sol
```diff
@@ -0,0 +1,754 @@
...
+            int128[] memory deltaAmounts,
+            uint160[] memory sqrtPriceX96AfterList,
+            uint32[] memory initializedTicksLoadedList
+        ) = quoter.quoteExactOutput(params);
+
+        assertEq(deltaAmounts[0], 10000);
+        assertEq(deltaAmounts[1], 0);
+        assertEq(deltaAmounts[2], -9745);
+        assertEq(sqrtPriceX96AfterList[0], 78461888503179331029803316753);
+        assertEq(sqrtPriceX96AfterList[1], 80007838904387594703933785072);
```
ConjunctiveNormalForm, line 617: lil sus that this reverse order of the equivalent v3 quoter test: https://github.com/Uniswap/v3-periphery/blob/697c2474757ea89fec12a4e6db16a574fe259610/test/QuoterV2.spec.ts#L528C9-L528C80

.solhint.json
marktoda: + remove, we don't use solhint anymore anyways - just forge fmt

contracts/interfaces/IQuoter.sol
```diff
@@ -0,0 +1,72 @@
...
+    /// poolKey The key for identifying a V4 pool
+    /// zeroForOne If the swap is from currency0 to currency1
+    /// recipient The indented recipient of the output tokens
+    /// amountIn The desired input amount
+    /// sqrtPriceLimitX96 The price limit of the pool that cannot be exceeded by the swap
+    /// hookData arbitrary hookData to pass into the associated hooks
+    /// @return deltaAmounts Delta amounts resulted from the swap
+    /// @return sqrtPriceX96After The sqrt price of the pool after the swap
+    /// @return initializedTicksLoaded The number of initialized ticks that the swap loaded
+
```
marktoda, line 35: nit: rm

contracts/interfaces/IQuoter.sol
```diff
@@ -0,0 +1,72 @@
...
+    /// zeroForOne If the swap is from currency0 to currency1
+    /// recipient The indented recipient of the output tokens
+    /// amountOut The desired input amount
+    /// sqrtPriceLimitX96 The price limit of the pool that cannot be exceeded by the swap
+    /// hookData arbitrary hookData to pass into the associated hooks
+    /// @return deltaAmounts Delta amounts resulted from the swap
+    /// @return sqrtPriceX96After The sqrt price of the pool after the swap
+    /// @return initializedTicksLoaded The number of initialized ticks that the swap loaded
+
+    function quoteExactOutputSingle(ExactOutputSingleParams calldata params)
```
marktoda, line 69: still TODO: exact output multiple?

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.20;
+
+import "../libraries/SwapIntention.sol";
```
marktoda, line 4: nit: local imports last

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+            uint32[] memory initializedTicksLoadedList
+        )
+    {
+        try manager.lock(abi.encode(SwapInfo(SwapType.ExactInput, abi.encode(params)))) {}
+        catch (bytes memory reason) {
+            return _handleRevert(reason);
+        }
+    }
+
+    function quoteExactInputBatch(ExactInputSingleBatchParams memory params)
```
marktoda, line 57: we can do batch quotes via `multicall` instead for a cleaner interface, unless there are important benefits here that I'm missing?

contracts/lens/Quoter.sol
marktoda: Still think it's worth looking into the self call pattern here for cleaner branching

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+        SwapInfo memory swapInfo = abi.decode(encodedSwapIntention, (SwapInfo));
+
+        if (swapInfo.swapType == SwapType.ExactInputSingle) {
+            (BalanceDelta deltas, uint160 sqrtPriceX96After, int24 tickAfter) =
+                _quoteExactInputSingle(abi.decode(swapInfo.params, (ExactInputSingleParams)));
+
+            bytes memory result = abi.encode(deltas, sqrtPriceX96After, tickAfter);
+            assembly {
+                revert(add(0x20, result), mload(result))
+            }
```
marktoda, line 174: this branch is repeated over and over, can factor out `result` and revert at the end?

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+        if (reason.length < 96) {
+            // function selector + length of bytes as uint256 + min length of revert reason padded to multiple of 32 bytes
+            if (reason.length < 68) {
+                revert UnexpectedRevertBytes();
+            }
+            assembly {
+                reason := add(reason, 0x04)
+            }
+            revert(abi.decode(reason, (string)));
+        }
```
marktoda, line 220: constants for the magic #s and better comments to explain where they come from would be great

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+        return reason;
+    }
+
+    function _handleRevertSingle(bytes memory reason, PoolKey memory poolKey)
```
marktoda, line 224: nit natspecs even on private fns

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+        return _quoteExact(
+            params.poolKey,
+            params.zeroForOne,
+            -int256(uint256(params.amountOut)),
+            params.sqrtPriceLimitX96,
+            params.hookData
+        );
+    }
+
+    function _quoteExact(
```
marktoda, line 372: _quoteExact is confusing for me here -- this is just doing a swap, so think just `_swap` is probably more accurate

contracts/libraries/SwapIntention.sol
marktoda: this library should go alongside PathKey in PathKe.sol and be called something more like PathKeyLib

contracts/libraries/SwapIntention.sol
marktoda: My feeling is that they're still mostly _router_ specific structs, but quoter happens to also use them since quoter is trying to adhere to router interface for easy integrations. Also since router is in a separate repo, now these are duplicated and liable to diverge? I'd prefer we just import them

marktoda (changes requested):

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+import "forge-std/console2.sol";
+import {Hooks} from "@uniswap/v4-core/contracts/libraries/Hooks.sol";
+import {TickMath} from "@uniswap/v4-core/contracts/libraries/TickMath.sol";
+import {IHooks} from "@uniswap/v4-core/contracts/interfaces/IHooks.sol";
+import {IPoolManager} from "@uniswap/v4-core/contracts/interfaces/IPoolManager.sol";
+import {BalanceDelta} from "@uniswap/v4-core/contracts/types/BalanceDelta.sol";
+import {Currency} from "@uniswap/v4-core/contracts/types/Currency.sol";
+import {PoolKey} from "@uniswap/v4-core/contracts/types/PoolKey.sol";
+import {PoolIdLibrary} from "@uniswap/v4-core/contracts/types/PoolId.sol";
+import "../libraries/SwapIntention.sol";
```
zhongeric, line 13: nit: think we prefer named imports

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+        }
+        assembly {
+            revert(add(0x20, result), mload(result))
+        }
+    }
+
+    function validateRevertReason(bytes memory reason) private pure returns (bytes memory) {
+        if (reason.length < MINIMUM_VALID_REASON_LENGTH) {
+            if (reason.length < MINIMUM_REASON_LENGTH) {
+                revert UnexpectedRevertBytes();
```
zhongeric, line 214: I'd double check that this works if we are reverting with custom errors which are length 4

contracts/libraries/PathKey.sol
```diff
@@ -0,0 +1,14 @@
+//SPDX-License-Identifier: UNLICENSED
+
+pragma solidity ^0.8.20;
+
+import {Currency} from "@uniswap/v4-core/contracts/types/Currency.sol";
+import {IHooks} from "@uniswap/v4-core/contracts/interfaces/IHooks.sol";
+
+struct PathKey {
+    Currency intermediateCurrency;
```
zhongeric, line 9: This concept of intermediateCurrency is a bit confusing since you have to know the prevToken which would require special casing when this is the first hop in a multi hop route and the final one. I wonder if it would be easier to make this `PathKey` more similar to the `PoolKey` type with the addition of `hookData`

zhongeric: And as long as `zeroForOne` is right and the pools are in the right order it should work out?

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+    using Hooks for IHooks;
+
+    /// @dev Transient storage variable used to check a safety condition in exact output swaps.
+    uint256 private amountOutCached;
+
+    // v4 Singleton contract
+    IPoolManager public immutable manager;
+
+    /// @dev function selector + length of bytes as uint256 + min length of revert reason padded to multiple of 32 bytes
+    uint256 internal constant MINIMUM_REASON_LENGTH = 68;
```
snreynolds, line 37: Technically I think the min is 4 if it reverts with InvalidLockAcquiredSender or InvalidQuoteType

snreynolds: Right now if that happens we will just bubble up UnexpectedRevertBytes. So would change this and then also add tests on these revert paths

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+
+    /// @dev int128[2] + sqrtPriceX96After padded to 32bytes + intializeTicksLoaded padded to 32bytes
+    uint256 internal constant MINIMUM_VALID_REASON_LENGTH = 96;
```
snreynolds, line 41: nit but maybe reason -> response?

contracts/lens/Quoter.sol
snreynolds: also this is minimum in the unhappy case right? so there is minimum valid and minimum invalid?

contracts/lens/Quoter.sol
snreynolds: ++ for multicall

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+import {IPoolManager} from "@uniswap/v4-core/contracts/interfaces/IPoolManager.sol";
+import {BalanceDelta} from "@uniswap/v4-core/contracts/types/BalanceDelta.sol";
+import {Currency} from "@uniswap/v4-core/contracts/types/Currency.sol";
+import {PoolKey} from "@uniswap/v4-core/contracts/types/PoolKey.sol";
+import {PoolIdLibrary} from "@uniswap/v4-core/contracts/types/PoolId.sol";
+import "../libraries/SwapIntention.sol";
+import {IQuoter} from "../interfaces/IQuoter.sol";
+import {PoolTicksCounter} from "../libraries/PoolTicksCounter.sol";
+
+contract Quoter is IQuoter {
```
snreynolds, line 26: This should inherit ILockCallback

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,396 @@
...
+    using PoolIdLibrary for PoolKey;
+    using Hooks for IHooks;
+
+    /// @dev Transient storage variable used to check a safety condition in exact output swaps.
```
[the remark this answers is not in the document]
ConjunctiveNormalForm: the latter - this is an artifact from v3 Quoter. You are right that I should change how this is worded

contracts/libraries/PathKey.sol
ConjunctiveNormalForm: True although first/last hop from token wouldn't be in the path so it's a bit cleaner. I kinda lean toward leaving this as is just because we already got it working, unless there are other reasons? Curious what people think

contracts/lens/Quoter.sol
ConjunctiveNormalForm: added `uint256 internal constant MINIMUM_CUSTOM_ERROR_LENGTH = 4;`

contracts/libraries/SwapIntention.sol
ConjunctiveNormalForm: fwiw router is in the same repo and will most likely be reusing the same structs so I still lean toward not putting them in IRouter? wonder what others think

contracts/lens/Quoter.sol
ConjunctiveNormalForm: removed

contracts/lens/Quoter.sol
ConjunctiveNormalForm: good callout, added another branch checking that codition

contracts/libraries/SwapIntention.sol
ConjunctiveNormalForm: ok I think Router needs an extra field `ExactInputMaximum/ExactOutputMinimum` which makes this tricker... Maybe should just have two sets of structs for Quoter and Router respectively (this is what v3 did)

contracts/libraries/PathKey.sol
zhongeric: yea fair enough

contracts/interfaces/IQuoter.sol
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.20;
+
+import "../libraries/SwapIntention.sol";
```
zhongeric, line 4: nit: can we do named imports?

contracts/libraries/SwapIntention.sol
```diff
@@ -0,0 +1,55 @@
...
+
+struct ExactOutputParams {
+    Currency currencyOut;
+    PathKey[] path;
+    address recipient;
+    uint128 amountOut;
+    uint160 sqrtPriceLimitX96;
+}
+
+library SwapIntention {}
```
zhongeric, line 55: nothing in this library?

contracts/libraries/SwapIntention.sol
zhongeric: SwapParameters? Yeah not super sure w.r.t the question of where these should live for the quoter/router

ConjunctiveNormalForm: yeah I like `SwapParameters` better, want to avoid `intent` so bad haha

ConjunctiveNormalForm: okay nvm friends, decided to move those structs into IQuoter interface and not share with Router (it needs a slightly different version of those structs) 🫠

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,320 @@
...
+    /// @dev quote an ExactInput swap on a pool, then revert with the result
+    function _quoteExactInputSingle(QuoteExactInputSingleParams memory params) public selfOnly returns (bytes memory) {
+        (BalanceDelta deltas, uint160 sqrtPriceX96After, int24 tickAfter) = _swap(
+            params.poolKey,
+            params.zeroForOne,
+            int256(int128(params.amountIn)),
+            params.sqrtPriceLimitX96,
+            params.hookData
+        );
+        bytes memory result = abi.encode(deltas, sqrtPriceX96After, tickAfter);
```
snreynolds, line 227: in the _quoteExactInput logic we directly encode the values we want in _handleRevert. Why don't we do the same for _quoteExactInputSingle / _handleRevertSingle? ie push the initializedTicksLoaded call and encoding the deltaAmounts as an array into _quoteExactInputSingle ?

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,320 @@
...
+
+        int128[] memory deltaAmounts = new int128[](pathLength + 1);
+        uint160[] memory sqrtPriceX96AfterList = new uint160[](pathLength);
+        uint32[] memory initializedTicksLoadedList = new uint32[](pathLength);
+        Currency prevCurrencyOut;
+        uint128 prevAmountOut;
+
+        for (uint256 i = 0; i < pathLength; i++) {
+            (PoolKey memory poolKey, bool zeroForOne) =
+                PathKeyLib.getPoolAndSwapDirection(params.path[i], i == 0 ? params.currencyIn : prevCurrencyOut);
```
snreynolds, line 190: very up to you but I like the library syntax where you can do 

pathKey.getPoolAndSwapDirection(currencyIn)

and you define above `using PathKeyLib for PathKey`.. but just syntactical

contracts/lens/Quoter.sol
snreynolds: whats the concern?

snreynolds (changes requested):

contracts/interfaces/IQuoter.sol
```diff
@@ -0,0 +1,122 @@
...
+        address recipient;
+        uint128 amountOut;
+        uint160 sqrtPriceLimitX96;
+    }
+
+    /// @notice Returns the delta amounts for a given exact input swap of a single pool
+    /// @param params The params for the quote, encoded as `QuoteExactInputSingleParams`
+    /// poolKey The key for identifying a V4 pool
+    /// zeroForOne If the swap is from currency0 to currency1
+    /// recipient The indented recipient of the output tokens
```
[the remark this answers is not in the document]
ConjunctiveNormalForm: 🤦‍♂️

contracts/lens/Quoter.sol
ConjunctiveNormalForm: idt that's possible because there is no `PathKey` instance here?

test/Quoter.t.sol
ConjunctiveNormalForm: now I think about it, this behavior is fine. Take `deltaAmounts` for example, trade path is 0 -> 2 -> 1 so `deltaAmounts[0]` is positive (into the pool) while `deltaAmounts[2]` is negative (taken out of the pool)

contracts/lens/Quoter.sol
snreynolds: Resolved offline, it's different behavior from v3 because v3 requires that the path is passed in backwards whereas here we pass it in normally and iterate backwards

test/Quoter.t.sol
ConjunctiveNormalForm: just to clarify, there is no right or wrong here just different behavior and I the ordering I have more

contracts/lens/Quoter.sol
snreynolds: isn't params.path[i] the PathKey?

snreynolds (dismissed):
changes look good, left 1 last comment but its syntax so up to you

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,325 @@
...
+    function quoteExactOutput(QuoteExactParams memory params)
+        public
+        override
+        returns (
+            int128[] memory deltaAmounts,
+            uint160[] memory sqrtPriceX96AfterList,
+            uint32[] memory initializedTicksLoadedList
+        )
+    {
+        try manager.lock(address(this), abi.encodeWithSelector(this._quoteExactOutput.selector, params)) {}
```
marktoda, line 99: why doesnt this one need amountOutCached?

contracts/lens/Quoter.sol
marktoda: yea path[i] is PathKey - agreed with sara on this preference

marktoda (dismissed):

contracts/lens/Quoter.sol
ConjunctiveNormalForm: ohh true, moved amountOutCached into the inner call (_quote...) so quoteExactOutput also uses it

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,270 @@
...
+        (slot0.sqrtPriceX96, slot0.tick,,) = poolManager.getSlot0(id);
+
+        return slot0;
+    }
+
+    function validateRevertReason(bytes memory reason) private pure returns (bytes memory) {
+        if (reason.length < 96) {
+            // function selector + length of bytes as uint256 + min length of revert reason padded to multiple of 32 bytes
+            if (reason.length < 68) {
+                revert UnexpectedRevertBytes();
```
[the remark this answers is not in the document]
zhongeric: todo?

contracts/lens/Quoter.sol
```diff
@@ -0,0 +1,333 @@
...
+    {
+        try manager.lock(address(this), abi.encodeWithSelector(this._quoteExactInputSingle.selector, params)) {}
+        catch (bytes memory reason) {
+            return _handleRevertSingle(reason);
+        }
+    }
+
+    /// @inheritdoc IQuoter
+    function quoteExactInput(QuoteExactParams memory params)
+        external
```
zhongeric, line 62: this should be public right? if we are doing the batched self call

ConjunctiveNormalForm: I think you are right but would rather change in the separate PR that uses batched self call

contracts/lens/Quoter.sol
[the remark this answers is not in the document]
ConjunctiveNormalForm: done!

contracts/lens/Quoter.sol
```diff
@@ -124,13 +124,13 @@ contract Quoter is IQuoter, ILockCallback {
     function validateRevertReason(bytes memory reason) private pure returns (bytes memory) {
         if (reason.length < MINIMUM_VALID_RESPONSE_LENGTH) {
             //if InvalidLockAcquiredSender()
-            if (reason.length == MINIMUM_CUSTOM_ERROR_LENGTH) {
+            if (reason.length <= MINIMUM_CUSTOM_ERROR_LENGTH) {
```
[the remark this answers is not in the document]
snreynolds: Can't you revert with no function selector? Or like no data?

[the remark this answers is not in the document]
snreynolds: It does seem odd though to special case this? We do the same logic on line 135 no?

snreynolds (approved):
solid. i think this is the most well written periphery contract atm 👏 👏 👏 👏

PR #26, opened 2023-06-15
Fix `v4-core` submodule URL

apoorvlathey:
## Related Issue
Fixes #25 

## Description of changes
Replace url in `.gitmodules` from `git@github.com:Uniswap/v4-core.git` to `https://github.com/Uniswap/v4-core`

PR #74, opened 2023-11-17
Update v4 core

marktoda:
## Related Issue

## Description of changes


test/TWAMM.t.sol
```diff
@@ -192,8 +186,8 @@ contract TWAMMTest is Test, Deployers, TokenFixture, GasSnapshot {
 
         assertEq(sellRate0For1, 2e18 / (expiration2 - submitTimestamp2));
         assertEq(sellRate1For0, 3e18 / (expiration2 - submitTimestamp1));
-        assertEq(earningsFactor0For1, 1712020976636017581269515821040000);
-        assertEq(earningsFactor1For0, 1470157410324350030712806974476955);
+        assertEq(earningsFactor0For1, 1636776489931663248324424309240000);
+        assertEq(earningsFactor1For0, 1534530274609724617872321172427618);
```
marktoda, line 190: This is sus - I'm not sure what could have changed to cause these values to be different..

.env
```diff
@@ -0,0 +1,9 @@
+FOUNDRY_FUZZ_SEED=0x4444
+
+if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+        export FOUNDRY_SOLC="./lib/v4-core/bin/solc-static-linux"
+elif [[ "$OSTYPE" == "darwin"* ]]; then
+        export FOUNDRY_SOLC="./lib/v4-core/bin/solc-mac"
+fi
+
+# FOUNDRY_SOLC="./bin/solc-static-linux"
```
hensha256, line 9: remove comment?

test/FullRange.t.sol
```diff
@@ -80,15 +77,17 @@ contract TestFullRange is Test, Deployers, GasSnapshot {
...
-    PoolModifyPositionTest modifyPositionRouter;
-    PoolSwapTest swapRouter;
-
     function setUp() public {
-        token0 = new MockERC20("TestA", "A", 18, 2 ** 128);
-        token1 = new MockERC20("TestB", "B", 18, 2 ** 128);
-        token2 = new MockERC20("TestC", "C", 18, 2 ** 128);
+        deployFreshManagerAndRouters();
+        router = new HookEnabledSwapRouter(manager);
+        (currency0, currency1) = deployMintAndApprove2Currencies();
```
hensha256, line 83: these seem to be unused, and then instead token0 token1 are deployed below?
I'd rather remove the `tokenX` variables and use global `currencyX` for consistency

test/FullRange.t.sol
```diff
@@ -102,17 +101,14 @@ contract TestFullRange is Test, Deployers, GasSnapshot {
         keyWithLiq = createPoolKey(token0, token2);
```
hensha256, line 101: where is `createPoolKey` from? we dont have that function in core 🤔

test/FullRange.t.sol
```diff
@@ -102,17 +101,14 @@ contract TestFullRange is Test, Deployers, GasSnapshot {
...
         token2.approve(address(fullRange), type(uint256).max);
-        token0.approve(address(swapRouter), type(uint256).max);
-        token1.approve(address(swapRouter), type(uint256).max);
-        token2.approve(address(swapRouter), type(uint256).max);
+        token0.approve(address(router), type(uint256).max);
+        token1.approve(address(router), type(uint256).max);
+        token2.approve(address(router), type(uint256).max);
 
-        manager.initialize(keyWithLiq, SQRT_RATIO_1_1, ZERO_BYTES);
+        initPool(keyWithLiq.currency0, keyWithLiq.currency1, fullRange, 3000, SQRT_RATIO_1_1, ZERO_BYTES);
```
hensha256, line 111: this `fullRange` hook address doesnt seem to be in the `keyWithLiq` key before?

test/TWAMM.t.sol
```diff
@@ -76,9 +76,6 @@ contract TWAMMTest is Test, Deployers, TokenFixture, GasSnapshot {
             }
         }
 
-        modifyPositionRouter = new PoolModifyPositionTest(IPoolManager(address(manager)));
-        swapRouter = new PoolSwapTest(IPoolManager(address(manager)));
-
         poolKey = PoolKey(Currency.wrap(address(token0)), Currency.wrap(address(token1)), 3000, 60, twamm);
```
hensha256, line 79: `Currency.wrap(address(token0))` is `currency0`

hensha256 (commented):
left some thoughts

test/TWAMM.t.sol
hensha256: also these lines can reduce to `initPool`

test/FullRange.t.sol
marktoda: changing to use `deployTokens` and remove unused. The currencyX are annoying here because:
- all the tests are written around MockERC20 and dont want to have to do Currency.unwrap everywhere
- we have 3, not just 2 so can't use the globals for all :(

test/FullRange.t.sol
marktoda: it's defined in this file, it was here when i got here

test/FullRange.t.sol
marktoda: yeah it is... in createPoolKey

test/TWAMM.t.sol
snreynolds: 👀

saucepoint:
any chance we can include a change like this https://github.com/Uniswap/v4-periphery/pull/26 for this PR?

Seeing some devs run into problems regarding this

marktoda:
> any chance we can include a change like this https://github.com/Uniswap/v4-periphery/pull/26 for this PR?

snreynolds:
Not sure why linting is failing ❓

contracts/hooks/examples/FullRange.sol
```diff
@@ -295,7 +299,7 @@ contract FullRange is BaseHook, ILockCallback {
         pool.hasAccruedFees = false;
     }
 
-    function lockAcquired(bytes calldata rawData)
+    function lockAcquired(address, /*sender*/ bytes calldata rawData)
```
hensha256, line 302: I think this contract now has a critical vuln in it.
Now that anyone can call `manager.lock()` with this hook as the locker and they as the lockCaller, they can pass in any data that they want. This means that `data.sender` can no longer be trusted. I think someone could call and get thisfunction to remove liquidity from the pool, and then send it to themselves.
I think we should no longer be using `data.sender` as a model, but instead be using `lockCaller` to find the sender

hensha256: also separately i think some of the functions in this hook need the `onlyPoolManager` modifier - e.g. `beforeSwap` doesnt have it?

hensha256 (changes requested):

contracts/hooks/examples/FullRange.sol
snreynolds: yeah good point. I think we can check that sender == address(this) and that the lock was started by the hook itself

contracts/libraries/PoolGetters.sol
```diff
@@ -62,7 +63,7 @@ library PoolGetters {
                 // all the 1s at or to the right of the current bitPos
                 uint256 mask = (1 << bitPos) - 1 + (1 << bitPos);
                 // uint256 masked = self[wordPos] & mask;
-                uint256 masked = getTickBitmapAtWord(poolManager, poolId, wordPos) & mask;
```
ConjunctiveNormalForm, line 65: can probably remove `getTickBitmapAtWord` btw

marktoda (commented):
cant approve my own PR :D think fine to merge - we def have some cleanup to do here but can be incremental

PR #75, opened 2023-11-21
Add base contracts and interfaces used in PositionManager

tinaszheng:
## Description of changes
This PR adds base contracts and interfaces that are used in PositionManagerV4 [here](https://github.com/Uniswap/v4-periphery/pull/76). But these are mostly copied from v3-periphery with a few changes (ill document as comments).

### DIFF
--- a/contracts/base/Multicall.sol
+++ b/contracts/base/Multicall.sol
@@ -1,0 +1,33 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.19;
+
+import {IMulticall} from "../interfaces/IMulticall.sol";
+
+/// @title Multicall
+/// @notice Enables calling multiple methods in a single call to the contract
+abstract contract Multicall is IMulticall {
+    /// @inheritdoc IMulticall
+    function multicall(bytes[] calldata data) public payable override returns (bytes[] memory results) {
+        results = new bytes[](data.length);
+        for (uint256 i = 0; i < data.length; i++) {
+            (bool success, bytes memory result) = address(this).delegatecall(data[i]);
+
+            if (!success) {
+                // handle custom errors
+                if (result.length == 4) {
+                    assembly {
+                        revert(add(result, 0x20), mload(result))
+                    }
+                }
+                // Next 5 lines from https://ethereum.stackexchange.com/a/83577
+                if (result.length < 68) revert();
+                assembly {
+                    result := add(result, 0x04)
+                }
+                revert(abi.decode(result, (string)));
+            }
+
+            results[i] = result;
+        }
+    }
+}
--- a/contracts/base/PeripheryPayments.sol
+++ b/contracts/base/PeripheryPayments.sol
@@ -1,0 +1,41 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+import {ERC20} from "solmate/tokens/ERC20.sol";
+import {Currency, CurrencyLibrary} from "@uniswap/v4-core/contracts/types/Currency.sol";
+import {SafeTransferLib} from "solmate/utils/SafeTransferLib.sol";
+import {IPeripheryPayments} from "../interfaces/IPeripheryPayments.sol";
+
+abstract contract PeripheryPayments is IPeripheryPayments {
+    using CurrencyLibrary for Currency;
+    using SafeTransferLib for address;
+    using SafeTransferLib for ERC20;
+
+    error InsufficientToken();
+    error NativeTokenTransferFrom();
+
+    /// @inheritdoc IPeripheryPayments
+    function sweepToken(Currency currency, uint256 amountMinimum, address recipient) public payable override {
+        uint256 balanceCurrency = currency.balanceOfSelf();
+        if (balanceCurrency < amountMinimum) revert InsufficientToken();
+
+        if (balanceCurrency > 0) {
+            currency.transfer(recipient, balanceCurrency);
+        }
+    }
+
+    /// @param currency The currency to pay
+    /// @param payer The entity that must pay
+    /// @param recipient The entity that will receive payment
+    /// @param value The amount to pay
+    function pay(Currency currency, address payer, address recipient, uint256 value) internal {
+        if (payer == address(this)) {
+            // pay with tokens already in the contract (for the exact input multihop case)
+            currency.transfer(recipient, value);
+        } else {
+            if (currency.isNative()) revert NativeTokenTransferFrom();
+            // pull payment
+            ERC20(Currency.unwrap(currency)).safeTransferFrom(payer, recipient, value);
+        }
+    }
+}
--- a/contracts/base/SelfPermit.sol
+++ b/contracts/base/SelfPermit.sol
@@ -1,0 +1,52 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {IERC20Permit} from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Permit.sol";
+
+import {IERC20PermitAllowed} from "../interfaces/external/IERC20PermitAllowed.sol";
+import {ISelfPermit} from "../interfaces/ISelfPermit.sol";
+
+/// @title Self Permit
+/// @notice Functionality to call permit on any EIP-2612-compliant token for use in the route
+/// @dev These functions are expected to be embedded in multicalls to allow EOAs to approve a contract and call a function
+/// that requires an approval in a single transaction.
+abstract contract SelfPermit is ISelfPermit {
+    /// @inheritdoc ISelfPermit
+    function selfPermit(address token, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
+        public
+        payable
+        override
+    {
+        IERC20Permit(token).permit(msg.sender, address(this), value, deadline, v, r, s);
+    }
+
+    /// @inheritdoc ISelfPermit
+    function selfPermitIfNecessary(address token, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s)
+        external
+        payable
+        override
+    {
+        if (IERC20(token).allowance(msg.sender, address(this)) < value) selfPermit(token, value, deadline, v, r, s);
+    }
+
+    /// @inheritdoc ISelfPermit
+    function selfPermitAllowed(address token, uint256 nonce, uint256 expiry, uint8 v, bytes32 r, bytes32 s)
+        public
+        payable
+        override
+    {
+        IERC20PermitAllowed(token).permit(msg.sender, address(this), nonce, expiry, true, v, r, s);
+    }
+
+    /// @inheritdoc ISelfPermit
+    function selfPermitAllowedIfNecessary(address token, uint256 nonce, uint256 expiry, uint8 v, bytes32 r, bytes32 s)
+        external
+        payable
+        override
+    {
+        if (IERC20(token).allowance(msg.sender, address(this)) < type(uint256).max) {
+            selfPermitAllowed(token, nonce, expiry, v, r, s);
+        }
+    }
+}
--- a/contracts/interfaces/IPeripheryPayments.sol
+++ b/contracts/interfaces/IPeripheryPayments.sol
@@ -1,0 +1,17 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+import {Currency} from "@uniswap/v4-core/contracts/types/Currency.sol";
+
+/// @title Periphery Payments
+/// @notice Functions to ease deposits and withdrawals of ETH
+interface IPeripheryPayments {
+    // TODO: figure out if we still need unwrapWETH9 from v3?
+
+    /// @notice Transfers the full amount of a token held by this contract to recipient
+    /// @dev The amountMinimum parameter prevents malicious contracts from stealing the token from users
+    /// @param currency The contract address of the token which will be transferred to `recipient`
+    /// @param amountMinimum The minimum amount of token required for a transfer
+    /// @param recipient The destination address of the token
+    function sweepToken(Currency currency, uint256 amountMinimum, address recipient) external payable;
+}
--- a/contracts/base/PeripheryValidation.sol
+++ b/contracts/base/PeripheryValidation.sol
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+abstract contract PeripheryValidation {
+    error TransactionTooOld();
+
+    modifier checkDeadline(uint256 deadline) {
+        if (block.timestamp > deadline) revert TransactionTooOld();
+        _;
+    }
+}
--- a/contracts/interfaces/IMulticall.sol
+++ b/contracts/interfaces/IMulticall.sol
@@ -1,0 +1,12 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+/// @title Multicall interface
+/// @notice Enables calling multiple methods in a single call to the contract
+interface IMulticall {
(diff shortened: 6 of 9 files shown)


contracts/base/PeripheryPayments.sol
```diff
@@ -0,0 +1,40 @@
...
+import "../libraries/TransferHelper.sol";
+
+abstract contract PeripheryPayments is IPeripheryPayments {
+    /// @inheritdoc IPeripheryPayments
+    function sweepToken(address token, uint256 amountMinimum, address recipient) public payable override {
+        uint256 balanceToken = IERC20(token).balanceOf(address(this));
+        require(balanceToken >= amountMinimum, "Insufficient token");
+
+        if (balanceToken > 0) {
+            TransferHelper.safeTransfer(IERC20Minimal(token), recipient, balanceToken);
```
tinaszheng, line 18: added the IERC20Minimal casting here

contracts/base/PeripheryPayments.sol
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    /// @param token The token to pay
+    /// @param payer The entity that must pay
+    /// @param recipient The entity that will receive payment
+    /// @param value The amount to pay
+    function pay(address token, address payer, address recipient, uint256 value) internal {
+        if (payer == address(this)) {
+            // pay with tokens already in the contract (for the exact input multihop case)
+            TransferHelper.safeTransfer(IERC20Minimal(token), recipient, value);
```
tinaszheng, line 34: added the IERC20Minimal casting here and on line 37

contracts/interfaces/IPeripheryPayments.sol
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+/// @title Periphery Payments
+/// @notice Functions to ease deposits and withdrawals of ETH
+interface IPeripheryPayments {
+    // TODO: figure out if we still need unwrapWETH9 from v3?
```
tinaszheng, line 7: add comment here and took out `unwrapWETH9` from v3

contracts/libraries/ChainId.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+/// @title Function for getting the current chain ID
+library ChainId {
+    /// @dev Gets the current chain ID
+    /// @return chainId The current chain ID
+    function get() internal view returns (uint256 chainId) {
```
tinaszheng, line 8: changed `pure` to `view`

contracts/base/ERC721Permit.sol
```diff
@@ -0,0 +1,77 @@
...
+            abi.encodePacked(
+                "\x19\x01",
+                DOMAIN_SEPARATOR(),
+                keccak256(abi.encode(PERMIT_TYPEHASH, spender, tokenId, _getAndIncrementNonce(tokenId), deadline))
+            )
+        );
+        address owner = ownerOf(tokenId);
+        require(spender != owner, "ERC721Permit: approval to current owner");
+
+        if (owner.code.length > 0) {
```
tinaszheng, line 67: changed this from `Address(owner).isContract` because `.isContract` was removed from the Address library

contracts/libraries/TransferHelper.sol
```diff
@@ -51,4 +51,13 @@ library TransferHelper {
             address(token).call(abi.encodeWithSelector(IERC20Minimal.transferFrom.selector, from, to, value));
         require(success && (data.length == 0 || abi.decode(data, (bool))), "STF");
     }
+
+    /// @notice Transfers ETH to the recipient address
+    /// @dev Fails with `STE`
+    /// @param to The destination of the transfer
+    /// @param value The value to be transferred
+    function safeTransferETH(address to, uint256 value) internal {
```
tinaszheng, line 59: added this because it's used in `PeripheryPayments`

contracts/base/BlockTimestamp.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+/// @title Function for getting block timestamp
+/// @dev Base contract that is overridden for tests
+abstract contract BlockTimestamp {
+    /// @dev Method that exists purely to be overridden for tests
+    /// @return The current block timestamp
+    function _blockTimestamp() internal view virtual returns (uint256) {
```
hensha256, line 9: I think now that we work in forge this shouldnt be needed?
you can do `vm.warp(timestamp)` in a test, and forge will set the current timestamp to that :)

contracts/base/ERC721Permit.sol
hensha256: I think they removed it because its not technically a secure check (its possible for a contract to spoof itself as a non-contract with this check).
That being said I think we're ok for this situation to keep using the check

contracts/base/ERC721Permit.sol
```diff
@@ -0,0 +1,77 @@
...
+            require(IERC1271(owner).isValidSignature(digest, abi.encodePacked(r, s, v)) == 0x1626ba7e, "Unauthorized");
```
hensha256, line 68: I'd change any 
```require(bool, error string)```
into custom errors (its cheaper from a gas PoV and neater
```
error Unauthorized();
if (!bool) revert Unauthorized();
```

Comment (hensha256) [reply contracts/base/ERC721Permit.sol:68]: it was introduced in 0.8.4 - you can see more about it [here](https://soliditylang.org/blog/2021/04/21/custom-errors/)

File: contracts/base/PeripheryPayments.sol (hunk 1 above)
Comment (hensha256) [reply contracts/base/PeripheryPayments.sol:18]: Probably neater to us "using"
So at the top of the contract you write
```
using TransferHelper for IERC20Minimal
```

and then this line can be
```
token.safeTransfer(recipient, balanceToken
```

File: contracts/base/PeripheryPayments.sol [hunk 9]
```diff
@@ -0,0 +1,40 @@
...
+        require(balanceToken >= amountMinimum, "Insufficient token");
+
+        if (balanceToken > 0) {
+            TransferHelper.safeTransfer(IERC20Minimal(token), recipient, balanceToken);
+        }
+    }
+
+    /// @inheritdoc IPeripheryPayments
+    function refundETH() external payable override {
+        if (address(this).balance > 0) TransferHelper.safeTransferETH(msg.sender, address(this).balance);
```
hensha256, line 24: you can add another line
```
using TransferHelper for address
```
and do the same for this line too

contracts/base/ERC721Permit.sol
```diff
@@ -0,0 +1,77 @@
...
+
+        bytes32 digest = keccak256(
+            abi.encodePacked(
+                "\x19\x01",
+                DOMAIN_SEPARATOR(),
+                keccak256(abi.encode(PERMIT_TYPEHASH, spender, tokenId, _getAndIncrementNonce(tokenId), deadline))
+            )
+        );
+        address owner = ownerOf(tokenId);
+        require(spender != owner, "ERC721Permit: approval to current owner");
```
zhongeric, line 65: Ik mostly copied over from V3 but we should prob use custom errors now that we have them

contracts/base/ERC721Permit.sol
zhongeric, line 68: nit: magic 4 byte selector can be a constant with a comment

hensha256 (commented):
Added some initial thoughts!

contracts/base/ERC721Permit.sol
```diff
@@ -0,0 +1,77 @@
...
+
+        if (owner.code.length > 0) {
+            require(IERC1271(owner).isValidSignature(digest, abi.encodePacked(r, s, v)) == 0x1626ba7e, "Unauthorized");
+        } else {
+            address recoveredAddress = ecrecover(digest, v, r, s);
+            require(recoveredAddress != address(0), "Invalid signature");
+            require(recoveredAddress == owner, "Unauthorized");
```
zhongeric, line 72: don't think you need the address(0) check because it would imply that someone has created a sig from the 0 address but maybe there's something with ecrecover that im missing

contracts/base/Multicall.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.19;
+
+import {IMulticall} from "../interfaces/IMulticall.sol";
+
+/// @title Multicall
+/// @notice Enables calling multiple methods in a single call to the contract
+abstract contract Multicall is IMulticall {
```
zhongeric, line 8: Ooo I've actually implemented another version of this multicall in this PR https://github.com/Uniswap/UniswapX/pull/202/files#diff-f83ec168220cc728379adbb931f4151db29f407d9cfa1130403782bcc251a6bf but added support for bubbling up custom errors

contracts/base/PeripheryPayments.sol
```diff
@@ -0,0 +1,40 @@
...
+        uint256 balanceToken = IERC20(token).balanceOf(address(this));
+        require(balanceToken >= amountMinimum, "Insufficient token");
+
+        if (balanceToken > 0) {
+            TransferHelper.safeTransfer(IERC20Minimal(token), recipient, balanceToken);
+        }
+    }
+
+    /// @inheritdoc IPeripheryPayments
+    function refundETH() external payable override {
```
zhongeric, line 23: nit: why payable here?

contracts/base/ERC721Permit.sol
hensha256: Oh agreed! But doesnt even need to be a constant
you could change `0x1626ba7e` to `IER1271.isValidSignature.selector`

contracts/base/ERC721Permit.sol
hensha256: ecrecover returns address(0) as its error - like if the signature is invalid or formatted incorrectly

zhongeric: oh gotcha, though it would be caught by the owner check below - but maybe its good to special case this case?

tinaszheng: yeah i think its more descriptive of the error - passing in an invalid signature format vs passing in a well-formated signature that decrypts to the wrong address seem like separate cases

contracts/base/Multicall.sol
tinaszheng: ah ty is it just adding these few lines?:

```
if (!success) {
 // handle custom errors
  if (result.length == 4) {
      assembly {
          revert(add(result, 0x20), mload(result))
      }
  }

...
}
```

zhongeric: ya should be

contracts/base/PeripheryPayments.sol
marktoda: yea thats weird - we should look in v3 sdk to see if we ever use this with value. since it just sends back to msg.sender anyways it seems useless

contracts/base/PeripheryPayments.sol
```diff
@@ -0,0 +1,45 @@
...
+import "../libraries/TransferHelper.sol";
+
+using TransferHelper for address;
+using TransferHelper for IERC20Minimal;
+
+error InsufficientToken();
+
+abstract contract PeripheryPayments is IPeripheryPayments {
+    /// @inheritdoc IPeripheryPayments
+    function sweepToken(address token, uint256 amountMinimum, address recipient) public payable override {
```
marktoda, line 18: I think you can merge this and `refundETH` into `sweep(Currency currency, ...)`
and just use CurrencyLibrary nicely here
```
currency.transfer(recipient, currency.balanceOfSelf());
```

marktoda: this abstracts native + ERC20!

contracts/base/PeripheryPayments.sol
```diff
@@ -0,0 +1,45 @@
...
+    /// @inheritdoc IPeripheryPayments
+    function refundETH() external payable override {
+        if (address(this).balance > 0) msg.sender.safeTransferETH(address(this).balance);
+    }
+
+    /// @param token The token to pay
+    /// @param payer The entity that must pay
+    /// @param recipient The entity that will receive payment
+    /// @param value The amount to pay
+    function pay(address token, address payer, address recipient, uint256 value) internal {
```
marktoda, line 36: same here can use CurrencyLibrary and accept Currency as param

contracts/base/PeripheryPayments.sol
marktoda: https://github.com/Uniswap/v4-core/blob/main/src/types/Currency.sol

contracts/libraries/ChainId.sol
marktoda: oo in solidity 0.8 we now have `block.chainid` so this lib is no longer useful

contracts/libraries/TransferHelper.sol
```diff
@@ -3,6 +3,10 @@ pragma solidity ^0.8.15;
 
 import {IERC20Minimal} from "@uniswap/v4-core/contracts/interfaces/external/IERC20Minimal.sol";
 
+error TransferFailed();
+error STF();
+error STE();
```
marktoda, line 8: now that we use custom errors we can make these more descriptive (they were acronyms before to save on bytes!)

maybe `error SafeTransferFailed` and `error SafeTransferEthFailed`

contracts/libraries/TransferHelper.sol
marktoda: consider using solmate's impl https://github.com/transmissions11/solmate/blob/main/src/utils/SafeTransferLib.sol

marktoda (commented):
Looks great!! Some ideas

contracts/libraries/TransferHelper.sol
tinaszheng: oo leaving a TODO here to remove `TransferHelper` to use solmate ver after we migrate twamm to it too

contracts/base/PeripheryPayments.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+import {ERC20} from "solmate/tokens/ERC20.sol";
+import {Currency, CurrencyLibrary} from "@uniswap/v4-core/contracts/types/Currency.sol";
+import {SafeTransferLib} from "solmate/utils/SafeTransferLib.sol";
+import {IPeripheryPayments} from "../interfaces/IPeripheryPayments.sol";
+
+using CurrencyLibrary for Currency;
```
hensha256, line 9: omg i didnt know you could put this above the contract 😱 
we normally put them just inside the contract header like [this](https://github.com/Uniswap/v4-core/blob/06564d33b2fa6095830c914461ee64d34d39c305/src/PoolManager.sol#L27)

contracts/base/PeripheryPayments.sol
```diff
@@ -0,0 +1,41 @@
...
+using SafeTransferLib for address;
+using SafeTransferLib for ERC20;
+
+error InsufficientToken();
```
hensha256, line 13: I'd personally define these inside the contract too? But dont feel as strongly about that... maybe @marktoda has an opinion?

hensha256: maybe this was even copied from before

contracts/base/SelfPermit.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity >=0.5.0;
```
hensha256, line 2: ^0.8.19 like the others?

contracts/base/PeripheryPayments.sol
tinaszheng: https://docs.soliditylang.org/en/v0.8.23/contracts.html#using-for o ya i was reading this and there are some tiny differences to file level vs contract level so if we've only done contract level elsewhere it makes sense to keep that consistent xD

hensha256: wow i had no idea, ty!

marktoda (approved):
LGTM!

PR #76, opened 2023-11-21
Add v4 NonfungiblePositionManager skeleton

tinaszheng:
## Description of changes

This PR adds the base skeleton for `NonfungiblePositionManagerV4` 

So this is mostly copied from v3-periphery but I made a few changes:
- Pool IDs are now just `PoolKeys` (instead of having a separate counter => address mapping)
- I removed references to WETH9 specifically (until we decide on what to do about WETH in v4)
- Added LiquidityManagement to manage lock/lock callback logic
- Added `createAndInitializePoolIfNecessary` directly to the contract and func args take in `PoolKey` instead of ids
- Added `tokenByIndex`, `tokenOfOwnerByIndex`, and `totalSupply` - think these were added to `ERC721Enumerable`?

TODO (next steps):
- [ ] fill in function definitions (everywhere that says `// TODO: implement this`)
- [ ] add tests 
- [ ] Use permit2 version of ERC721Permit?


contracts/base/Multicall.sol
```diff
@@ -0,0 +1,33 @@
...
+        for (uint256 i = 0; i < data.length; i++) {
+            (bool success, bytes memory result) = address(this).delegatecall(data[i]);
+
+            if (!success) {
+                // handle custom errors
+                if (result.length == 4) {
+                    assembly {
+                        revert(add(result, 0x20), mload(result))
+                    }
+                }
```
zhongeric, line 21: oops mark pointed out that custom errors can revert with params so this doesn't necessarily catch all custom errors 😅 , we could probably just revert with the reason again or we can look for a fix

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -0,0 +1,201 @@
...
+        isAuthorizedForToken(params.tokenId)
+        returns (uint256 amount0, uint256 amount1)
+    {
+        // TODO: implement this
+    }
+
+    /// @inheritdoc INonfungiblePositionManagerV4
+    function burn(uint256 tokenId) external payable override isAuthorizedForToken(tokenId) {
+        TokenPosition storage position = _positions[tokenId];
+        if (position.liquidity != 0 || position.tokensOwed0 != 0 || position.tokensOwed1 != 0) revert NotCleared();
```
hensha256, line 163: do we need to make sure theyve collected all their rewards before they can burn? or does this already check that?

contracts/interfaces/IPeripheryPayments.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
+
+import {Currency} from "@uniswap/v4-core/contracts/types/Currency.sol";
+
+/// @title Periphery Payments
+/// @notice Functions to ease deposits and withdrawals of ETH
+interface IPeripheryPayments {
+    // TODO: figure out if we still need unwrapWETH9 from v3?
```
hensha256, line 9: hmmm good question. I think its more of an opinionated decision nowadays... I would lean towards not adding it? People can add liquidity in ETH and remove it in ETH, people can add liquidity in WETH and remove it in WETH. It feels opinionated now to allow the WETH people to remove it as ETH if we arent doing it the other way around?

hensha256: however we do like to be helpful if we think its a common usecase of course

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -0,0 +1,201 @@
...
+        // how many uncollected tokens are owed to the position, as of the last computation
+        uint128 tokensOwed0;
+        uint128 tokensOwed1;
+    }
+
+    /// @dev Pool keys by poolIds
+    mapping(bytes32 => PoolKey) private _poolIdToPoolKey;
+
+    /// @dev The token ID position data
+    mapping(uint256 => TokenPosition) private _positions;
```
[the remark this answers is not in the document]
tinaszheng: thinking they probably want to use the `positions` getter for getting position data, since it includes helpers for unwrapping `poolId` - otherwise they're kind of exposed to our internal representation of how we store pools right? they'll need both `_poolIdToPoolKey` and `_positions` to get useful position data

contracts/interfaces/IPeripheryPayments.sol
tinaszheng: yeah agreed on not making it opinionated

contracts/NonfungiblePositionManagerV4.sol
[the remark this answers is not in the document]
hensha256: oh yeah good point

contracts/NonfungiblePositionManagerV4.sol
hensha256: ok ive dug into it and these checks should be enough for burn

contracts/base/LiquidityManagement.sol
```diff
@@ -0,0 +1,103 @@
...
+            uint256(int256(delta.amount0())) < params.amount0Min || uint256(int256(delta.amount1())) < params.amount1Min
+        ) revert PriceSlippage();
+    }
+
+    function settleDeltas(address from, PoolKey memory poolKey, BalanceDelta delta) internal {
+        if (delta.amount0() > 0) {
+            pay(poolKey.currency0, from, address(poolManager), uint256(int256(delta.amount0())));
+            poolManager.settle(poolKey.currency0);
+        } else if (delta.amount0() < 0) {
+            poolManager.take(poolKey.currency0, address(this), uint128(-delta.amount0()));
```
hensha256, line 80: i feel like we might want this to send directly to the recipient instead of to `address(this)` - so that we dont then have to send it onwards which is extra gas?

contracts/base/LiquidityManagement.sol
```diff
@@ -0,0 +1,103 @@
...
+
+    enum CallbackType {AddLiquidity}
+
+    struct CallbackData {
+        CallbackType callbackType;
+        address sender;
+        bytes params;
+    }
+
+    struct AddLiquidityParams {
```
snreynolds, line 33: It seems weird to have both`AddLiquidityParams` and `IncreaseLiquidityParams`.  Like hard to know why you use one and not the other.

Also, its possible that we actually can just have ModifyLiquidityParams that handle both increase and decrease..

snreynolds: And we should be consistent with the naming add/remove vs. increase/decrease. I prefer increase/decrease I think

contracts/base/LiquidityManagement.sol
```diff
@@ -0,0 +1,103 @@
...
+            pay(poolKey.currency0, from, address(poolManager), uint256(int256(delta.amount1())));
+            poolManager.settle(poolKey.currency1);
+        } else if (delta.amount1() < 0) {
+            poolManager.take(poolKey.currency1, address(this), uint128(-delta.amount1()));
+        }
+    }
+
+    function lockAcquired(bytes calldata data) external override returns (bytes memory) {
+        CallbackData memory callbackData = abi.decode(data, (CallbackData));
+        if (callbackData.callbackType == CallbackType.AddLiquidity) {
```
snreynolds, line 93: I'm guessing this will be a big block of if statements? 

probably will then want to move the settle logic and the return statement outside the if statement since we will want to settle/return deltas in all of the cases

contracts/base/LiquidityManagement.sol
```diff
@@ -0,0 +1,103 @@
...
+        internal
+        returns (uint128 liquidity, uint256 amount0, uint256 amount1)
+    {
+        (liquidity, amount0, amount1) = abi.decode(
+            poolManager.lock(abi.encode(CallbackData(CallbackType.AddLiquidity, msg.sender, abi.encode(params)))),
+            (uint128, uint256, uint256)
+        );
+    }
+
+    function addLiquidityCallback(AddLiquidityParams memory params)
```
snreynolds, line 55: addLiquidityCallback is a strange name for an internal handler function because callbacks imply an external caller. Also for internal functions we can use the _

maybe `_addLiquidity`?

contracts/interfaces/INonfungiblePositionManagerV4.sol
```diff
@@ -0,0 +1,161 @@
...
+    /// @return amount1 The amount of token1
+    function mint(MintParams calldata params)
+        external
+        payable
+        returns (uint256 tokenId, uint128 liquidity, uint256 amount0, uint256 amount1);
+
+    struct IncreaseLiquidityParams {
+        uint256 tokenId;
+        uint256 amount0Desired;
+        uint256 amount1Desired;
```
snreynolds, line 99: Also is there a reason IncreaseLiquidityParams take the amounts vs. DecreaseLiquidityParams take the liquidity type? Can we make them consistent. 

Also as mentioned elsewhere it actually might be nice to just have one struct for all change liquidity params and make the amounts ints which will tell us if we are increasing or decreasing. 

Some of this also will depend on how modifyPosition is changed in core. I'm not up to date with how @rileydcampbell is currently implementing the interfaces rn

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -0,0 +1,201 @@
...
+
+    /// @dev The ID of the next token that will be minted. Skips 0
+    uint176 private _nextId = 1;
```
snreynolds, line 64: can we call this _nextTokenId? We have tokenIds and poolIds so just want to be clear when we have id var names

contracts/NonfungiblePositionManagerV4.sol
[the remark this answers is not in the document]
snreynolds: nit: Would just add to the natspec that they key is specifically the tokenId (tokenId) ie mapping from tokenId to position data

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -0,0 +1,201 @@
...
+        override
+        isAuthorizedForToken(params.tokenId)
+        returns (uint256 amount0, uint256 amount1)
+    {
+        // TODO: implement this
+    }
+
+    /// @inheritdoc INonfungiblePositionManagerV4
+    function burn(uint256 tokenId) external payable override isAuthorizedForToken(tokenId) {
+        TokenPosition storage position = _positions[tokenId];
```
snreynolds, line 162: the checks below will all be sloads. I think you can copy to memory, do the checks (mloads) below and then delete the storage var with delete _positions[tokenId];

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -0,0 +1,201 @@
...
+        if (position.liquidity != 0 || position.tokensOwed0 != 0 || position.tokensOwed1 != 0) revert NotCleared();
+        delete _positions[tokenId];
+        _burn(tokenId);
+    }
+
+    modifier isAuthorizedForToken(uint256 tokenId) {
+        if (!_isApprovedOrOwner(msg.sender, tokenId)) revert NotApproved();
```
snreynolds, line 169: Hm.. I think we are not using the most up to date erc721 contract from in openzeppelin?

Also I prefer using solmate packages although I think for erc721 open zeppelin may have a few more helpers. @hensha256 thoughts here?

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -0,0 +1,201 @@
...
+    error NotApproved();
+    error NotCleared();
+    error NonexistentToken();
+
+    // details about the Uniswap position
+    struct TokenPosition {
+        // the nonce for permits
+        uint96 nonce;
+        // the address that is approved for spending this token
+        address operator;
```
snreynolds, line 41: hm this is kind of strange because you can have more than 1 operator with the 721 spec... so not sure why we save just one. did we do the same in v3? if so.. why?

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -0,0 +1,201 @@
...
+
+    /// @inheritdoc IERC721
+    function getApproved(uint256 tokenId) public view override(ERC721, IERC721) returns (address) {
+        if (!_exists(tokenId)) revert NonexistentToken();
+
+        return _positions[tokenId].operator;
+    }
+
+    /// @dev Overrides _approve to use the operator in the position, which is packed with the position permit nonce
+    function _approve(address to, uint256 tokenId) internal override(ERC721) {
```
snreynolds, line 185: hmmm I'm kinda confused with the approval/auth scheme we are defining. It sounds like we are trying to override the traditional 721 spec but we are still using(not dis-allowing) the old 721 interfaces.

and we are using the old traditional 721 checks like `_isApprovedOrOwner` which doesnt look at the operator in the TokenPositions struct

snreynolds: Also it seems like we want to allow permits... but where is all that sig verification?

contracts/base/LiquidityManagement.sol
```diff
@@ -0,0 +1,103 @@
...
+            IPoolManager.ModifyPositionParams(params.tickLower, params.tickUpper, int256(int128(liquidity))),
+            params.hookData
+        );
+        if (
+            uint256(int256(delta.amount0())) < params.amount0Min || uint256(int256(delta.amount1())) < params.amount1Min
+        ) revert PriceSlippage();
+    }
+
+    function settleDeltas(address from, PoolKey memory poolKey, BalanceDelta delta) internal {
+        if (delta.amount0() > 0) {
```
snreynolds, line 76: note that this won't work for pools that have custom accounting

contracts/base/LiquidityManagement.sol
```diff
@@ -0,0 +1,103 @@
...
+            pay(poolKey.currency0, from, address(poolManager), uint256(int256(delta.amount0())));
```
snreynolds, line 77: how are we handling native?

contracts/base/LiquidityManagement.sol
```diff
@@ -0,0 +1,103 @@
...
+            params.poolKey,
+            IPoolManager.ModifyPositionParams(params.tickLower, params.tickUpper, int256(int128(liquidity))),
+            params.hookData
+        );
+        if (
+            uint256(int256(delta.amount0())) < params.amount0Min || uint256(int256(delta.amount1())) < params.amount1Min
+        ) revert PriceSlippage();
+    }
+
+    function settleDeltas(address from, PoolKey memory poolKey, BalanceDelta delta) internal {
```
snreynolds, line 75: One high level thing here... I don't think we want to use `take` and `settle`. We will want to use `mint`/ `burn`. 

When the user is withdrawing fully we will use `take`/`settle` though to give them the underlying erc20s. But when we are in the state where the position manager is still managing positions for a user we just want to hold claims to underlying imo (which will interface with the 6909 path)

snreynolds (changes requested):

contracts/interfaces/INonfungiblePositionManagerV4.sol
rileydcampbell: As of right now I haven't changed the types for the modifyLiquidity params. Just broke up the entry point into addLiquidity/removeLiquidity. I'll put up my PR for a review tomorrow! 😃

snreynolds:
Closing in favor of https://github.com/Uniswap/v4-periphery/pull/141

PR #80, opened 2023-11-28
feat: PositionManager mint

tinaszheng:


contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -53,6 +59,14 @@ contract NonfungiblePositionManagerV4 is
         uint128 tokensOwed1;
     }
 
+    enum CallbackAction {Mint}
```
tinaszheng, line 62: or maybe `CallbackType`?

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -113,8 +127,35 @@ contract NonfungiblePositionManagerV4 is
...
+            pay(poolKey.currency0, from, address(poolManager), uint256(int256(delta.amount1())));
+            poolManager.settle(poolKey.currency1);
+        } else if (delta.amount1() < 0) {
+            poolManager.take(poolKey.currency1, address(this), uint128(-delta.amount1()));
+        }
+    }
+
     function lockAcquired(bytes calldata rawData) external returns (bytes memory) {
-        // TODO: implement this
+        require(msg.sender == address(poolManager));
```
tinaszheng, line 147: use custom err here too?

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -133,7 +174,47 @@ contract NonfungiblePositionManagerV4 is
         checkDeadline(params.deadline)
         returns (uint256 tokenId, uint128 liquidity, uint256 amount0, uint256 amount1)
     {
-        // TODO: implement this
+        (tokenId, liquidity, amount0, amount1) = abi.decode(
```
tinaszheng, line 177: is it better if i split this out into two lines?
```
bytes result = poolManager.lock(abi.encode(CallbackData(CallbackAction.Mint, msg.sender, abi.encode(params))));
(tokenId, liquidity, amount0, amount1) = abi.decode(result, (uint256, uint128, uint256, uint256));

File: contracts/NonfungiblePositionManagerV4.sol [hunk 4]
```diff
@@ -133,7 +174,47 @@ contract NonfungiblePositionManagerV4 is
...
+            })
+        );
+
+        tokenId = _nextId++;
+        _mint(params.recipient, tokenId);
+
+        Position.Info memory positionInfo =
+            poolManager.getPosition(poolId, address(this), params.tickLower, params.tickUpper);
+        _positions[tokenId] = TokenPosition({
+            nonce: 0,
```
tinaszheng, line 207: 🤔 copied from v3 but not sure why this, operator, tokensOwed are all initialized to 0

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -211,4 +292,33 @@ contract NonfungiblePositionManagerV4 is
...
+        int24 tickUpper;
+        uint256 amount0Desired;
+        uint256 amount1Desired;
+        uint256 amount0Min;
+        uint256 amount1Min;
+        bytes hookData;
+    }
+
+    /// @notice Add liquidity to an initialized pool
+    function addLiquidity(AddLiquidityParams memory params) internal returns (uint128 liquidity, BalanceDelta delta) {
```
tinaszheng, line 308: might move to this to a separate file

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -120,6 +121,15 @@ contract NonfungiblePositionManagerV4 is
         // TODO: implement this
     }
 
+    /// @dev Caches a pool key
+    function cachePoolKey(PoolKey memory poolKey) private returns (PoolId poolId) {
+        poolId = poolKey.toId();
+        // uninitialized check
+        if (_poolIdToPoolKey[PoolId.unwrap(poolId)].tickSpacing == 0) {
```
tinaszheng, line 128: is checking for `tickSpacing` the right thing to do here? just want to check if this key exists in the mapping yet

tinaszheng: also big difference between this mapping vs the one that exists in v3:
- in v3, there's a poolId indicator PER nft so during burn, we deleted the poolId from the mapping
- but in v4, this is a poolId PER poolkey and there could be multiple NFTs per poolKey. we don't want to burn during mint, so this is essentially an ever-growing list. but it also means not every `mint` needs to store an extra poolId mapping

contracts/NonfungiblePositionManagerV4.sol
```diff
@@ -120,7 +130,37 @@ contract NonfungiblePositionManagerV4 is
         checkDeadline(params.deadline)
         returns (uint256 tokenId, uint128 liquidity, uint256 amount0, uint256 amount1)
     {
-        // TODO: implement this
+        (liquidity, amount0, amount1) = addLiquidity(
+            AddLiquidityParams({
```
hensha256, line 134: this re-defining of AddLiquidityParams is kinda ugly... i think i'd prefer either:

1. We could have AddLiqParams as a sub-struct of MintParams like
```
struct MintParams {
   AddLiquidityParams addLiqParams
   address recipient
   uint256 deadline
}
```

orrrr
2. We could not even bother with the `MintParams` entirely and just have
```
function mint(AddLiquidityParams calldata params, address recipient, uint256 deadline)
```

hensha256: then these 12 lines could just be like `(liquidity, amount0, amount1) = addLiquidity(params);` 😏

contracts/NonfungiblePositionManagerV4.sol
marktoda: yea i think its a fair chec as `tickSpacing` can't be 0 in an initialized pool

contracts/NonfungiblePositionManagerV4.sol
marktoda: ++ option 2

marktoda (commented):
some tests for `mint` would be good?

snreynolds:
Closing in favor of https://github.com/Uniswap/v4-periphery/pull/141

PR #81, opened 2023-12-06
Introduce standard batched call under a lock

snreynolds:
## Related Issue

It might be helpful for integrating contracts to have a way to call multiple functions on core under one lock. This is a start to making a standard for that.

## Description of changes
- Add SafeCallback to enforce that lockAcquired is only callable by the pool manager
- Add abstract contract LockAndBatchCall which handles the base contract for batching calls under a lock
- Add example simple batch contract SimpleBatchCall showing how a contract could be built on top of LockAndBatchCall

I think there could be some improvements in a more standard interface for _settle and _handleAfterExecute. I wonder if there is a way we could also be passing LESS data up and back.. lmk thoughts/improvements/suggestions


test/SimpleBatchCallTest.t.sol
```diff
@@ -0,0 +1,78 @@
...
+        batchCall = new SimpleBatchCall(poolManager);
+        ERC20(Currency.unwrap(currency0)).approve(address(batchCall), 2 ** 255);
+        ERC20(Currency.unwrap(currency1)).approve(address(batchCall), 2 ** 255);
+    }
+
+    function test_initialize() public {
+        bytes[] memory calls = new bytes[](1);
+        calls[0] = abi.encodeWithSelector(ICallsWithLock.initializeWithLock.selector, key, SQRT_RATIO_1_1, ZERO_BYTES);
+        bytes memory settleData =
+            abi.encode(SimpleBatchCall.SettleConfig({withdrawTokens: true, settleUsingTransfer: true}));
```
saucepoint, line 44: unused, no?

contracts/base/LockAndBatchCall.sol
```diff
@@ -0,0 +1,41 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.19;
```
saucepoint, line 2: should be 0.8.22 if you want automatically unchecked `i++` in for-loops 😄

contracts/base/LockAndBatchCall.sol
```diff
@@ -0,0 +1,41 @@
...
+import {CallsWithLock} from "./CallsWithLock.sol";
+
+abstract contract LockAndBatchCall is CallsWithLock, SafeCallback {
+    error CallFail(bytes reason);
+
+    function _settle(address sender, bytes memory data) internal virtual returns (bytes memory settleData);
+    function _handleAfterExecute(bytes memory callReturnData, bytes memory settleReturnData) internal virtual;
+
+    /// @param executeData The function selectors and calldata for any of the function selectors in ICallsWithLock encoded as an array of bytes.
+    function execute(bytes memory executeData, bytes memory settleData) external {
```
saucepoint, line 16: just my opinion and happy to keep as-is, but I think we should type settleData with `SettleConfig`

the serialization/deserialization feels cumbersome from a devex POV

contracts/base/CallsWithLock.sol
saucepoint, line 1: i prefer cleaner function signatures so I think you can drop the `WithLock` suffixes

contracts/SimpleBatchCall.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.19;
```
saucepoint, line 2: 0.8.22 for unchecked `i++` 😉

contracts/SimpleBatchCall.sol
```diff
@@ -0,0 +1,60 @@
...
+                            address(this), address(poolManager), currency.toId(), uint256(delta), new bytes(0)
+                        );
+                    }
+                }
+                if (delta < 0) {
+                    if (config.withdrawTokens) {
+                        poolManager.mint(currency, address(this), uint256(-delta));
+                    } else {
+                        poolManager.take(currency, address(this), uint256(-delta));
+                    }
```
saucepoint, line 47: this is mixed up right? `withdraw=true` should be `take`

saucepoint (commented):
cool stuff. will prob have more feedback once I try building with it but so far looks like a very great start

Issue #84, opened 2023-12-10
FullRange hook could benefit from some gas optimizations


atiselsts:
### Component

Gas Optimization, General design optimization (improving efficiency, cleanliness, or developer experience)

### Describe the suggested feature and problem it solves.

The FullRange hook would be very useful to emulate v2-pool behavior, but it's code needs some love.

### Describe the desired implementation.

* use larger tick spacing to make swaps slightly cheaper
* precompute the sqrtRatioX96 where possible
* simplify logic

### Describe alternatives.

The current tick spacing for the full range pool is set to 60, probably to match 0.3% pools in v3.
Other than familiarity with this number, there's no particular reason to do so – at least not one that I know of. If there are some reasons related to Uniswap integrations, let me know.
In contrast, the most gas efficient approach would be to set a single tick range for the whole pool.
That is clearly not possible at the moment, as the max tick range is limited by a maximal value of a 16-bit integer. However, even changing from 60 to a number closer to 30000 should give noticable gas savings for swappers, on the aggregate.

saucepoint:
Hi! thanks for flagging -- going to close this issue because:

1) hook examples no longer live in the main branch

2) custom accounting will allow for a more v2-style curve with v2 interfaces. reference implementation https://github.com/hensha256/v2-on-v4

PR #85, opened 2023-12-10
gas optimizations for the FullRange hook

atiselsts:
## Related Issue
#84

## Description of changes

This PR attempts to improve the FullRange hook code.

The rationale for increasing the tick spacing is described in #84.

The existing tests show some gas improvements, for most of the tests. However, the main benefits would for large swaps, something that is not captured by the existing tests. A new test `testFullRange_swap_LargeSwap` is added, with a gas snapshot of its own. With 60 tick spacing the gas cost is 175,523. With the new 0x7000 tick spacing it is just 150,173.

Unfortunately the large tick spacing adds some rounding errors to the liquidity distribution, so most of the existing tests were failing. I modified the tests, mostly by changing from `assertEq` to `assertApproxEqAbs` with the accuracy of `DUST` (30 wei).

Let me know how you feel about increasing the tick spacing, if this is contentious I can make a simple PR that just adds the `immutable` fields and the other minor changes.

Also I did not attempt to do the more tricky optimizations like adding `unchecked` to the code, so I believe there's room for further improvements.


saucepoint (commented):
I'm generally in favor of this change as it provides an example on the tradeoffs for tick space selection.

Was wondering how you came up with 0x7000?

Also -- you'll need to sign your commits and run `forge fmt` i believe

atiselsts:
I have signed and updated the PR @saucepoint 

Some of the less common operations, like the second swap in a tx, actually got a little bit more expensive, but there are big gas savings on the first swap, and the code is simpler as well.

About why use 0x7000 as tick spacing -- good question.

I touched this a bit in issue #84. To expand further:
- the max value supported by the v4-core is 0x7fff (max of a signed 16-bit integer)
- however, we can save some gas in calldata by sending zero bytes, so 0x7f00 or 0x7000 are going to be slightly more efficient
- there's not that big difference from the liquidity perspective between 0x7fff, 0x7f00, or 0x7000 ticks.
- however, there's a big difference between 0x7000 and a more "round" number such as 0x1000 or (decimal) 10000.

The reasoning above caused me to prefer 0x7000 (28672 in decimal).

saucepoint:
thank you for the suggestion atis! we arent featuring example hooks anymore. however if that changes in the future, ive added a revisit tag

PR #86, opened 2023-12-11
feat: abstract router

ConjunctiveNormalForm:
## Related Issue
Picked up @ewilz's router work https://github.com/Uniswap/v4-periphery/tree/routing 

## Description of changes


test/V4Router.t.sol
```diff
@@ -0,0 +1,403 @@
...
+        uint256 expectedAmountOut = 992054607780215625;
+
+        tokenPath.push(token0);
+        tokenPath.push(token1);
+        IV4Router.ExactInputParams memory params = getExactInputParams(tokenPath, amountIn);
+
+        uint256 prevBalance0 = token0.balanceOf(address(this));
+        uint256 prevBalance1 = token1.balanceOf(address(this));
+
+        snapStart("RouterExactIn1Hop");
```
zhongeric, line 120: nit: can we refactor the test names to match the tests in core? like `swapExactInOneHop` , and add `_gas` to the end of the test names :)

contracts/V4Router.sol
```diff
@@ -0,0 +1,189 @@
...
+            uint256 pathLength = params.path.length;
+            uint128 amountOut;
+
+            for (uint256 i = 0; i < pathLength; i++) {
+                (PoolKey memory poolKey, bool zeroForOne) = _getPoolAndSwapDirection(params.path[i], params.currencyIn);
+                amountOut = uint128(
+                    -_swapExactPrivate(
+                        poolKey,
+                        zeroForOne,
+                        int256(int128(params.amountIn)),
```
zhongeric, line 79: should just be able to cast to int256 directly

contracts/V4Router.sol
```diff
@@ -0,0 +1,189 @@
...
+                );
+
+                params.amountOut = amountIn;
+                params.currencyOut = params.path[i - 1].intermediateCurrency;
+            }
+            if (amountIn > params.amountInMaximum) revert TooMuchRequested();
+        }
+    }
+
+    function _swapExactPrivate(
```
zhongeric, line 137: can we name this like `_swap`? It differentiates it from the other entrypoints and is a bit more concise

contracts/V4Router.sol
```diff
@@ -0,0 +1,189 @@
...
+                amountIn = uint128(
+                    _swapExactPrivate(
+                        poolKey,
+                        !oneForZero,
+                        -int256(int128(params.amountOut)),
+                        0,
+                        msgSender,
+                        i == 1,
+                        i == pathLength,
+                        params.path[i - 1].hookData
```
zhongeric, line 126: would be nice for `_swapExactPrivate` to take a struct for params so we can have named values, a bit hard to tell what these values represent

contracts/V4Router.sol
```diff
@@ -0,0 +1,189 @@
...
+        _;
+    }
+
+    constructor(IPoolManager _poolManager) {
+        poolManager = _poolManager;
+    }
+
+    function _v4Swap(SwapType swapType, bytes memory params) internal {
+        poolManager.lock(abi.encode(SwapInfo(swapType, msg.sender, params)));
+    }
```
zhongeric, line 34: What's the motivation to having this contract be abstract and not having this be the entrypoint?

zhongeric (commented):
gave it a quick pass but coming together nicely!

contracts/V4Router.sol
ConjunctiveNormalForm: think that's disallowed since [v0.8](https://docs.soliditylang.org/en/latest/080-breaking-changes.html):
>The conversion is only allowed when there is at most one change in sign, width or type-category (int, address, bytesNN, etc.). To perform multiple changes, use multiple conversions.

contracts/V4Router.sol
```diff
@@ -0,0 +1,189 @@
...
+        } else if (swapInfo.swapType == SwapType.ExactOutputSingle) {
+            _swapExactOutputSingle(abi.decode(swapInfo.params, (IV4Router.ExactOutputSingleParams)), swapInfo.msgSender);
+        } else {
+            revert InvalidSwapType();
+        }
+
+        return bytes("");
+    }
+
+    function _swapExactInputSingle(IV4Router.ExactInputSingleParams memory params, address msgSender) private {
```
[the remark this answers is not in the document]
ConjunctiveNormalForm: I think derived contracts would just use the single entry point _swap()?

PR #98, opened 2024-02-29
fix(Multicall): use correct assembly for revert bubbling

penandlim:
## Related Issue
N/A

## Description of changes

image

The implementation taken from stackoverflow uses incorrect assembly.

Since the first 32 bytes of `bytes memory result` will be the length as uint256, only adding 4 to the pointer makes the `Error(string)` signature (0x6408c379a0) as the length, which is incorrect. [See tests here ](https://github.com/penandlim/delegatecall-comparison)

This medium article by @0xdeadbeef0x goes over the possible things that can go wrong with this exact implementation:
https://medium.com/@0xdeadbeef0x/the-double-edged-sword-of-abi-decode-f81529e62bcc

Therefore the correct way is to add 32 and revert the data with the length specified.

Also this solution works for both string style (`revert("reason")`, `require(false, "reason")`) and custom error style `revert CustomError()`, making the code cleaner.


snreynolds:
something like this has already been implemented, thank you for raising this!

PR #100, opened 2024-03-11
[Chore] Update v4-core:latest

qiwu7:
## Related Issue
Updating the repository to the latest v4-core changes

## Description of changes
Update 
- [x] remove access lock flag per https://github.com/Uniswap/v4-core/pull/442
- [x] remove nested locking for LimitOrder hook per https://github.com/Uniswap/v4-core/pull/442
- [x] remove EOA calling capability per https://github.com/Uniswap/v4-core/pull/480/files
- [x] fix tests accordingly


contracts/hooks/examples/LimitOrder.sol
```diff
@@ -144,27 +142,24 @@ contract LimitOrder is BaseHook {
...
 
-            (uint256 amount0, uint256 amount1) = abi.decode(
-                poolManager.lock(
-                    address(this),
-                    abi.encodeCall(this.lockAcquiredFill, (key, lower, -int256(uint256(epochInfo.liquidityTotal))))
-                ),
-                (uint256, uint256)
-            );
+            address locker = poolManager.getLocker();
+            require(locker == sender, "invalid locker");
```
qiwu7, line 160: Is this the right way to remove nested locking?

test/Quoter.t.sol
```diff
@@ -121,9 +121,9 @@ contract QuoterTest is Test, Deployers {
 
     // nested self-call into lockAcquired reverts
     function testQuoter_callLockAcquired_reverts() public {
-        vm.expectRevert(IQuoter.InvalidLockAcquiredSender.selector);
+        vm.expectRevert(IQuoter.LockFailure.selector);
```
qiwu7, line 124: Is this the expected behaviour? I don't see why we are actually getting this error vs the previous after removing the callerAddress from the function signature.

contracts/hooks/examples/LimitOrder.sol
saucepoint: replacing the .lock with _lockAcquiredFill is right

I dont think we need this require statement since its an internal function thats only called from `afterSwap` which already has `poolManagerOnly` modifier

test/Quoter.t.sol
saucepoint: this is a weird test. its testing re-entrancy of lockAcquired?

```solidity
// initiate a reentrancy call
quoter.lockAcquired(abi.encodeWithSelector(quoter.lockAcquired.selector, address(this), "0x"));

// inside lockAcquired, we are re-entering into a 2nd lockAcquired call
address(this).call(...);

// inside the 2nd lockAcquired call, we revert since the sender is the Quoter itself (1st lockAcquired)
if (msg.sender != address(manager)) {
    revert InvalidLockAcquiredSender();
}

// returning to the first lockAcquired call:
(bool success, bytes memory returnData) = address(this).call(data);
if (success) return returnData;
if (returnData.length == 0) revert LockFailure();
```

its the last LockFailure that is getting caught by the test

contracts/hooks/examples/LimitOrder.sol
```diff
@@ -131,7 +129,7 @@ contract LimitOrder is BaseHook {
     }
 
     function afterSwap(
-        address,
+        address sender,
```
hensha256, line 132: this still isnt used right? should be able to leave it unnamed

contracts/hooks/examples/FullRange.sol
```diff
@@ -336,17 +330,17 @@ contract FullRange is BaseHook, ILockCallback {
...
 
-        (uint160 sqrtPriceX96,,) = poolManager.getSlot0(poolId);
+        (uint160 sqrtPriceX96,,,) = poolManager.getSlot0(poolId);
 
         poolManager.swap(
             key,
             IPoolManager.SwapParams({
                 zeroForOne: newSqrtPriceX96 < sqrtPriceX96,
-                amountSpecified: MAX_INT,
+                amountSpecified: -MAX_INT,
```
hensha256, line 343: `- type(intX).max != type(intX).min` - ints are not symmetrical as theres 1 extra negative number than positive. So I think instead you might need a new `MIN_INT` variable to hold `type(int256).min`

hensha256: the minimum integer is actually `-MAX_INT - 1`

saucepoint: good catch!

contracts/hooks/examples/TWAMM.sol
```diff
@@ -136,17 +134,15 @@ contract TWAMM is BaseHook, ITWAMM {
...
 
         (bool zeroForOne, uint160 sqrtPriceLimitX96) = _executeTWAMMOrders(
             twamm, poolManager, key, PoolParamsOnExecute(sqrtPriceX96, poolManager.getLiquidity(poolId))
         );
 
         if (sqrtPriceLimitX96 != 0 && sqrtPriceLimitX96 != sqrtPriceX96) {
-            poolManager.lock(
-                address(this), abi.encode(key, IPoolManager.SwapParams(zeroForOne, type(int256).max, sqrtPriceLimitX96))
-            );
+            poolManager.lock(abi.encode(key, IPoolManager.SwapParams(zeroForOne, type(int256).max, sqrtPriceLimitX96)));
```
hensha256, line 145: this function is called within `beforeSwap` hook, and it then tries to take out a new lock? I feel like that should no longer be possible

test/Quoter.t.sol
snreynolds: Hm maybe we should be bubbling up the InvalidLockAcquiredSender() revert then?

snreynolds (approved):

PR #116, opened 2024-05-28
add hook helpers library

xyek:
This PR adds hook helper library which automates the calculation of flags in the deployment/testing scripts.

### DIFF
--- a/contracts/hooks/examples/FullRange.sol
+++ b/contracts/hooks/examples/FullRange.sol
@@ -26,6 +26,25 @@ import {BeforeSwapDelta, BeforeSwapDeltaLibrary} from "@uniswap/v4-core/src/type
 
 import "../../libraries/LiquidityAmounts.sol";
 
+function hookPermissions() pure returns (Hooks.Permissions memory) {
+    return Hooks.Permissions({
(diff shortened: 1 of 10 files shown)


saucepoint:
hi thanks for the PR!

i think this syntax is interesting, but we're likely to move hook test deployments to use foundry cheatcodes (`deployCodeTo`)

xyek:
That makes sense, how would the `deployCodeTo` address be calculated? Just [this code](https://github.com/Uniswap/v4-periphery/blob/508277999222ee20207cd63efe50397760c2abe1/test/FullRange.t.sol#L90-L91) to be updated right?

Do you think I can do that in this PR? Or is it in the works already?

saucepoint:
the PR to move to deployCodeTo lives here https://github.com/Uniswap/v4-periphery/pull/122

taking a look at this PR again, I'm quite hesitant on defining an arbitrary solidity function outside of a contract. The pattern feels a little too unnatural

if my understanding is correct, this PR lets hook developers adjust their flags in 1 place without needing to update test & scripts. Useful, but I dont think hook developers are changing their flags all that often. Thank you for the consideration!

PR #117, opened 2024-06-04
feat: add UniswapV4DeployerCompetition

marktoda:
this commit adds a contract that creates a competition to generate the
vanity address for Uniswap V4. It does so using CREATE2 salts,
pre-calculating the address at which the contract will be deployed and
applying a score to the address based on its vanity.

## Related Issue

## Description of changes


marktoda:
hackathon project :D

contracts/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,78 @@
...
+    bytes32 public bestAddressSalt;
+    address public bestAddress;
+    address public bestAddressSender;
+
+    address public immutable v4Owner;
+    uint256 public immutable competitionDeadline = block.timestamp + 7 days;
+    uint256 public immutable exclusiveDeployDeadline = competitionDeadline + 1 days;
+    bytes32 public immutable initCodeHash;
+
+    constructor(bytes32 _initCodeHash, address _v4Owner) payable ERC721("UniswapV4 Deployer", "V4D") {
```
hensha256, line 31: hmmmmmm trying to think through how we can get everything to do the same address on every chain. we would definitely love to not have params here 🤔

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,87 @@
...
+                        calculatedScore += 40;
+                        // If the leading 4 4s are also the last 4 nibbles, add 20 points
+                        if (i == addrBytes.length * 2 - 1) {
+                            calculatedScore += 20;
+                        }
+                    }
+                } else {
+                    // If the first nibble after the 4 4s is not a 4, add 20 points
+                    if (fourCounts == 4) {
+                        calculatedScore += 20;
```
marktoda, line 69: we need to reset this to 0 right? else every nibble after the 4 4s will give 20 points

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,87 @@
...
+                // If the first nonzero nibble is not 4, the score is an automatic 0
+                if (firstFour && currentNibble != 4) {
+                    return 0;
+                }
+
+                if (currentNibble == 4) {
+                    fourCounts += 1;
+                    if (fourCounts == 4) {
+                        calculatedScore += 40;
+                        // If the leading 4 4s are also the last 4 nibbles, add 20 points
```
marktoda, line 61: this isn't feasible to mine haha, going to remove

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,87 @@
...
+            if (startingFours) {
+                // If the first nonzero nibble is not 4, the score is an automatic 0
+                if (firstFour && currentNibble != 4) {
+                    return 0;
+                }
+
+                if (currentNibble == 4) {
+                    fourCounts += 1;
+                    if (fourCounts == 4) {
+                        calculatedScore += 40;
```
marktoda, line 60: this seems like a double-counting of 4 4s, no?

Philogy:
Probably want to add some frontrunning protection to `updateBestAddress` e.g. by adding `require(msg.sender == address(bytes20(salt)), "Not original miner");` this still leaves 96-bits of entropy for mining

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,78 @@
...
+        if (block.timestamp < competitionDeadline) {
+            revert CompetitionNotOver(block.timestamp, competitionDeadline);
+        }
+
+        if (msg.sender != deployer && block.timestamp < exclusiveDeployDeadline) {
+            // anyone can deploy after the deadline
+            revert NotAllowedToDeploy(msg.sender, deployer);
+        }
+
+        Create2.deploy(0, bestAddressSalt, bytecode);
```
hensha256, line 73: is this more complicated now that we need to give it a parameter? core doesnt have a parameterless deploy anymore, the v4Owner is passed in

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,99 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import {console2} from "forge-std/console2.sol";
```
hensha256, line 4: removeeee 😁

test/UniswapV4DeployerCompetition.t.sol
```diff
@@ -0,0 +1,146 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+import {Owned} from "solmate/src/auth/Owned.sol";
+import {Test, console2} from "forge-std/Test.sol";
```
hensha256, line 5:
```suggestion
import {Test} from "forge-std/Test.sol";
```

src/UniswapV4DeployerCompetition.sol
hensha256: oh i see in t he tests how you can handle that. So we just need to remove the line below
```
        Owned(bestAddress).transferOwnership(v4Owner);
```

and instead pass in v4Owner as part of the bytecode :)

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,78 @@
...
+
+        // set owner of the pool manager contract
```
hensha256, line 75: should we also maybe add a require statement that code size at bestAddress is now non-0? Or is that overkill lol

src/interfaces/IUniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,24 @@
...
+/// @notice A competition to deploy the UniswapV4 contract with the best address
+interface IUniswapV4DeployerCompetition {
+    event NewAddressFound(address indexed bestAddress, address indexed submitter, uint256 score);
+
+    error InvalidBytecode();
+    error CompetitionNotOver(uint256 currentTime, uint256 deadline);
+    error CompetitionOver(uint256 currentTime, uint256 deadline);
+    error NotAllowedToDeploy(address sender, address deployer);
+    error WorseAddress(address newAddress, address bestAddress, uint256 newScore, uint256 bestScore);
+    error InvalidTokenId(uint256 tokenId);
```
hensha256, line 14: this isnt used

test/UniswapV4DeployerCompetition.t.sol
```diff
@@ -69,7 +74,52 @@ contract UniswapV4DeployerCompetitionTest is Test {
...
+
+        vm.expectRevert(abi.encodeWithSelector(IUniswapV4DeployerCompetition.InvalidSender.selector, salt, winner));
+        vm.prank(winner);
+        competition.updateBestAddress(salt);
+    }
+
+    function test_updateBestAddress_equalSalt_reverts_WorseAddress(bytes32 salt) public {
+        vm.assume(salt != bytes32(0));
+        console2.logBytes32(salt);
+        salt = (salt & mask20bytes) | bytes32(bytes20(winner));
```
marktoda, line 90: I think maybe prefer having the address in the top bits rather than bottom? Just makes it a prefix and incrementing the lower bits feels more natural

hensha256: maybe i'm wrong but i think because bytes are left-aligned this is already in the top bits. Let me check 👀

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+    uint256 public immutable exclusiveDeployDeadline;
+
+    constructor(
+        bytes32 _initCodeHash,
+        address _v4Owner,
+        uint256 _competitionDeadline,
+        address _exclusiveDeployer,
+        uint256 _exclusiveDeployLength
+    ) {
+        initCodeHash = _initCodeHash;
```
gretzke, line 41: let's import the initcode directly from v4-core perhaps
`type(PoolManager).creationCode`

gretzke: maybe also explicitly set the initial owner in the constructor to calculate the correct init code hash

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+
+/// @title UniswapV4DeployerCompetition
+/// @notice A contract to crowdsource a salt for the best Uniswap V4 address
+contract UniswapV4DeployerCompetition is IUniswapV4DeployerCompetition {
+    using VanityAddressLib for address;
+
+    /// @dev The salt for the best address found so far
+    bytes32 public bestAddressSalt;
+    /// @dev The best address found so far
+    address public bestAddress;
```
Jeiwan, line 17: GAS

Not necessary to store since it can always be derived from the salt, which is stored.

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+        if (saltSubAddress != msg.sender && saltSubAddress != address(0)) revert InvalidSender(salt, msg.sender);
+
+        address newAddress = Create2.computeAddress(salt, initCodeHash, address(this));
+        if (bestAddress != address(0) && !newAddress.betterThan(bestAddress)) {
+            revert WorseAddress(newAddress, bestAddress, newAddress.score(), bestAddress.score());
+        }
+
+        bestAddress = newAddress;
+        bestAddressSalt = salt;
+        bestAddressSubmitter = msg.sender;
```
Jeiwan, line 64: GAS

Not used, so likely not necessary to store. The address can be obtained from the `NewAddressFound` event if needed.

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+
+        emit NewAddressFound(newAddress, msg.sender, newAddress.score());
+    }
+
+    /// @inheritdoc IUniswapV4DeployerCompetition
+    function deploy(bytes memory bytecode) external override {
+        if (keccak256(bytecode) != initCodeHash) {
```
Jeiwan, line 71: GAS

Since only a certain bytecode is expected to be deployed, it can be stored in this contract's code. So instead of taking it as an argument, it can be deployed like `Create2.deploy(type(ContractToDeploy).creationCode)`. This will save gas on storing `initCodeHash` and taking `bytecode` as an argument, but will increase the cost of the deployment of this contract, so it should be checked additionally.

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+            revert InvalidBytecode();
+        }
+
+        if (block.timestamp < competitionDeadline) {
```
Jeiwan, line 75: Should be `<=`. Competition is over only after `competitionDeadline`: https://github.com/Uniswap/v4-periphery/pull/117/files#diff-8ef3f95c29150aad2bcff03e1d3142e10c5873dfc46ca1706a967c953513c606R50

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+            revert CompetitionNotOver(block.timestamp, competitionDeadline);
+        }
+
+        if (msg.sender != deployer && block.timestamp < exclusiveDeployDeadline) {
```
Jeiwan, line 79: Shouldn't allow to deploy if `bestAddress`/`bestAddressSalt` are zero.

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+            // anyone can deploy after the deadline
```
Jeiwan, line 80: Technically, a deployment can be made at exactly `exclusiveDeployDeadline`. So, the comment needs to be fixed or `<=` should be used instead of `<` in the check above.

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,97 @@
...
+
+        // 10 points per leading zero nibble
+        uint256 leadingZeroCount = getLeadingNibbleCount(addrBytes, 0, 0);
+        calculatedScore += (leadingZeroCount * 10);
+
+        // special handling for 4s immediately after leading 0s
+        uint256 leadingFourCount = getLeadingNibbleCount(addrBytes, leadingZeroCount, 4);
+        // If the first nonzero nibble is not 4, return 0
+        if (leadingFourCount == 0) {
+            return 0;
```
Jeiwan, line 38: Should revert instead. A 0-score address is not supposed to be accepted, but it can still be saved in `UniswapV4DeployerCompetition` as the first best address.

Or, the revert can be implemented in `UniswapV4DeployerCompetition.updateBestAddress()`.

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,97 @@
...
+        } else if (leadingFourCount == 4) {
+            // 60 points if exactly 4 4s
+            calculatedScore += 60;
+        } else if (leadingFourCount > 4) {
+            // 40 points if more than 4 4s
+            calculatedScore += 40;
```
Jeiwan, line 44: These are not reflected in the scoring rules comment above:
1. the rules say 4 4s score 40 points, but 60 points are in fact scored;
2. the rules don't mention that 4+ 4s are scored less (40) than exactly 4 4s (60).

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,97 @@
...
+    /// @return better True if the first address has a better vanity score
+    function betterThan(address first, address second) internal pure returns (bool better) {
+        return score(first) > score(second);
+    }
+
+    /// @notice Scores an address based on its vanity
+    /// @param addr The address to score
+    /// @return calculatedScore The vanity score of the address
+    function score(address addr) internal pure returns (uint256 calculatedScore) {
+        //  Scoring rules:
```
Jeiwan, line 19: Should be moved to the NatSpec annotation of the function so the rules end up in the docs.

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,97 @@
...
+            // 60 points if exactly 4 4s
+            calculatedScore += 60;
+        } else if (leadingFourCount > 4) {
+            // 40 points if more than 4 4s
+            calculatedScore += 40;
+        }
+
+        // handling for remaining nibbles
+        // for (uint256 i = leadingZeroCount + leadingFourCount; i < addrBytes.length * 2; i++) {
+        for (uint256 i = 0; i < addrBytes.length * 2; i++) {
```
Jeiwan, line 49: The commented out line should be used instead since we're checking the remaining nibbles. Currently, it counts the 4s that were counted earlier.

src/libraries/VanityAddressLib.sol
Jeiwan: Figured it out, it's implemented as per the rules.

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+
+    constructor(
+        bytes32 _initCodeHash,
+        address _v4Owner,
+        uint256 _competitionDeadline,
+        address _exclusiveDeployer,
+        uint256 _exclusiveDeployLength
+    ) {
+        initCodeHash = _initCodeHash;
+        v4Owner = _v4Owner;
```
Jeiwan, line 42: Not used.

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+        if (block.timestamp < competitionDeadline) {
+            revert CompetitionNotOver(block.timestamp, competitionDeadline);
+        }
+
+        if (msg.sender != deployer && block.timestamp < exclusiveDeployDeadline) {
+            // anyone can deploy after the deadline
+            revert NotAllowedToDeploy(msg.sender, deployer);
+        }
+
+        // the owner of the contract must be encoded in the bytecode
```
Jeiwan, line 84: This is not enforced or verified. Probably, the `v4Owner` storage var has something to do with that, but it's currently unused.

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+    ) {
+        initCodeHash = _initCodeHash;
+        v4Owner = _v4Owner;
+        competitionDeadline = _competitionDeadline;
+        exclusiveDeployDeadline = _competitionDeadline + _exclusiveDeployLength;
+        deployer = _exclusiveDeployer;
+    }
+
+    /// @inheritdoc IUniswapV4DeployerCompetition
+    function updateBestAddress(bytes32 salt) external override {
```
Jeiwan, line 49: This function can be front-run. Given that this is a competition and that there might be an incentive to participate, an attacker can front run a competitor's tx to record their address as a best address submitter.

One solution would be implementing a commit-reveal scheme, which will make the contract a little more complex. Another solution is to recommend competitors to use private (and trusted) mempools.

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+        if (block.timestamp > competitionDeadline) {
+            revert CompetitionOver(block.timestamp, competitionDeadline);
+        }
+
+        address saltSubAddress = address(bytes20(salt));
+        if (saltSubAddress != msg.sender && saltSubAddress != address(0)) revert InvalidSender(salt, msg.sender);
```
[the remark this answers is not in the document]
Jeiwan: Seems like a protection from front-running attacks. Related: https://github.com/Uniswap/v4-periphery/pull/117/files#r1828784690

If the zero address is used, there's no protection.

src/UniswapV4DeployerCompetition.sol
Jeiwan: Might be not that beneficial since since it's deployed only once.

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
+// SPADIX-License-Identifier: UNLICENSED
+pragma solidity 0.8.26;
+
+import {Create2} from "@openzeppelin/contracts/utils/Create2.sol";
+import {Owned} from "solmate/src/auth/Owned.sol";
```
NoahMarconi, line 5: unused import

src/UniswapV4DeployerCompetition.sol
NoahMarconi: same for `bestAddressSubmitter` if one of two conditions met:

- the value is not needed on chain in another contract
- `saltSubAddress` not permitted to be anything other than `msg.sender`

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,97 @@
...
+            uint8 currentNibble = getNibble(addrBytes, i);
+
+            // 1 extra point for any 4 nibbles
+            if (currentNibble == 4) {
+                calculatedScore += 1;
+            }
+        }
+
+        // If the last 4 nibbles are 4s, add 20 points
+        if (addrBytes[18] == 0x44 && addrBytes[19] == 0x44) {
```
NoahMarconi, line 59: can shave some gas doing 1 `shl` and comparing the full byte

The generated yul is doing 

```
                /// @src 4:2492:2538  "addrBytes[18] == 0x44 && addrBytes[19] == 0x44"
                let expr_1 := /** @src 4:2492:2513  "addrBytes[18] == 0x44" */ eq(/** @src 2:462:3483  "contract UniswapV4DeployerCompetition is IUniswapV4DeployerCompetition {..." */ and(shl(248, /** @src 4:2492:2505  "addrBytes[18]" */ byte(/** @src 4:2502:2504  "18" */ 0x12, /** @src 4:2492:2505  "addrBytes[18]" */ _1)), /** @src 2:462:3483  "contract UniswapV4DeployerCompetition is IUniswapV4DeployerCompetition {..." */ 0xff00000000000000000000000000000000000000000000000000000000000000), 0x4400000000000000000000000000000000000000000000000000000000000000)
                /// @src 4:2492:2538  "addrBytes[18] == 0x44 && addrBytes[19] == 0x44"
                if expr_1
                {
                    expr_1 := /** @src 4:2517:2538  "addrBytes[19] == 0x44" */ eq(/** @src 2:462:3483  "contract UniswapV4DeployerCompetition is IUniswapV4DeployerCompetition {..." */ and(shl(248, /** @src 4:2517:2530  "addrBytes[19]" */ byte(/** @src 4:2527:2529  "19" */ 0x13, /** @src 4:2517:2530  "addrBytes[19]" */ _1)), /** @src 2:462:3483  "contract UniswapV4DeployerCompetition is IUniswapV4DeployerCompetition {..." */ 0xff00000000000000000000000000000000000000000000000000000000000000), 0x4400000000000000000000000000000000000000000000000000000000000000)
                }
```

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,97 @@
...
+        //     20 points if the first nibble after the 4 4s is NOT a 4
+        //     20 points if the last 4 nibbles are 4s
+        //     1 point for every 4
+
+        // convert the address to bytes for easier parsing
+        bytes20 addrBytes = bytes20(addr);
+
+        // 10 points per leading zero nibble
+        uint256 leadingZeroCount = getLeadingNibbleCount(addrBytes, 0, 0);
+        calculatedScore += (leadingZeroCount * 10);
```
NoahMarconi, line 32: gas: is checked math required throughout?

src/UniswapV4DeployerCompetition.sol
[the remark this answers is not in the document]
hensha256: yeah frontrunning protection of the salts. 0 address is allowed just to allow some flexibility for competitors who have frontrun protection already - but its still required that they use the 0 address so they still only get 12 bytes to play with

src/libraries/VanityAddressLib.sol
iamchrissmith: I don't think this is true.  If we are dealing with the first address (i.e. `bestAddress == address(0)`) then we will not hit the call to `betterThan`.  In `updateBestAddress` we first check if `bestAddress != address(0)`, if it is `address(0)` then we never evaluate `.betterThan` and don't hit this code.  We do however, accept this address and hit this code in the event emission for `score` which we wouldn't want to revert unless we wanted to reject first addresses unless then had a score > 0.

If we already have a `bestAddress` and someone submits a newAddress that doesn't have a four as the first non-zero number, we want it to return 0 here because that will revert with `WorseAddress` in `updateBestAddress`.

src/UniswapV4DeployerCompetition.sol
hensha256: I think we solved this with the salt scheme but do let me know if thats not the case

src/libraries/VanityAddressLib.sol
iamchrissmith: I think this is okay since the rule comments say you get 1 pt for every 4, that would imply it includes the first block of 4 and the last 4 4s

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,97 @@
...
+    /// @param addr The address to score
+    /// @return calculatedScore The vanity score of the address
+    function score(address addr) internal pure returns (uint256 calculatedScore) {
+        //  Scoring rules:
+        //     Requirement: The first nonzero nibble must be 4
+        //     10 points for every leading 0 nibble
+        //     40 points if the first 4 is followed by 3 more 4s
+        //     20 points if the first nibble after the 4 4s is NOT a 4
+        //     20 points if the last 4 nibbles are 4s
+        //     1 point for every 4
```
iamchrissmith, line 25: Confirming a design decision, but from the PR description:
> leading 0's are weighted most heavily, followed by other 0's and 4's

While the most optimal address has many leading 0s (`0x000000000000000000000000000000000000004444` with 360 points), `0x4444000000000000000000000000000000000000` gets a score of 64 whereas `0x0004000000000000000000000000000000000000` would only get a score of 31 pts.

The implemented scoring system optimizes for:
1. Leading 0s only if there are more than 6 of them. (> 60pts)
2. a group of 4s after the first zero (or at the beginning if there is no leading 0) (60 pts)
3. fewer than 6 leading 0s (10-60 pts)
4. 4 4s at the end (20 pts)
5. Other 4s that don't qualify for 2 or 4
It does not give any weight to "other 0's" after the leading ones.

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,97 @@
...
+
+        // convert the address to bytes for easier parsing
+        bytes20 addrBytes = bytes20(addr);
+
+        // 10 points per leading zero nibble
+        uint256 leadingZeroCount = getLeadingNibbleCount(addrBytes, 0, 0);
+        calculatedScore += (leadingZeroCount * 10);
+
+        // special handling for 4s immediately after leading 0s
```
iamchrissmith, line 34: There is no enforcement that there are leading 0s (`leadingZeroCount > 0`)

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,97 @@
...
+        //     40 points if the first 4 is followed by 3 more 4s
+        //     20 points if the first nibble after the 4 4s is NOT a 4
+        //     20 points if the last 4 nibbles are 4s
+        //     1 point for every 4
+
+        // convert the address to bytes for easier parsing
+        bytes20 addrBytes = bytes20(addr);
+
+        // 10 points per leading zero nibble
+        uint256 leadingZeroCount = getLeadingNibbleCount(addrBytes, 0, 0);
```
NoahMarconi, line 31: Hacker's delight has a binary search that does bring in some gas savings (example below tested but not reviewed)

```solidity
function hdGetLeadingNibbleCount(bytes20 addrBytes) internal pure returns (uint256) {
    uint256 n;
    bytes20 x = addrBytes;
    // assume addrBytes validated to be non zero elsewhere
    assembly {
        x := shr(96, x)
        if iszero(shr(128, x)) {
            n := add(n, 32)
            x := shl(128, x)
        }
        if iszero(shr(192, x)) {
            n := add(n, 16)
            x := shl(64, x)
        }
        if iszero(shr(224, x)) {
            n := add(n, 8)
            x := shl(32, x)
        }
        if iszero(shr(240, x)) {
            n := add(n, 4)
            x := shl(16, x)
        }
        if iszero(shr(248, x)) {
            n := add(n, 2)
            x := shl(8, x)
        }
        if iszero(shr(249, x)) {
            n := add(n, 1)
            x := shl(4, x)
        }
    }
    return n - 24;
}
```

NoahMarconi, line 31: gas 

can use a binary search (Hacker's Delight has an overview).

Caution partially tested but not fully reviewed

```solidity
    function hdGetLeadingNibbleCount(bytes20 addrBytes) internal pure returns (uint256) {
        uint256 n;
        bytes20 x = addrBytes;
        // assume addrBytes validated to be non zero elsewhere
        assembly {
            x := shr(96, x)
            if iszero(shr(128, x)) {
                n := add(n, 32)
                x := shl(128, x)
            }
            if iszero(shr(192, x)) {
                n := add(n, 16)
                x := shl(64, x)
            }
            if iszero(shr(224, x)) {
                n := add(n, 8)
                x := shl(32, x)
            }
            if iszero(shr(240, x)) {
                n := add(n, 4)
                x := shl(16, x)
            }
            if iszero(shr(248, x)) {
                n := add(n, 2)
                x := shl(8, x)
            }
            if iszero(shr(252, x)) { n := add(n, 1) }
        }
        return n - 24;
    }
```

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+    /// @inheritdoc IUniswapV4DeployerCompetition
+    function updateBestAddress(bytes32 salt) external override {
+        if (block.timestamp > competitionDeadline) {
+            revert CompetitionOver(block.timestamp, competitionDeadline);
+        }
+
+        address saltSubAddress = address(bytes20(salt));
+        if (saltSubAddress != msg.sender && saltSubAddress != address(0)) revert InvalidSender(salt, msg.sender);
+
+        address newAddress = Create2.computeAddress(salt, initCodeHash, address(this));
```
iamchrissmith, line 57: It looks like there is a `Create2.computeAddress(salt, bytecodeHash)` function signature. It would be slightly simpler and match the `Create2.deploy` interface a little better to use that

src/UniswapV4DeployerCompetition.sol
iamchrissmith: > GAS
> 
> Not necessary to store since it can always be derived from the salt, which is stored.

I agree with this, it could be converted to a public function:
```solidity
function bestAddress() external view returns (address) {
  return Create2.computeAddress(bestAddressSalt, initCodeHash);
}
```

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,97 @@
...
+        } else if (leadingFourCount == 4) {
+            // 60 points if exactly 4 4s
+            calculatedScore += 60;
+        } else if (leadingFourCount > 4) {
+            // 40 points if more than 4 4s
+            calculatedScore += 40;
+        }
+
+        // handling for remaining nibbles
+        // for (uint256 i = leadingZeroCount + leadingFourCount; i < addrBytes.length * 2; i++) {
```
NoahMarconi, line 48: info: some old code commented out

src/UniswapV4DeployerCompetition.sol
iamchrissmith, line 49: this does not need to `override` keyword as it is just inheriting an interface.

src/UniswapV4DeployerCompetition.sol
```diff
@@ -0,0 +1,87 @@
...
+
+        bestAddress = newAddress;
+        bestAddressSalt = salt;
+        bestAddressSubmitter = msg.sender;
+
+        emit NewAddressFound(newAddress, msg.sender, newAddress.score());
+    }
+
+    /// @inheritdoc IUniswapV4DeployerCompetition
+    function deploy(bytes memory bytecode) external override {
```
iamchrissmith, line 70: this does not need to `override` keyword as it is just inheriting an interface.

src/UniswapV4DeployerCompetition.sol
iamchrissmith: I think its still a question of why require them to pass `bytecode` why not just use `initCodeHash` and save this code and custom error?

src/libraries/VanityAddressLib.sol
marktoda: since this is only done at-will and not many times total and maximum iterations is 40, not sure if this is worth it

src/UniswapV4DeployerCompetition.sol
marktoda: https://github.com/Uniswap/v4-periphery/pull/117/commits/46557ecd84147df4f60c98c6b5d999976e2ef3ed

src/UniswapV4DeployerCompetition.sol
marktoda: https://github.com/Uniswap/v4-periphery/pull/117/commits/46557ecd84147df4f60c98c6b5d999976e2ef3ed

src/libraries/VanityAddressLib.sol
marktoda: https://github.com/Uniswap/v4-periphery/pull/117/commits/17a81bf432cd2cb13c356b1c59a4902a90131ef2

src/UniswapV4DeployerCompetition.sol
marktoda: good catch! https://github.com/Uniswap/v4-periphery/pull/117/commits/bc2098560c23093a49ad151e716c6fc407bf1272

src/UniswapV4DeployerCompetition.sol
marktoda: https://github.com/Uniswap/v4-periphery/pull/117/commits/f7889985bc3acb883f24bd254241af7f7028605c

src/UniswapV4DeployerCompetition.sol
marktoda: https://github.com/Uniswap/v4-periphery/pull/117/commits/8055ecb9cf377003930001b82a5f389d797b9172

src/libraries/VanityAddressLib.sol
marktoda: https://github.com/Uniswap/v4-periphery/pull/117/commits/df948e33ecd0a42d5081dc35eac19882e75952fe

src/UniswapV4DeployerCompetition.sol
Jeiwan: On second thought, there's no impact if not a single salt has been submitted; the deployer is absolutely free to not call the `deploy` function. And if someone else calls it, it's they'll only waste gas.

So, not an issue.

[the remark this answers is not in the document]
Jeiwan: Since there are no rewards, and since deploying is not enforced, getting a low final score won't really impact the project. Not an issue.

src/UniswapV4DeployerCompetition.sol
Jeiwan: Cannot be checked on-chain, needs to be ensured by the contract deployer.

src/UniswapV4DeployerCompetition.sol
marktoda: Main concern with this check is that - what do we do if it's not? would necessarily mean some bug in our create2 calculations, so the whole competition is bricked anyways - reverting in that case not really adding any safety

src/UniswapV4DeployerCompetition.sol
marktoda: open to this, but want to take the constructor params inline though

src/UniswapV4DeployerCompetition.sol
marktoda: good call - fixed here! https://github.com/Uniswap/v4-periphery/pull/117/commits/fbd159640ead8792b33bc309cd5071235c41f949

marktoda: good call - fixed here! https://github.com/Uniswap/v4-periphery/pull/117/commits/fbd159640ead8792b33bc309cd5071235c41f949

src/UniswapV4DeployerCompetition.sol
marktoda: storing just for fun rn to be honest

src/UniswapV4DeployerCompetition.sol
marktoda: want constructor params to be included as well, and think this approach is more clear for salt grinders since they can easily fetch the initcodehash they need rather than having to calculate it manually. not too stressed about minor gas for single use

src/UniswapV4DeployerCompetition.sol
marktoda: in latest commit, removed special casing of bestAddressSalt == 0, so it just acts as initial / default value and is fine to deploy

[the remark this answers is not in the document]
marktoda: meh yea, think its fine as the competition runs long enough we can guarantee at least a semi decent result

src/libraries/VanityAddressLib.sol
```diff
@@ -0,0 +1,97 @@
...
+        for (uint256 i = 0; i < addrBytes.length * 2; i++) {
+            uint8 currentNibble = getNibble(addrBytes, i);
+
+            // 1 extra point for any 4 nibbles
+            if (currentNibble == 4) {
+                calculatedScore += 1;
```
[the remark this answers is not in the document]
marktoda: true, think simpler readability here is preferable for me

src/libraries/VanityAddressLib.sol
marktoda: not too stressed about gas optimization on this contract since its mostly one time use - opting ofr readablity

marktoda: not too stressed about gas optimization on this contract since its mostly one time use - opting ofr readablity

src/libraries/VanityAddressLib.sol
marktoda: eh true probably not - adding unchecked block

marktoda: https://github.com/Uniswap/v4-periphery/pull/117/commits/1a7f11772c750dc851b312bd43df2d047f8f522d

src/libraries/VanityAddressLib.sol
marktoda: yea that sounds right. do you have any suggestions here?

.forge-snapshots/positionDescriptor bytecode size.snap
```diff
@@ -1 +1 @@
-31847
\ No newline at end of file
+31728
```
hensha256, line 1: WHYYYYYYY 😂

test/UniswapV4DeployerCompetition.t.sol
hensha256: confirmed this btw - address is left aligned

hensha256 (approved):

PR #118, opened 2024-06-05
rename EPOCH_DEFAULT to EPOCH_NOT_INITIALIZED

Jun1on:
## Related Issue
improves code clarity

## Description of changes
rename EPOCH_DEFAULT to EPOCH_NOT_INITIALIZED in LimitOrder.sol

### DIFF
--- a/contracts/hooks/examples/LimitOrder.sol
+++ b/contracts/hooks/examples/LimitOrder.sol
@@ -57,7 +57,7 @@ contract LimitOrder is BaseHook {
 
     bytes internal constant ZERO_BYTES = bytes("");
 
-    Epoch private constant EPOCH_NOT_INITIALIZED = Epoch.wrap(0);
+    Epoch private constant EPOCH_EMPTY = Epoch.wrap(0);
 
     mapping(PoolId => int24) public tickLowerLasts;
     Epoch public epochNext = Epoch.wrap(1);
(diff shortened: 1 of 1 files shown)


contracts/hooks/examples/LimitOrder.sol
```diff
@@ -159,7 +159,7 @@ contract LimitOrder is BaseHook {
 
     function _fillEpoch(PoolKey calldata key, int24 lower, bool zeroForOne) internal {
         Epoch epoch = getEpoch(key, lower, zeroForOne);
-        if (!epoch.equals(EPOCH_DEFAULT)) {
+        if (!epoch.equals(EPOCH_NOT_INITIALIZED)) {
```
snreynolds, line 162: hmm i actually find this harder to read..

if epoch does not equal epoch not initialized ? its like double negative

maybe could you provide some more context/motivation for this name change?

Jun1on: i feel like it makes the logic easier to follow. it emphasizes that there is nothing in that specific epoch. "default" feels confusing. maybe change the constant to EPOCH_EMPTY?

contracts/hooks/examples/LimitOrder.sol
```diff
@@ -57,7 +57,7 @@ contract LimitOrder is BaseHook {
 
     bytes internal constant ZERO_BYTES = bytes("");
 
-    Epoch private constant EPOCH_DEFAULT = Epoch.wrap(0);
+    Epoch private constant EPOCH_EMPTY = Epoch.wrap(0);
```
hensha256, line 60: yeah i dont mind this - but i'd maybe go for `ZERO_EPOCH`? to match when we use things like `ZERO_ADDRESS`?

saucepoint:
we arent featuring example hooks in this repo at this time, so closing for now

PR #120, opened 2024-06-07
Position Manager

saucepoint:
## Related Issue

## Description of changes

The initial position manager. The incomplete items below to be made into new issues/PRs

- [x] Base functionality: mint/burn/increaseLiquidity/decreaseLiquidity/collect
- [x] proper fee accounting
- [ ] recipient behavior (specify destination when withdrawing tokens)
- [ ] operator/delegate functionality + tests
- [ ] NFT transfer tests (transfer and burn,  transfer and increase/decrease liquidity, transfer and collect)
- [ ] ERC721Permit tests
- [ ] deadline checks + tests
- [ ] slippage checks + tests
- [ ] advanced liquidity mangement (moving to a new pool, adjusting range, etc)
- [ ] exposing arbitrary batch and call operations
- [ ] initialize pool if it doesnt exist


.gitmodules
```diff
@@ -1,6 +1,3 @@
-[submodule "lib/forge-std"]
-	path = lib/forge-std
-	url = https://github.com/foundry-rs/forge-std
```
saucepoint, line 3: instead of maintaining two versions of forge-std, figured we can just use v4-core's forge-std. its a bit more to update with some of the cheatcodes

contracts/NonfungiblePositionManager.sol
```diff
@@ -0,0 +1,142 @@
...
+
+    /// @dev The ID of the next token that will be minted. Skips 0
+    uint256 private _nextId = 1;
+
+    struct TokenPosition {
+        address owner;
+        LiquidityRange range;
+    }
+
+    mapping(uint256 tokenId => TokenPosition position) public tokenPositions;
```
snreynolds, line 35: maybe this will become more clear to me.. but its not clear why we have this mapping and the positions mapping? seems like this just stores LiquidityRange? ... is that really necessary? can we not just pass that in?

and then we only need one mapping:
positions[msg.sender][rangeId]

contracts/NonfungiblePositionManager.sol
```diff
@@ -0,0 +1,142 @@
...
+        // transfer position data to destination
+        positions[to][rangeId] = position;
+        delete positions[from][rangeId];
+
+        // update token position
+        tokenPositions[firstTokenId] = TokenPosition({owner: to, range: tokenPosition.range});
+    }
+
+    modifier isAuthorizedForToken(uint256 tokenId) {
+        require(_isApprovedOrOwner(msg.sender, tokenId), "Not approved");
```
snreynolds, line 139: hm I think _isApprovedOrOwner actually got deprecated in the most up to date version of the OZ ERC721 contracts.

solmates version also doesnt use this

so we should look into updating? or is there a specific reason to still use this?

contracts/libraries/CurrencySenderLibrary.sol
```diff
@@ -0,0 +1,28 @@
...
+    /// @param recipient The recipient address
+    /// @param amount The amount to send
+    /// @param useClaims If true, transfer ERC-6909 tokens
+    function send(Currency currency, IPoolManager manager, address recipient, uint256 amount, bool useClaims)
+        internal
+    {
+        if (useClaims) {
+            manager.transfer(recipient, currency.toId(), amount);
+        } else {
+            manager.burn(address(this), currency.toId(), amount);
```
snreynolds, line 24: this functionality to burn/take mint/settle should be encapsulated in the CurrencySettleTake library

contracts/base/BaseLiquidityHandler.sol
```diff
@@ -0,0 +1,231 @@
+// SPDX-License-Identifier: UNLICENSED
```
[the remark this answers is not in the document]
snreynolds: another high level design comment - its not actually really clear the split between these contracts anyways? BaseLiquidityManagement seems like it could actually be radically simplified (& combined with this contract) 

All of those functions in BaseLiquidityManagement just end up calling _modifyLiquidity on core.. so why not just expose external functions:

collect
add
remove

that then all just call the same _internal_ _modifyLiquidity function with the correct parameters that then are used to determine the proper encoding for the callback data?

contracts/base/BaseLiquidityHandler.sol
```diff
@@ -0,0 +1,231 @@
...
+
+            position.tokensOwed0 = 0;
+            position.tokensOwed1 = 0;
+            position.liquidity -= liquidityToRemove;
+        }
+        {
+            delta = toBalanceDelta(int128(token0Owed), int128(token1Owed));
+
+            // sending tokens to the owner
+            if (token0Owed > 0) range.key.currency0.send(poolManager, owner, token0Owed, useClaims);
```
snreynolds, line 167: i personally dont think we need to support sending 6909 to user.

i do think this contract could keep them as 6909s though (& maybe just by default it should? i think it would be cheaper)

contracts/NonfungiblePositionManager.sol
```diff
@@ -0,0 +1,142 @@
...
+    //     require(params.amount0Min <= uint256(uint128(delta.amount0())), "INSUFFICIENT_AMOUNT0");
+    //     require(params.amount1Min <= uint256(uint128(delta.amount1())), "INSUFFICIENT_AMOUNT1");
+    // }
+
+    function increaseLiquidity(uint256 tokenId, uint256 liquidity, bytes calldata hookData, bool claims)
+        external
+        isAuthorizedForToken(tokenId)
+        returns (BalanceDelta delta)
+    {
+        delta = _increaseLiquidity(tokenPositions[tokenId].range, liquidity, hookData, claims, msg.sender);
```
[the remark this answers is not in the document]
saucepoint: handling the deltas after `manager.modifyLiquidity` is a bit different for increase liquidity and decrease liquidity

decrease liquidity: take all tokens and only send the user's principal + their respective fees

increase liquidity: theres three cases to account for, 1) when a user needs to provide net new tokens, 2) when a user autocompounds their fees perfectly, and 3) when a user needs to take tokens -- a portion of the fees are autocompounded and a remainder is withdrawn

contracts/NonfungiblePositionManager.sol
saucepoint: a Position is keyed by its owner and the range

some interfaces (increaseLiquidity, decreaseLiquidity, and collect) operate on tokenId, so we need a way to recover Position from tokenId

contracts/NonfungiblePositionManager.sol
saucepoint: can look into updating -- grabbed it from v3

contracts/NonfungiblePositionManager.sol
```diff
@@ -0,0 +1,142 @@
...
+    uint256 private _nextId = 1;
+
+    struct TokenPosition {
+        address owner;
+        LiquidityRange range;
+    }
+
+    mapping(uint256 tokenId => TokenPosition position) public tokenPositions;
+
+    constructor(IPoolManager _poolManager) BaseLiquidityManagement(_poolManager) ERC721("Uniswap V4 LP", "LPT") {}
```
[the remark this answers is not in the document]
qiwu7: LPT = Liquidity Provider Token if i have to guess?

contracts/NonfungiblePositionManager.sol
```diff
@@ -0,0 +1,144 @@
...
+    function collect(uint256 tokenId, address recipient, bytes calldata hookData, bool claims)
+        external
+        returns (BalanceDelta delta)
+    {
+        delta = modifyLiquidity(tokenPositions[tokenId].range, 0, hookData, claims);
+    }
+
+    function feesOwed(uint256 tokenId) external view returns (uint256 token0Owed, uint256 token1Owed) {
+        TokenPosition memory tokenPosition = tokenPositions[tokenId];
+        return feesOwed(tokenPosition.owner, tokenPosition.range);
```
qiwu7, line 123: question: why are we using the liquidity range as the fees owed for the lp token?

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -0,0 +1,260 @@
...
+                abi.encodeCall(this.handleModifyLiquidity, (msg.sender, range, liquidityDelta, hookData, claims))
+            ),
+            (BalanceDelta)
+        );
+    }
+
+    function _unlockCallback(bytes calldata data) internal override returns (bytes memory) {
+        (bool success, bytes memory returnData) = address(this).call(data);
+        if (success) return returnData;
+        if (returnData.length == 0) revert LockFailure();
```
qiwu7, line 73: nit: UnlockFailure

contracts/libraries/CurrencySettleTake.sol
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.24;
+
+import {Currency, CurrencyLibrary} from "v4-core/types/Currency.sol";
```
qiwu7, line 4: nit: Let's use consistent remappings. Otherwise this could leads to inconsistent build issues.
```suggestion
import {Currency, CurrencyLibrary} from "@uniswap/v4-core/src/types/Currency.sol";
```

contracts/interfaces/INonfungiblePositionManager.sol
gabririgo, line 1: maintaining the same naming as in the v3 NPM will result in selector clash for contracts/protocols that map selectors to target applications, i.e. diamond pattern. Would it be possible to rename the methods to something like "methodV4", so to allow external protocols to easily interact with v3 and v4?

contracts/NonfungiblePositionManager.sol
```diff
@@ -0,0 +1,146 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.24;
+
+import {ERC721} from "openzeppelin-contracts/contracts/token/ERC721/ERC721.sol";
```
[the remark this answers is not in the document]
saucepoint: Copied ERC721Permit from v3. Stuck with OZ for now since they have helpful modifiers and their _beforeTokenTransfer is a bit easier to use than overriding transfer

contracts/NonfungiblePositionManager.sol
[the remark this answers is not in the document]
saucepoint: Updated to mirror v3's syntax

contracts/interfaces/INonfungiblePositionManager.sol
saucepoint: its something we can keep in mind, but we expect to keep the interfaces as clean as possible

saucepoint:
replaced by https://github.com/Uniswap/v4-periphery/pull/124

PR #121, opened 2024-06-07
Taking fee hook

Jun1on:
## Description of changes
Added taking fee hook to example hooks


test/TakingFee.t.sol
```diff
@@ -0,0 +1,150 @@
...
+        currencies[0] = key.currency0;
+        currencies[1] = key.currency1;
+        takingFee.withdraw(TREASURY, currencies);
+        assertEq(manager.balanceOf(address(this), CurrencyLibrary.toId(key.currency0)), 0);
+        assertEq(manager.balanceOf(address(this), CurrencyLibrary.toId(key.currency1)), 0);
+        assertEq(currency0.balanceOf(TREASURY) / R, expectedFee2 / R);
+        assertEq(currency1.balanceOf(TREASURY) / R, expectedFee / R);
+    }
+
+    function testEdgeCase() public {
```
ConjunctiveNormalForm, line 109: exactly what edge case(s) are you testing? would it be possible to break it into separate test cases?

contracts/hooks/examples/TakingFee.sol
ConjunctiveNormalForm, line 1: nit: maybe `TakingFee` -> `FeeTaking`

test/TakingFee.t.sol
Jun1on: the first swap exhausts the pool of its supply of currency1 so that it is only left with 1 wei. in the second swap, the user wants exact of currency0, so the fee would be taken in currency1.

this test previously failed before ERC6909 implementation because the pool had insufficient tokens to transfer.

Jun1on (commented):
reply

contracts/hooks/examples/FeeTaking.sol
```diff
@@ -0,0 +1,99 @@
...
+
+    function afterSwap(
+        address,
+        PoolKey calldata key,
+        IPoolManager.SwapParams calldata params,
+        BalanceDelta delta,
+        bytes calldata
+    ) external override returns (bytes4, int128) {
+        // fee will be in the unspecified token of the swap
+        bool currency0Specified = (params.amountSpecified < 0 == params.zeroForOne);
```
snreynolds, line 59: This is always so difficult to read (not your fault lol this is exactly how we have to do it) but I think there is a great opportunity here to write a library to help others extract the currencySpecified and currencyUnspecified ie.. calling the lib could look something like:

```
Currency currencySpecified = key.getCurrencySpecified(params);
Currency currencyUnspecified = key.getCurrencyUnspecified(params);
```

snreynolds: or maybe:

```
(Currency currencySpecified, amountSpecified) = key.getSpecified(params);
(Currency currencyUnspecified, amountUnspecified) = key.getUnspecified(params);
```

contracts/hooks/examples/FeeTaking.sol
```diff
@@ -0,0 +1,99 @@
...
+        (Currency feeCurrency, int128 swapAmount) =
+            (currency0Specified) ? (key.currency1, delta.amount1()) : (key.currency0, delta.amount0());
+        // if fee is on output, get the absolute output amount
```
snreynolds, line 62: i think this is saying "if the swap is an exactOutput swap, take the fee of the absolute value of the amt"?

snreynolds: is that right?

contracts/hooks/examples/FeeTaking.sol
```diff
@@ -0,0 +1,99 @@
...
+        if (swapAmount < 0) swapAmount = -swapAmount;
+
+        uint256 feeAmount = (uint128(swapAmount) * swapFeeBips) / TOTAL_BIPS;
```
snreynolds, line 65: so the fee amount is calculated on the specified currency but applied in the unspecified currency.. I think this is odd... What if the two currencies have different decimals? or the conversion between them is not 1-1. 

ie maybe this ends up being a fee amount of like 10000 in currency0 which is worth like $0.50 in USD but that same amount in currency1 is $1000 in USD?

contracts/hooks/examples/FeeTaking.sol
```diff
@@ -0,0 +1,99 @@
...
+        // mint ERC6909 instead of take to avoid edge case where PM doesn't have enough balance
+        poolManager.mint(address(this), CurrencyLibrary.toId(feeCurrency), feeAmount);
+
+        return (BaseHook.afterSwap.selector, feeAmount.toInt128());
+    }
+
+    function setSwapFeeBips(uint128 _swapFeeBips) external onlyOwner {
```
snreynolds, line 72: in the world where there is no owner, maybe this hook is just initialized with a fee?

contracts/hooks/examples/FeeTaking.sol
```diff
@@ -0,0 +1,99 @@
...
+        require(_swapFeeBips <= MAX_BIPS);
```
snreynolds, line 73: instead of require we like to use custom reverts 

`if (_swapFeeBips > MAX_BIPS) revert FeeTooLarge();`

contracts/hooks/examples/FeeTaking.sol
```diff
@@ -0,0 +1,99 @@
...
+        swapFeeBips = _swapFeeBips;
+    }
+
+    function withdraw(address to, Currency[] calldata currencies) external onlyOwner {
```
snreynolds, line 77: in a world where the hook doesn't have an owner, you could potentially allow anyone to call withdraw and just encode that the receiving address is in the CallbackData is an address initialized with the hook

or maybe you provide extreme limitations to the owner, whos only ability is to change that address?

contracts/hooks/examples/FeeTaking.sol
```diff
@@ -0,0 +1,99 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import {BaseHook} from "../../BaseHook.sol";
+import {Hooks} from "@uniswap/v4-core/src/libraries/Hooks.sol";
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+import {BalanceDelta} from "@uniswap/v4-core/src/types/BalanceDelta.sol";
```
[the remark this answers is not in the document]
Jun1on: we should mass rename everything here once that goes through

contracts/hooks/examples/FeeTaking.sol
snreynolds: jk I read this wrong -  I thought swapAmount meant currencySpecified. I think writing a lib like I suggest above plus renaming the amount to amountUnspecified could help here with readability

contracts/hooks/examples/FeeTaking.sol
```diff
@@ -0,0 +1,99 @@
...
+import {BeforeSwapDelta, BeforeSwapDeltaLibrary} from "@uniswap/v4-core/src/types/BeforeSwapDelta.sol";
+import {Currency, CurrencyLibrary} from "@uniswap/v4-core/src/types/Currency.sol";
+import {SafeCast} from "@uniswap/v4-core/src/libraries/SafeCast.sol";
+import {Owned} from "solmate/auth/Owned.sol";
+import {IUnlockCallback} from "@uniswap/v4-core/src/interfaces/callback/IUnlockCallback.sol";
+
+contract FeeTaking is BaseHook, IUnlockCallback, Owned {
+    using SafeCast for uint256;
+
+    bytes internal constant ZERO_BYTES = bytes("");
```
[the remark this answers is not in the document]
Jun1on: BalanceDelta.sol has ZERO_DELTA but not ZERO_BYTES

i saw other example hooks use
bytes internal constant ZERO_BYTES = bytes("");

saucepoint:
closing as were not including example hooks in this repo at this time. hopefully we can repurpose this somewhere!

PR #122, opened 2024-06-12
use deployCodeTo cheatcode for testing hooks

saucepoint:
## Related Issue
The mock & etch pattern for testing hooks is very verbose and adds repo clutter. Conversely, mining addresses for a 14-bit pattern has been reported to be slow by some users on older hardware.

We should replace mock & etch with the `deployCodeTo` cheatcode which is much more concise and easier to reference for hook developers

## Description of changes

Replaced mock & etch with `deployCodeTo` cheatcode


akshatmittal:
Consider doing something like this in order to avoid etching over precompiles:

```solidity
uint160 flags = uint160(Hooks.AFTER_SWAP_FLAG | Hooks.BEFORE_ADD_LIQUIDITY_FLAG | Hooks.BEFORE_INITIALIZE_FLAG) ^ (0x7634 << 144);
TargetHook hook = TargetHook(address(flags));
deployCodeTo("...", abi.encode(manager), address(hook));
```

This is what I usually do in tests, gives a separate namespace to hooks as well. (add another xor if you are testing multiple hooks at once)

(`0x7634` is `v4` in hex)

test/utils/HookTestAddress.sol
saucepoint, line 1: would prefer to see this in v4-core's Deployers (or if periphery's equivalent) to remove one less import required for devs

saucepoint:
closing because example hooks were removed from the main branch

PR #123, opened 2024-06-18
add safecallback

snreynolds:
## Related Issue

## Description of changes


contracts/base/ImmutableState.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.19;
+
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+
+contract ImmutableState {
+    IPoolManager public immutable poolManager;
```
saucepoint, line 7: thoughts on naming the variable `manager`? Feels cleaner when accessing the PM in hooks, its more similar to v4-core (i think), and is more aligned with the modifier / custom error

saucepoint (approved):

contracts/base/ImmutableState.sol
snreynolds: I definitely prefer manager I just thought with position manger it would get confusing but tbh im pro manager haha

saucepoint (approved):

PR #126, opened 2024-06-20
change add liq accounting

snreynolds:
## Related Issue

## Description of changes


contracts/base/BaseLiquidityManagement.sol
```diff
@@ -96,39 +99,77 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
...
+        // if callerDelta <= 0, then tokensOwed0 and tokensOwed1 should be zero'd out as all fees were re-invested into a new position.
+        // if callerDelta > 0, then even after re-investing old fees, the caller still has some fees to collect that were not added into the position so they are accounted.
+        uint128 tokensOwed0 = callerDelta0 > 0 ? uint128(callerDelta0) : 0;
+        uint128 tokensOwed1 = callerDelta1 > 0 ? uint128(callerDelta1) : 0;
+
+        position.updateTokensOwed(tokensOwed0, tokensOwed1);
+        position.add(liquidityToAdd);
+        position.updateFeeGrowthInside(feeGrowthInside0X128, feeGrowthInside1X128);
+
+        return toBalanceDelta(callerDelta0, callerDelta1);
```
saucepoint, line 131: is this set to zero when `callerDelta > 0` (and we credit the remainder fees to `position.tokensOwed`?)

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -96,39 +99,77 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
...
+    }
+
+    // Returns the new sanitized delta for the caller.
+    function _accountNewCallerDeltas(
+        BalanceDelta liquidityDelta,
+        BalanceDelta totalFeesAccrued,
+        BalanceDelta callerFeesAccrued,
+        PoolKey memory poolKey
+    ) internal returns (int128 callerDelta0, int128 callerDelta1) {
```
saucepoint, line 140: another candidate to move the return type o BalanceDelta ?

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -96,39 +99,77 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
...
+        int128 callerFeesAccrued,
+        Currency currency
+    ) internal returns (int128 callerDelta) {
+        int128 feesAccruedOutsideCaller;
+        unchecked {
+            feesAccruedOutsideCaller = totalFeesAccrued - callerFeesAccrued;
+            callerDelta = liquidityDelta - feesAccruedOutsideCaller;
+        }
+        // We must take the fees accrued to other users on the same range, so the accounting in pool manager is up to date before we take/settle all open deltas.
+        currency.take(manager, address(this), uint128(feesAccruedOutsideCaller), true);
```
saucepoint, line 172: theres a case where we also want to `take` the remainder fees, but this might break compatability with the other operations?

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -137,9 +188,24 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
...
+        callerDelta = _increaseLiquidity(owner, range, liquidityToAdd, hookData);
+        _closeCallerDeltas(callerDelta, range.poolKey.currency0, range.poolKey.currency1, owner, claims);
+        _closeAllDeltas(range.poolKey.currency0, range.poolKey.currency1);
+    }
+
+    // When chaining many actions, this should be called at the very end to close out any open deltas owed to or by this contract for other users on the same range.
+    // This is safe because any amounts the caller should not pay or take have already been accounted for in closeCallerDeltas.
+    function _closeAllDeltas(Currency currency0, Currency currency1) internal {
+        int128 delta0 = int128(manager.currencyDelta(address(this), currency0));
+        int128 delta1 = int128(manager.currencyDelta(address(this), currency1));
```
saucepoint, line 201: you can use CurrencyDeltas, which uses sparse extload to load both deltas in a single external call 😎

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -96,39 +110,76 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
...
+
+        return toBalanceDelta(callerDelta0, callerDelta1);
+    }
+
+    // Returns the new sanitized delta for the caller.
+    function _calculateCallerDeltas(
+        BalanceDelta liquidityDelta,
+        BalanceDelta totalFeesAccrued,
+        BalanceDelta callerFeesAccrued
+    ) private pure returns (int128 callerDelta0, int128 callerDelta1) {
```
saucepoint, line 159: I think we're missing `position.tokensOwed` in the callerDelta calculation?

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -34,17 +37,27 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
...
+        int128 callerDelta0 = callerDeltas.amount0();
+        int128 callerDelta1 = callerDeltas.amount1();
 
-        if (delta.amount1() < 0) currency1.settle(manager, owner, uint256(int256(-delta.amount1())), claims);
-        else if (delta.amount1() > 0) currency1.send(manager, owner, uint128(delta.amount1()), claims);
+        if (callerDelta0 < 0) currency0.settle(manager, owner, uint256(int256(-callerDelta0)), claims);
+        else if (callerDelta0 > 0) currency0.send(manager, owner, uint128(callerDelta0), claims);
+
+        if (callerDelta1 < 0) currency1.settle(manager, owner, uint256(int256(-callerDelta1)), claims);
+        else if (callerDelta1 > 0) currency1.send(manager, owner, uint128(callerDelta1), claims);
```
saucepoint, line 60: we should in-line a comment saying that on liquidity-increase, we _never_ send any positive delta, and instead is collected to `position.tokensOwed` (for callerDelta > 0)

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -249,7 +316,8 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
         returns (BalanceDelta delta)
     {
         delta = _collect(owner, range, hookData);
-        zeroOut(delta, range.key.currency0, range.key.currency1, owner, claims);
+        _closeCallerDeltas(delta, range.poolKey.currency0, range.poolKey.currency1, owner, claims);
+        _closeAllDeltas(range.poolKey.currency0, range.poolKey.currency1);
```
saucepoint, line 320: in decrease/collect scenarios (where we send positive deltas to the user), this ordering doesnt smell right to me..

there might be a case where we should collect tokens to posm first _then_ send them to the user

saucepoint (commented):
directionally i think its getting there. makes a lot of sense to separately handle callerDeltas vs externalDeltas and remove PM currency transfers within the operation functions (_increaseLiq, _decreaseLiq, _collect). feels like we can reduce the round trips to PM

have picked up on some edge cases we might need to think thru:

contracts/base/BaseLiquidityManagement.sol
snreynolds: why is that?

snreynolds: I guess in my head the idea is that the `callerDelta` should always be the correct number that is owed to the caller or owed from the caller. 

and the _closeAllDeltas just uses the number open in core. so if callerDelta is up to date then the number queried in core should be up to date.

the alternative design is that we calculate exactly what THIS contract is owed or should owe, do that closing first, and then give the caller the leftover but that felt weirder

contracts/base/BaseLiquidityManagement.sol
snreynolds: thats a great idea

contracts/base/BaseLiquidityManagement.sol
snreynolds: why do we need that?

contracts/base/BaseLiquidityManagement.sol
snreynolds: nice!

snreynolds: can we move that function to TransientStateLibrary.sol ?

saucepoint (approved):

PR #127, opened 2024-06-21
Hook Middleware

Jun1on:
## Related Issue
Hooks can be made to do bad things

## Description of changes
Hook middleware factories that performs sanity checks on the result of a hook to block malicious action


test/utils/HookMiner.sol
saucepoint, line 1: ive changed my mind about using HookMiner for testing. handful of devs reporting its a bit slow on their machines. im in favor of using deployCodeTo instead

Jun1on: i need the HookMiner to get salt to deploy from the factory

PR #130, opened 2024-06-26
fix tests and use BalanceDeltas

saucepoint:
- [x] use BalanceDelta to avoid helper functions, and uniformly apply arithmetic to both `amount0()/amount1()`. cleans up the code
- [x] extended BalanceDelta with support for setting one amount, or adding to an amount
- [x] added autocompounding gas snapshots
- [x] fix / adjust tests


contracts/base/BaseLiquidityManagement.sol
```diff
@@ -126,66 +126,45 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
...
-        console2.log(totalFeesAccrued.amount1());
-
-        // Calculate the accurate tokens owed to the caller.
-        // If the totalFeesAccrued equals the callerFeesAccrued then the total owed to the caller is just the liquidityDelta.
-        // If the totalFeesAccrued is greater than the callerFeesAccrued, we must account for the difference.
-        // TODO: If totalFeesAccrued == callerFeesAccrued, I think we can just apply the entire delta onto the caller, even if this implicitly collects on behalf of another user in the same range.
-        (int128 callerDelta0, int128 callerDelta1) = totalFeesAccrued != callerFeesAccrued
-            ? _calculateCallerDeltas(liquidityDelta, totalFeesAccrued, callerFeesAccrued)
-            : (liquidityDelta.amount0(), liquidityDelta.amount1());
+        // the delta for increase liquidity assuming that totalFeesAccrued was not applied
```
snreynolds, line 129: I still think its a slight optimization to skip all the math stuff if totalFeesAccrued != callerFeesAccrued.. which I think might happen most of the time (that they are equal)

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -126,66 +126,45 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
...
-        position.updateFeeGrowthInside(feeGrowthInside0X128, feeGrowthInside1X128);
+            // this contract is responsible for custodying the excess tokens
+            thisDelta = thisDelta + toBalanceDelta(callerDelta.amount0(), 0);
 
-        // The delta owed or credited by this contract.
-        // TODO @sauce check that if callerDelta == 0 (zerod out from above), then this line just credits the posm to takes on behalf of the caller
-        int128 thisDelta0 = liquidityDelta.amount0() - callerDelta0;
-        int128 thisDelta1 = liquidityDelta.amount1() - callerDelta1;
+            // the caller is not expected to collect the excess tokens
+            callerDelta = toBalanceDelta(0, callerDelta.amount1());
```
snreynolds, line 149: since all of this is shared once for token0 and once for token1 can we just put it in a helper?

snreynolds (commented):
awesome

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -126,66 +128,37 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
...
-        (tokensOwed1, callerDelta1) = callerDelta1 > 0 ? (uint128(callerDelta1), int128(0)) : (uint128(0), callerDelta1);
+        BalanceDelta tokensOwed;
+        if (callerDelta.amount0() > 0) {
+            (tokensOwed, callerDelta, thisDelta) =
+                _moveCallerDeltaToTokensOwed(true, tokensOwed, callerDelta, thisDelta);
+        }
+
+        if (callerDelta.amount1() > 0) {
+            (tokensOwed, callerDelta, thisDelta) =
+                _moveCallerDeltaToTokensOwed(false, tokensOwed, callerDelta, thisDelta);
```
snreynolds, line 156: i think we can get rid of this function when we move to transient stuff but fine for now

snreynolds (approved):

PR #131, opened 2024-06-26
Move gas snapshots to snapLastCall and forge test --isolate

Jun1on:


test/Oracle.t.sol
saucepoint, line 1: looking at the branch locally, theres still some `snapEnd();` calls in this file -- any reason why we arent changing those?

Jun1on: i noticed it as a pattern in v4-core where single tests would still use snapEnd
i wasn't sure, but is there a reason for that?

saucepoint (approved):

snreynolds:
closing in favor of native snapshots!

PR #133, opened 2024-06-26
update decrease

snreynolds:
## Related Issue

## Description of changes


.forge-snapshots/decreaseLiquidity_erc20.snap
```diff
@@ -1 +1 @@
-190026
\ No newline at end of file
+177014
```
snreynolds, line 1: less gas!

contracts/libraries/Position.sol
```diff
@@ -6,18 +6,32 @@ import {BalanceDelta} from "v4-core/types/BalanceDelta.sol";
...
+    }
+
     function addLiquidity(IBaseLiquidityManagement.Position storage position, uint256 liquidity) internal {
         unchecked {
             position.liquidity += liquidity;
         }
     }
 
+    function subtractLiquidity(IBaseLiquidityManagement.Position storage position, uint256 liquidity) internal {
+        if (position.liquidity < liquidity) revert InsufficientLiquidity();
```
snreynolds, line 29: We could probably move this outside of here tbh and check it sooner but for now fine?

saucepoint: i dont mind it here, gives a nice assurance with the following unchecked block

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -291,48 +257,52 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
...
+        }
 
-        position.tokensOwed0 = 0;
-        position.tokensOwed1 = 0;
+        // Allow the caller to collect the tokens owed.
+        // Tokens owed that the caller collects is paid for by this contract.
+        // ie. Transfer the tokensOwed amounts to the caller from the position manager through the pool manager.
+        // TODO case where this contract does not have enough credits to pay the caller?
+        BalanceDelta tokensOwed =
+            toBalanceDelta(uint256(position.tokensOwed0).toInt128(), uint256(position.tokensOwed1).toInt128());
```
saucepoint, line 291: another reason to move `Positon.tokensOwed0/Positioon.tokensOwed1` to BalanceDelta? 🤨

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -236,41 +203,40 @@ contract BaseLiquidityManagement is IBaseLiquidityManagement, SafeCallback {
...
+        if (callerDelta.amount0() > 0) {
+            (tokensOwed, callerDelta, thisDelta) =
+                _moveCallerDeltaToTokensOwed(true, tokensOwed, callerDelta, thisDelta);
+        }
 
-        return callerFeesAccrued;
+        if (callerDelta.amount1() > 0) {
+            (tokensOwed, callerDelta, thisDelta) =
+                _moveCallerDeltaToTokensOwed(false, tokensOwed, callerDelta, thisDelta);
+        }
```
saucepoint, line 236: certain that the `callerDelta.amount1() > 0` isnt needed since decrease liquidity should always create a credit / positive delta

---

noting for myself and can fix with transient library

contracts/libraries/LiquidityDeltaAccounting.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.20;
+
+import {BalanceDelta} from "@uniswap/v4-core/src/types/BalanceDelta.sol";
+
+import "forge-std/console2.sol";
+
+library LiquidityDeltaAccounting {
+    function split(BalanceDelta liquidityDelta, BalanceDelta callerFeesAccrued, BalanceDelta totalFeesAccrued)
```
saucepoint, line 9: i like this library a lot and how its used. its a super minor nit but i think we should call it `discern` and switch the parameters so the syntax becomes:

`liquidityDelta.discern(totalFeesAccrued, callerFeesAccrued)`

saucepoint (approved):
honestly cool to merge so i work on a clean branch for execute and/or transient library

contracts/base/BaseLiquidityManagement.sol
snreynolds: Nice yeah, I think I put that there in case it ever was exactly 0 to just skip all that..

since technically someone could call decrease with liquidityAmount = 0?

PR #138, opened 2024-07-01
Rip Out Vanilla

saucepoint:
gas not the good.. reeeeeee


contracts/NonfungiblePositionManager.sol
```diff
@@ -90,26 +88,14 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
         uint256 deadline,
         address owner,
         bytes calldata hookData
-    ) public payable returns (BalanceDelta delta) {
-        // TODO: optimization, read/write manager.isUnlocked to avoid repeated external calls for batched execution
-        if (manager.isUnlocked()) {
```
snreynolds, line 95: I actually think all of these functions should have a onlyIfUnlocked modifer that checks a variable we transiently store at the beginning of `unlockAndExecute`

```
modifer onlyIfUnlocked() {

if (locked) revert MustBeUnlockedByThisContract();

}
```

snreynolds (commented):
i think we should merge w this one lock change, i still need to think through security model of leaving it open, so lets close it off for now

contracts/NonfungiblePositionManager.sol
```diff
@@ -51,35 +51,33 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
...
     function _unlockCallback(bytes calldata payload) internal override returns (bytes memory) {
         (bytes[] memory data, Currency[] memory currencies) = abi.decode(payload, (bytes[], Currency[]));
 
         bool success;
-        bytes memory returnData;
+
         for (uint256 i; i < data.length; i++) {
             // TODO: bubble up the return
-            (success, returnData) = address(this).call(data[i]);
+            (success,) = address(this).call(data[i]);
```
snreynolds, line 66: I rlly wish we could just parse selector and internally call.. now that I wanna add this modifer :(

contracts/NonfungiblePositionManager.sol
```diff
@@ -187,22 +151,14 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
...
-
-            Currency[] memory currencies = new Currency[](2);
-            currencies[0] = tokenPos.range.poolKey.currency0;
-            currencies[1] = tokenPos.range.poolKey.currency1;
-            bytes memory result = unlockAndExecute(data, currencies);
-            delta = abi.decode(result, (BalanceDelta));
+
+        // TODO: security -- what happens if PoolManager is already unlocked by UR???
+        _collect(recipient, tokenPos.range, hookData);
+        if (recipient != _msgSenderInternal()) {
```
snreynolds, line 159: this check be moved inside _collect() imo? any reason why its here?

saucepoint: agreed -- should be consistent for the vanilla and batch cases

contracts/NonfungiblePositionManager.sol
```diff
@@ -38,8 +38,9 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
     // maps the ERC721 tokenId to the keys that uniquely identify a liquidity position (owner, range)
     mapping(uint256 tokenId => TokenPosition position) public tokenPositions;
 
-    // TODO: TSTORE this jawn
+    // TODO: TSTORE these jawns
```
snreynolds, line 41: lol

snreynolds (approved):

PR #144, opened 2024-07-10
Base Hook Middleware

Jun1on:
## Related Issue
Hooks can be made to do bad things. Because anyone can create their own arbitrary logic for a hook contract, it's difficult for third parties to decide which hooks are "safe" and which are "dangerous". We propose a hook middleware that performs sanity checks on the result of a hooks to block malicious actions.

## Description of changes
Middleware factory creates middlewares.

https://github.com/user-attachments/assets/a7016ed2-2863-42aa-bc69-54fe2549e016

Each middleware is the hook and points to another hook as the implementation. This allows for some convenient configurations where a pool can use a hook and another pool can use the middlewared hook.

 middleware configurations

_all are valid configurations_

Best of all, attaching a middleware to a hook is easy and usually requires no extra coding. The main caveat[^1] is (because of the proxy pattern) constructors will never be called, so it may be necessary to revise the implementation contract to use an initialize function if the constructor needs to set non-immutable variables.

For now, this branch implements and tests only base contracts, which do nothing. [middleware-remove](https://github.com/Uniswap/v4-periphery/pull/145) and [middleware-protect](https://github.com/Uniswap/v4-periphery/pull/152) extend these base contracts to include specific protections. Of course, these are only proposed solutions, and anyone can create their own middleware factory design as they see fit.

[^1]: there are some more small caveats. let's say hook A calls a permissioned function on external contract E. a middleware pointing to hook A would then not be able to call contract E.


contracts/middleware/BaseMiddleware.sol
```diff
@@ -0,0 +1,30 @@
...
+    constructor(IPoolManager _manager, address _impl) {
+        manager = _manager;
+        implementation = _impl;
+    }
+
+    function _implementation() internal view override returns (address) {
+        return implementation;
+    }
+
+    // yo i wanna delete this function but how do i remove this warning
```
ConjunctiveNormalForm, line 26: elaborate?

test/utils/HookMiner.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.21;
```
ConjunctiveNormalForm, line 2: any specific reason we use different versions (0.8.19, 0.8.21, 0.8.24) in different files?

test/middleware/Counter.sol
```diff
@@ -0,0 +1,167 @@
...
+
+    mapping(PoolId => uint256) public beforeAddLiquidityCount;
+    mapping(PoolId => uint256) public afterAddLiquidityCount;
+    mapping(PoolId => uint256) public beforeRemoveLiquidityCount;
+    mapping(PoolId => uint256) public afterRemoveLiquidityCount;
+
+    mapping(PoolId => uint256) public beforeDonateCount;
+    mapping(PoolId => uint256) public afterDonateCount;
+
+    bytes public lastHookData;
```
ConjunctiveNormalForm, line 29: What's the use of this? And does it make sense to make it a mapping from selectors to lastHookData instead

ConjunctiveNormalForm (commented):
I think it's worth writing a more thorough PR description in this case, given we are working with a public repo here, and many (most) people stumble upon this won't have the slightest of context of what hook middleware is and can do. It's even more worth it if the goal is to have people adopt it to write 'certified safe' hooks

test/utils/HookMiner.sol
Jun1on: i copied HookMiner from another repo so it was 0.8.21 - no reason for the others. i can update them all to 0.8.24

test/middleware/Counter.sol
Jun1on: it's just for testing to make sure hookData is passed through correctly

contracts/middleware/BaseMiddleware.sol
Jun1on: i don't need a receive function (because fallback catches it), but if i delete it there is a warning about not having a receive function

contracts/middleware/BaseMiddleware.sol
```diff
@@ -0,0 +1,39 @@
...
+    receive() external payable {
+        _delegate(_implementation());
+    }
+
+    function _ensureValidFlags(address _impl) internal view virtual {
```
marktoda, line 34: I wonder if the underlying implementation actually even needs to support the flags in this case given the middleware already does. What are your thoughts for requiring both addresses to be mined?

contracts/middleware/BaseMiddleware.sol
```diff
@@ -0,0 +1,39 @@
...
+
+import {Hooks} from "@uniswap/v4-core/src/libraries/Hooks.sol";
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {IHooks} from "@uniswap/v4-core/src/interfaces/IHooks.sol";
+import {BalanceDelta} from "@uniswap/v4-core/src/types/BalanceDelta.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+import {BeforeSwapDelta} from "@uniswap/v4-core/src/types/BeforeSwapDelta.sol";
+import {Proxy} from "@openzeppelin/contracts/proxy/Proxy.sol";
+
+contract BaseMiddleware is Proxy {
```
marktoda, line 12: @notice natspec explaining what this does and why

contracts/middleware/BaseMiddleware.sol
```diff
@@ -0,0 +1,39 @@
...
+    /// @notice The address of the pool manager
+    IPoolManager public immutable manager;
```
marktoda, line 14: you're never using the manager immutable as far as I can tell - is this just to be referenced by the underlying impl? if so please elaborate in the comment

contracts/middleware/BaseMiddleware.sol
```diff
@@ -0,0 +1,39 @@
...
+    address public immutable implementation;
```
marktoda, line 15: natspec

test/middleware/Counter.sol
```diff
@@ -0,0 +1,167 @@
...
+
+import {BaseHook} from "./../../contracts/BaseHook.sol";
+import {Hooks} from "@uniswap/v4-core/src/libraries/Hooks.sol";
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+import {BeforeSwapDelta, BeforeSwapDeltaLibrary} from "@uniswap/v4-core/src/types/BeforeSwapDelta.sol";
+import {BalanceDelta, BalanceDeltaLibrary} from "@uniswap/v4-core/src/types/BalanceDelta.sol";
+import {PoolId, PoolIdLibrary} from "@uniswap/v4-core/src/types/PoolId.sol";
+
+contract Counter is BaseHook {
```
marktoda, line 12: nit: name - CounterHook is more clear

contracts/middleware/BaseMiddleware.sol
marktoda: also - given this is meant to be overridden, consider making it an `abstract contract`

contracts/middleware/BaseMiddleware.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.24;
+
+import {Hooks} from "@uniswap/v4-core/src/libraries/Hooks.sol";
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {IHooks} from "@uniswap/v4-core/src/interfaces/IHooks.sol";
```
marktoda, line 6: not used afaict

contracts/middleware/BaseMiddleware.sol
```diff
@@ -0,0 +1,39 @@
...
+import {BalanceDelta} from "@uniswap/v4-core/src/types/BalanceDelta.sol";
```
marktoda, line 7: not used

contracts/middleware/BaseMiddleware.sol
```diff
@@ -0,0 +1,39 @@
...
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
```
marktoda, line 8: not used

contracts/middleware/BaseMiddleware.sol
```diff
@@ -0,0 +1,39 @@
...
+import {BeforeSwapDelta} from "@uniswap/v4-core/src/types/BeforeSwapDelta.sol";
```
marktoda, line 9: not used

contracts/middleware/BaseMiddleware.sol
```diff
@@ -0,0 +1,39 @@
...
+import {Proxy} from "@openzeppelin/contracts/proxy/Proxy.sol";
+
+contract BaseMiddleware is Proxy {
+    /// @notice The address of the pool manager
+    IPoolManager public immutable manager;
+    address public immutable implementation;
+
+    error FlagsMismatch();
+
+    constructor(IPoolManager _manager, address _impl) {
```
marktoda, line 19: what is the purpose of this BaseMiddleware? It seems to be mostly just a proxy wrapper. Are there any useful things that we can add for downstream implementers? Some ideas - 
- Standard modifiers i.e. onlyPoolManager at least?
- Standard implementations of some hook functions, with virtual functions that downstream implementers can override with their custom checks?

contracts/middleware/BaseMiddleware.sol
Jun1on: initially, i did not require the implementation to be mined. however, when working on middleware-protect, i realized doing so would save a bit of gas on flag checks. as the dev would already have to mine a middleware, i don't think it's a big deal to enforce them to mine an implementation as well. also, it's likely that some implementations are already hooks and are already mined.

contracts/middleware/BaseMiddleware.sol
Jun1on: it's pretty much just oz proxy yeah.
- i think it's up to the implementation to add the onlyPoolManager modifier. most hooks already use it, so adding onlyPoolManager to the middleware itself would result in a redundant check.
- like a beforeBeforeSwap, afterBeforeSwap pattern? this sounds a bit complicated and also creates deployment size bloat for middlewares that don't implement the hook.

snreynolds (commented):
Maybe a diagram or added test that I'm missing but.. how does a hook developer interact with these middlewear contracts to set their own hook/deploy their own middlewear ontop of their hook? Just would like to understand the DevX of using this a bit better

contracts/middleware/BaseMiddleware.sol
Jun1on: there's a [caveat in middleware-protect](https://github.com/Uniswap/v4-periphery/pull/152/files#diff-871f5aec4b02f672d30cd275b29b6fae4a49586661e4ad6477547ae8d00c5435R91) that requires it to read the implementation permissions

snreynolds:
Im going to be closing these with the revist label so we have the work saved when we come back to exploring this routing issue!

PR #145, opened 2024-07-11
Middleware remove

Jun1on:
## Related Issue
An incorrectly written or malicious hook could revert in the remove liquidity function, bricking funds or holding user funds hostage.

## Description of changes
This PR implements the MiddlewareRemoveFactory, which deploys either a new MiddlewareRemoveNoDeltas or MiddlewareRemove. This middleware protects users when removing liquidity. Removing can never revert. There is a gas limit of 5M units. If the hook returns deltas, this amount is capped to an immutable value maxFeeBips.

## Rationale
I think remove protections are the best use of middlewares on-chain. While routers can protect against frontrunning during swaps and adding liquidity, they can't prevent a hook from withholding tokens. This middleware doesn't stop the hook from swapping before a user removes liquidity. While this changes the token composition withdrawn, it's a stretch to frame it as malicious towards the user. I’ve added a maxFeeBips parameter for more developer flexibility instead of banning deltas.


contracts/middleware/MiddlewareRemove.sol
```diff
@@ -0,0 +1,95 @@
...
+        BalanceDelta delta,
+        bytes calldata hookData
+    ) external returns (bytes4, BalanceDelta) {
+        address(this).delegatecall{gas: GAS_LIMIT}(
+            abi.encodeWithSelector(this._callAndEnsureZeroDeltas.selector, msg.data)
+        );
+        return (BaseHook.afterRemoveLiquidity.selector, BalanceDeltaLibrary.ZERO_DELTA);
+    }
+
+    function _callAndEnsureZeroDeltas(bytes calldata data) external {
```
snreynolds, line 73: Can't we just not allow the hook to have the flags for `AFTER_ADD_LIQUIDITY_RETURNS_DELTA_FLAG` and `AFTER_REMOVE_LIQUIDITY_RETURNS_DELTA_FLAG`

also I can semi see a case where we'd just want to protect no reverts on remove, but not necessarily no deltas? I wonder if we could split out these two implementations and then have a third implementation that inherits them both?

contracts/middleware/MiddlewareRemove.sol
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    function afterRemoveLiquidity(
+        address sender,
+        PoolKey calldata key,
+        IPoolManager.ModifyLiquidityParams calldata params,
+        BalanceDelta delta,
+        bytes calldata hookData
+    ) external returns (bytes4, BalanceDelta) {
+        address(this).delegatecall{gas: GAS_LIMIT}(
```
snreynolds, line 67: is there a reason to delegatecall here? Is it cause we need the GAS_LIMIT enforcement for the whole remainder or the call? What if we just enforced that gasLeft() > GAS_LIMIT by some proportion so that we could internally call `_callAndEnsurePrice` and then do a delegatecall() with the gas limit externally.. just feels odd that we delegatecall twice?

contracts/middleware/MiddlewareRemove.sol
```diff
@@ -0,0 +1,95 @@
...
+        bytes32 slot = bytes32(NonZeroDeltaCount.NONZERO_DELTA_COUNT_SLOT);
+        uint256 countBefore = uint256(IExttload(address(manager)).exttload(slot));
+        (bool success,) = address(implementation).delegatecall(data);
+        if (!success) {
+            revert FailedImplementationCall();
+        }
+        uint256 countAfter = uint256(IExttload(address(manager)).exttload(slot));
+        if (countAfter != countBefore) {
+            // purpousely revert to cause the whole hook to reset
+            revert HookModifiedDeltas();
```
snreynolds, line 83: What exactly are you trying to prevent with this check? A hook could technically still modify deltas on the pool so long as they settle it back to the count before the call.

contracts/middleware/MiddlewareRemove.sol
```diff
@@ -0,0 +1,95 @@
...
+    function _callAndEnsurePrice(bytes calldata data) external {
+        (, PoolKey memory key,,) = abi.decode(data[4:], (address, PoolKey, IPoolManager.ModifyLiquidityParams, bytes));
+
+        (uint160 priceBefore,,,) = manager.getSlot0(key.toId());
+        (bool success,) = address(implementation).delegatecall(data);
+        if (!success) {
+            revert FailedImplementationCall();
+        }
+        (uint160 priceAfter,,,) = manager.getSlot0(key.toId());
+        if (priceAfter != priceBefore) {
```
snreynolds, line 54: this is an awesome check! Similar to my comment below I wonder if for each check we can abstract them out. And then have a Middlewear contract that inherits any combination of these checks to allow for all sorts of combinations of checks

contracts/middleware/MiddlewareRemove.sol
Jun1on: it's totally ok if the hook modifies deltas as long as they settle it. as i mentioned previously, the goal is to catch reverts when a hook incorrectly settles deltas. if countAfter != countBefore, it's a heuristic that the hook messed something up, in which the **entire** thing reverts (see why i needed two delegatecalls?) and it becomes a vanilla hookless withdrawal.

contracts/middleware/MiddlewareRemove.sol
Jun1on: that would be cool, like lego blocks you can stack to create middleware checks. but is this level of modularity really necessary? i feel like there are only a few middleware designs that truly make sense. and let's say uniswap frontend decides to whitelist middleware-remove hooks for example. there's not really any incentive for new devs to make new middleware designs, as one is all we really need.

unless... the plan is to make these protections modular on the frontend. eg: middleware XYZ has these permissions but is blocked from these middleware selections. almost like an anti-virus checklist that shows up per hook. this would give so much flexibility to the hook developer!

contracts/middleware/MiddlewareRemove.sol
snreynolds: ah gotcha - maybe a comment in the contract explaining this? However this problem is not really solved by the middlewear I think. IE... it doesn't really matter if we revert here or revert from within the hook, the user still won't be able to remove their liquidity...

contracts/middleware/MiddlewareRemove.sol
snreynolds: where is the try/catch?

contracts/middleware/MiddlewareRemove.sol
snreynolds: yeah exactly I like the lego block metaphor

contracts/middleware/MiddlewareRemove.sol
Jun1on: oh well technically not a try/catch. it just doesn't require success

contracts/middleware/MiddlewareRemove.sol
Jun1on: it actually just reverts the entire hook call. see my testFeeOnRemove hook, how the swap still goes through

contracts/middleware/BaseMiddleware.sol
```diff
@@ -3,28 +3,37 @@ pragma solidity ^0.8.24;
...
     function _implementation() internal view override returns (address) {
         return implementation;
     }
 
-    // yo i wanna delete this function but how do i remove this warning
-    receive() external payable {
-        _delegate(_implementation());
+    /// @notice Ensure that the implementation contract has the correct hook flags.
+    /// @dev Override to enforce hook flags.
+    function _ensureValidFlags(address _impl) internal view virtual {
```
snreynolds, line 34: Is there a reason the implementation needs to have the same flags as the middleware?

contracts/middleware/BaseRemove.sol
```diff
@@ -0,0 +1,64 @@
...
+    error FailedImplementationCall();
+
+    bytes internal constant ZERO_BYTES = bytes("");
+    uint256 public constant GAS_LIMIT = 5_000_000;
+    uint256 public constant MAX_BIPS = 10_000;
+
+    // use this hookdata to override checks to save gas. keccak256("override") - 1
+    bytes32 constant OVERRIDE_BYTES = 0x23b70c8dec38c3dec67a5596870027b04c4058cb3ac57b4e589bf628ac6669e7;
+
+    uint256 public immutable maxFeeBips;
```
snreynolds, line 28: why declare this variable in this contract? seems like its never used here, and actually the implementing contract would set this value?

contracts/middleware/BaseRemove.sol
```diff
@@ -0,0 +1,64 @@
...
+
+    constructor(IPoolManager _manager, address _impl) BaseMiddleware(_manager, _impl) {}
+
```
snreynolds, line 31: add natspec!, this would be a good spot to document the hookData param as well

contracts/middleware/BaseRemove.sol
```diff
@@ -0,0 +1,64 @@
...
+        PoolKey calldata key,
+        IPoolManager.ModifyLiquidityParams calldata params,
+        bytes calldata hookData
+    ) external returns (bytes4) {
+        if (bytes32(hookData) == OVERRIDE_BYTES) {
+            (bool success, bytes memory returnData) = address(implementation).delegatecall(
+                abi.encodeWithSelector(this.beforeRemoveLiquidity.selector, sender, key, params, hookData[32:])
+            );
+            return BaseHook.beforeRemoveLiquidity.selector;
+            return abi.decode(returnData, (bytes4));
```
snreynolds, line 43: this line is unreachable

contracts/middleware/BaseRemove.sol
```diff
@@ -0,0 +1,64 @@
...
+        }
+        address(this).delegatecall{gas: GAS_LIMIT}(
+            abi.encodeWithSelector(this._beforeRemoveLiquidity.selector, msg.data)
+        );
+        return BaseHook.beforeRemoveLiquidity.selector;
```
snreynolds, line 48: this is returned twice, i think it would read a little easier if we did

```
if (bytes == overridebytes) {

} else {

}

return BaseHook.beforeRemoveLiquidity.selector;
 ```

File: contracts/middleware/BaseRemove.sol (hunk 8 above)
Comment (snreynolds) [reply contracts/middleware/BaseRemove.sol:43]: I am guessing you want to return what the impl hook returned in which case you'd delete L42

File: contracts/middleware/BaseRemove.sol [hunk 10]
```diff
@@ -0,0 +1,64 @@
...
+    function _beforeRemoveLiquidity(bytes calldata data) external {
+        (bool success, bytes memory returnData) = address(implementation).delegatecall(data);
+        if (!success) {
+            revert FailedImplementationCall();
+        }
+        (bytes4 selector) = abi.decode(returnData, (bytes4));
+        if (selector != BaseHook.beforeRemoveLiquidity.selector) {
+            revert Hooks.InvalidHookResponse();
+        }
+        if (manager.getNonzeroDeltaCount() != 0) {
```
snreynolds, line 60: technically nonzeroDeltaCount could be nonzero so long as its the same value that it was before the delegatecall to the impl

contracts/middleware/MiddlewareRemove.sol
```diff
@@ -0,0 +1,128 @@
...
+        BalanceDelta delta,
+        bytes calldata hookData
+    ) external returns (bytes4, BalanceDelta) {
+        if (bytes32(hookData) == OVERRIDE_BYTES) {
+            (, bytes memory returnData) = address(implementation).delegatecall(
+                abi.encodeWithSelector(this.afterRemoveLiquidity.selector, sender, key, params, delta, hookData[32:])
+            );
+            return abi.decode(returnData, (bytes4, BalanceDelta));
+        }
+        (bool success, bytes memory returnData) = address(this).delegatecall{gas: GAS_LIMIT}(
```
snreynolds, line 44: theres a compiler warning here bc you've already declared a returnData var in this func

contracts/middleware/MiddlewareRemove.sol
```diff
@@ -0,0 +1,128 @@
...
+            revert Hooks.InvalidHookResponse();
+        }
+        uint256 nonzeroDeltaCount = manager.getNonzeroDeltaCount();
+        if (nonzeroDeltaCount == 0 && returnDelta == BalanceDeltaLibrary.ZERO_DELTA) {
+            return returnDelta;
+        }
+        if (
+            returnDelta.amount0() > int256(uint256(int256(delta.amount0())) * maxFeeBips / MAX_BIPS)
+                || returnDelta.amount1() > int256(uint256(int256(delta.amount1())) * maxFeeBips / MAX_BIPS)
+        ) {
```
snreynolds, line 78: throw this in a library to make this code path more readable! and since you are doing this calculation twice.. maybe something like:

feeTake0, feeTaken1 = returnDelta.calculateFeesFrom(delta)
if (maxFee0 > feeTaken0 || maxFee1 > feeTaken1) revert HookTookTooMuchFee();

contracts/middleware/MiddlewareRemove.sol
```diff
@@ -0,0 +1,128 @@
...
+                return returnDelta;
+            }
+        }
+        hookDelta = manager.currencyDelta(address(this), key.currency1);
+        if (hookDelta != 0) {
+            if (-hookDelta != returnDelta.amount1()) {
+                revert HookInvalidDeltasAfterRemove();
+            }
+            nonzeroHookDeltaCount++;
+            if (nonzeroHookDeltaCount == nonzeroDeltaCount) {
```
snreynolds, line 100: is this not the same code twice? also Im confused what this is checking? put into helper to reduce code repetition

contracts/middleware/MiddlewareRemove.sol
```diff
@@ -0,0 +1,128 @@
...
+                return returnDelta;
+            }
+        }
+
+        // weird edge case in case the hook settled the caller's deltas
```
snreynolds, line 105: are you planning on removing? i think we should probably be ok with a hook settling a callers deltas (thats safe behavior)

snreynolds (changes requested):
havent looked at the tests but i left some comments on the base contracts. 

can you please also add 
- natspec!!
- fix compiler warnings (ik there was one we couldn't figure out, can you remind me tmr & I'll try to resolve, but everything else you should be able to fix)
- also you should merge main in all your branches so you also dont have old compiler issues that we dont have in main rn

contracts/middleware/BaseRemove.sol
Jun1on: you're right. while i initially thought of only the user case where someone removes liquidity through posm, someone could purpousely make nonzeroDeltaCount != 0 to circumvent the hook.

contracts/middleware/BaseRemove.sol
Jun1on: i was thinking it was better for standardization. if a third party calls a middleware and wants to read its maxFee, it would return zero. maybe i could make it constant?

contracts/middleware/BaseMiddleware.sol
Jun1on: no, will remove this actually

PR #148, opened 2024-07-14
feat: posm, use salts for all positions

snreynolds:
## Related Issue

## Description of changes


snreynolds:
Still need to remove positions struct, but we need to move operator/nonce @saucepoint 
also still need to update input params for decrease, increase to not accept claims
and support 6909 input on close

saucepoint:
flagging as a followup -- should we add a simple view function to posm to return liquidity / feeGrowth state from PM?

working thru some permit tests, and it doesnt feel good that developers need to know the tokenId is used in salt for `StateLibrary.getPositionInfo`

.forge-snapshots/autocompound_excessFeesCredit.snap
```diff
@@ -1 +1 @@
-311783
\ No newline at end of file
+192327
```
hensha256, line 1: is this a real decrease??

contracts/NonfungiblePositionManager.sol
```diff
@@ -76,25 +66,23 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
...
             } else if (actions[i] == Actions.MINT) {
                 (LiquidityRange memory range, uint256 liquidity, uint256 deadline, address owner, bytes memory hookData)
                 = abi.decode(params[i], (LiquidityRange, uint256, uint256, address, bytes));
-                (BalanceDelta delta, uint256 tokenId) = mint(range, liquidity, deadline, owner, hookData, sender);
-                returnData[i] = abi.encode(delta, tokenId);
+                returnData[i] = abi.encode(mint(range, liquidity, deadline, owner, hookData));
+            } else if (actions[i] == Actions.CLOSE_CURRENCY) {
+                (Currency currency) = abi.decode(params[i], (Currency));
+                returnData[i] = abi.encode(close(currency, sender));
             } else if (actions[i] == Actions.BURN) {
```
hensha256, line 83: Now that v4 has combined `decrease` and `collect` into 1, it feels like we no longer need the multi-step process to close out a position? i.e. if decrease decreases down to 0, shouldnt we just automatically burn the position and not need a burn command? similarly if someone calls increase it could mint a new one?
or is this too complicated?

contracts/NonfungiblePositionManager.sol
```diff
@@ -54,16 +50,10 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
         // TODO: Fix double encode/decode
```
hensha256, line 50: does this need a reentrancy lock?

contracts/NonfungiblePositionManager.sol
```diff
@@ -76,25 +66,23 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
...
-                returnData[i] = abi.encode(decreaseLiquidity(tokenId, liquidity, hookData, claims, sender));
+                (uint256 tokenId, uint256 liquidity, bytes memory hookData) =
+                    abi.decode(params[i], (uint256, uint256, bytes));
+                returnData[i] = abi.encode(decreaseLiquidity(tokenId, liquidity, hookData, sender));
             } else if (actions[i] == Actions.MINT) {
                 (LiquidityRange memory range, uint256 liquidity, uint256 deadline, address owner, bytes memory hookData)
                 = abi.decode(params[i], (LiquidityRange, uint256, uint256, address, bytes));
-                (BalanceDelta delta, uint256 tokenId) = mint(range, liquidity, deadline, owner, hookData, sender);
-                returnData[i] = abi.encode(delta, tokenId);
+                returnData[i] = abi.encode(mint(range, liquidity, deadline, owner, hookData));
```
hensha256, line 79: why is there only a deadline on mint? imo this should go on the high level call and not in these commands - to prevent duplication and to separate it from the internals
like
```
function modifyLiquidities(bytes calldata unlockData, uint256 deadline) checkDeadline(deadline) public ...
```

and then you dont need it anywhere else

contracts/NonfungiblePositionManager.sol
```diff
@@ -106,34 +94,54 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
...
+    function increaseLiquidity(uint256 tokenId, uint256 liquidity, bytes memory hookData, address sender)
         internal
         isAuthorizedForToken(tokenId, sender)
         returns (BalanceDelta delta)
     {
         TokenPosition memory tokenPos = tokenPositions[tokenId];
-
-        delta = _increaseLiquidity(tokenPos.owner, tokenPos.range, liquidity, hookData, sender);
+        // Note: The tokenId is used as the salt for this position, so every minted liquidity has unique storage in the pool manager.
+        (delta,) = _modifyLiquidity(tokenPos.range, liquidity.toInt256(), bytes32(tokenId), hookData);
```
hensha256, line 119: nit: could make liquidity a uint128 so that its guaranteed to fit in `toInt256`

contracts/NonfungiblePositionManager.sol
```diff
@@ -106,34 +94,54 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
...
         isAuthorizedForToken(tokenId, sender)
         returns (BalanceDelta delta)
     {
         TokenPosition memory tokenPos = tokenPositions[tokenId];
+        (delta,) = _modifyLiquidity(tokenPos.range, -(liquidity.toInt256()), bytes32(tokenId), hookData);
+    }
 
-        delta = _decreaseLiquidity(tokenPos.owner, tokenPos.range, liquidity, hookData);
+    // there is no authorization scheme because the payer/recipient is always the sender
+    // TODO: Add more advanced functionality for other payers/recipients, needs auth scheme.
```
hensha256, line 133: yeah i feel like it would be particularly nice for `take` to be able to have a `recipient` and theres no security problems there right

contracts/NonfungiblePositionManager.sol
```diff
@@ -106,34 +94,54 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
...
+    function close(Currency currency, address sender) internal returns (int256 currencyDelta) {
+        // this address has applied all deltas on behalf of the user/owner
+        // it is safe to close this entire delta because of slippage checks throughout the batched calls.
+        currencyDelta = manager.currencyDelta(address(this), currency);
+
+        // the sender is the payer or receiver
+        if (currencyDelta < 0) {
+            currency.settle(manager, sender, uint256(-int256(currencyDelta)), false);
```
hensha256, line 141: nit: its already a int256 no need to cast

contracts/NonfungiblePositionManager.sol
```diff
@@ -106,34 +94,54 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
...
+        } else {
+            currency.take(manager, sender, uint256(int256(currencyDelta)), false);
```
hensha256, line 143: if we're always going to pass in `false` i feel like we should use a more optimised library that doesnt need to do an `if (bool)` on every call to take

hensha256 (commented):
Loveeee the new design. So much simpler to read and understand. Left some random comments/thoughts but generally love it!

hensha256:
particular things you wanted feedback on:

1. **close**
- honestly i think its fine not to have `settle` be able to be from another address. we dont support that in UR for paying for a swap either. i think all good to always pull from msg.sender
- `take` i think it would be great to just add a `recipient` param and send the money there, and that needs no auth so is super simple to implement
- as i said in a comment, would love to see an optimised library that doesnt need the bool? feels like a waste of gas for a highly used contract if we just always pass in false

contracts/base/BaseLiquidityManagement.sol
```diff
@@ -18,15 +18,10 @@ import {SafeCast} from "@uniswap/v4-core/src/libraries/SafeCast.sol";
 import {CurrencySettleTake} from "../libraries/CurrencySettleTake.sol";
 import {CurrencyDeltas} from "../libraries/CurrencyDeltas.sol";
 
-import {FeeMath} from "../libraries/FeeMath.sol";
 import {LiquiditySaltLibrary} from "../libraries/LiquiditySaltLibrary.sol";
```
saucepoint, line 21: we should probably delete the LiquiditySaltLibrary file as we're not deriving salts anymore

gretzke:
can we change the contracts directory to be `src`?

.forge-snapshots/autocompound_excessFeesCredit.snap
snreynolds: what do u mean real... a lot of the testing infra changed so could partially be attributed to that too

PR #139, opened 2024-07-07
PosM: owner-level nonce for permit

saucepoint:
in v3, the nonce for `permit` was at the token level and managed by `NonfungiblePositionManager`. This was gas-inefficient where a user paid clean / cold-storage writes for each position.

in v4, we're putting the nonce at the owner level in `ERC721Permit` -- enabling a one-time cold-storage write, regardless of token / pool.

PR #153, opened 2024-07-15
Owner-level ERC721 Permit

saucepoint:
Owner-level ERC721 Permit (instead of token-level), which reduces frequent cold-storage writes

replaces https://github.com/Uniswap/v4-periphery/pull/139


contracts/base/ERC721Permit.sol
```diff
@@ -46,23 +45,18 @@ abstract contract ERC721Permit is ERC721, IERC721Permit {
         0x49ecf333e5b8c95c40fdafc95c1ad136e8914a8fb55e9dc8bb01eaa83a2df9ad;
```
snreynolds, line 45: I prefer saving the constant as keccak256(string..) 

also can we move all this to a hashing library? `PermitHash.hash(spender, tokenId, nonce, deadline)`?

And then we can build another library to verify the signature after the hash

so permit code looks something like:

```
function permit() checkDeadline(deadline) {

_useNonce()


bytes32 permitHash = PermitHash.hash(spender, tokenId, deadline, nonce);
signature.verify(hash, owner); // handles sc signatures as well

_approve(operator) 

}


```

contracts/base/ERC721Permit.sol
```diff
@@ -71,6 +65,40 @@ abstract contract ERC721Permit is ERC721, IERC721Permit {
...
+        bitPos = uint8(nonce);
+    }
+
+    function _useUnorderedNonce(address from, uint256 nonce) internal {
+        (uint256 wordPos, uint256 bitPos) = bitmapPositions(nonce);
+        uint256 bit = 1 << bitPos;
+        uint256 flipped = nonces[from][wordPos] ^= bit;
+
+        if (flipped & bit == 0) revert NonceAlreadyUsed();
+    }
```
snreynolds, line 103: Nice!! I imagine this was just taken directly from permit2? 

I wonder if we could actually build a UnorderedNonce library thats inheritable/useable for the future?

snreynolds (commented):
Nice! Love the tests too. Just some cleanup ideas - happy for them to be done in a later PR if you can keep track, up to you

PR #154, opened 2024-07-16
Multicall & initialize

saucepoint:
Added a generic multicall interface (from v3) to posm

Also added an external interface for initializing a v4 pool, thats tested with multicall


test/base/Multicall.t.sol
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    function test_multicall() public {
+        bytes[] memory calls = new bytes[](2);
+        calls[0] = abi.encodeWithSelector(MockMulticall(multicall).functionThatReturnsTuple.selector, 10, 20);
+        calls[1] = abi.encodeWithSelector(MockMulticall(multicall).functionThatReturnsTuple.selector, 1, 2);
+
+        bytes[] memory results = multicall.multicall(calls);
+
+        // First call should revert
```
snreynolds, line 21: seems like it does not revert? is this an old comment?

test/base/Multicall.t.sol
```diff
@@ -0,0 +1,68 @@
...
+        calls[1] =
+            abi.encodeWithSelector(MockMulticall(multicall).functionThatRevertsWithError.selector, "Second call failed");
+
+        vm.expectRevert("Second call failed");
+        multicall.multicall(calls);
+    }
+
+    function test_multicall_pays() public {
+        assertEq(address(multicall).balance, 0);
+        multicall.pays{value: 100}();
```
snreynolds, line 54: how come you dont encode the pays call? you just call it directly?

test/base/Multicall.t.sol
```diff
@@ -0,0 +1,68 @@
...
+        assertEq(address(multicall).balance, 100);
+    }
+
+    function test_multicall_returnSender() public view {
+        assertEq(multicall.returnSender(), address(this));
```
snreynolds, line 59: same here

test/mock/MockMulticall.sol
```diff
@@ -0,0 +1,29 @@
...
+    }
+
+    function functionThatReturnsTuple(uint256 a, uint256 b) external pure returns (Tuple memory tuple) {
+        tuple = Tuple({a: a, b: b});
+    }
+
+    uint256 public paid;
+
+    function pays() external payable {
+        paid += msg.value;
```
snreynolds, line 23: if you call pays twice in the multicall wont it set `paid` to msg.value *2, but you've actually only sent in msg.value ?

snreynolds (approved):

PR #156, opened 2024-07-17
Update v4-core

hensha256:
update v4-core, forge-std, and forge-gas-snapshot


test/FullRange.t.sol
```diff
@@ -140,7 +140,12 @@ contract TestFullRange is Test, Deployers, GasSnapshot {
     function testFullRange_beforeInitialize_RevertsIfWrongSpacing() public {
         PoolKey memory wrongKey = PoolKey(key.currency0, key.currency1, 0, TICK_SPACING + 1, fullRange);
 
-        vm.expectRevert(FullRange.TickSpacingNotDefault.selector);
+        vm.expectRevert(
```
snreynolds, line 143: I'm like these bubble ups!

snreynolds (approved):

snreynolds:
Note that we think this was easier to update because settle logic has been continued to be updated in core.. definitely pretty strong argument there to then leave that in core 🤔

PR #160, opened 2024-07-18
Shared actions router

hensha256:
A base shared action router for the V4 swap router and the V4 liquidity router


src/base/ActionsRouter.sol
```diff
@@ -0,0 +1,143 @@
...
+
+            // separate actions on a pool, and settling deltas
+            if (action < 0x10) {
+                if (action == Actions.SWAP) _swap(params[actionIndex]);
+                else if (action == Actions.INCREASE_LIQUIDITY) _increaseLiquidity(params[actionIndex]);
+                else if (action == Actions.DECREASE_LIQUIDITY) _decreaseLiquidity(params[actionIndex]);
+                else if (action == Actions.MINT_POSITION) _mintPosition(params[actionIndex]);
+                else if (action == Actions.BURN_POSITION) _burnPosition(params[actionIndex]);
+                else if (action == Actions.DONATE) _donate(params[actionIndex]);
+                else _handleAdditionalActions2(action, params[actionIndex]);
```
gretzke, line 58: should this be `_handleAdditionalActions1`?

src/base/ActionsRouter.sol
```diff
@@ -0,0 +1,143 @@
...
+        (Currency currency, address recipient) = abi.decode(params, (Currency, address));
+
+        int256 delta = poolManager.currencyDelta(address(this), currency);
+        if (delta < 0) revert();
+
+        poolManager.take(currency, recipient, uint256(delta));
+    }
+
+    /// @notice function to sweep any excess ETH back to a recipient
+    function _sweepETH(bytes memory params) internal virtual {
```
gretzke, line 114: do we need a sweep function for tokens?

src/base/ActionsRouter.sol
```diff
@@ -0,0 +1,143 @@
...
+
+    /// @notice internal function that triggers the execution of a set of actions on v4
+    function _executeActions(bytes calldata params) internal isNotLocked {
+        poolManager.unlock(params);
+    }
+
+    /// @notice function that is called by the PoolManager through the SafeCallback.unlockCallback
+    function _unlockCallback(bytes calldata data) internal override returns (bytes memory) {
+        // TODO decode in calldata directly for gas
+        (uint256[] memory actions, bytes[] memory params) = abi.decode(data, (uint256[], bytes[]));
```
gretzke, line 42: does it make sense to have an array of structs instead of two arrays?

src/base/ActionsRouter.sol
hensha256: yesss good spot

src/base/ActionsRouter.sol
hensha256: yeah i did wonder that... i assumed the only reason people will ever handle tokens are for a final `settle` or `take` on their actions and that tokens should never need to be in the contract? But if you can think of an example i could add it

gretzke: someone could send tokens to the contract by accident

src/base/ActionsRouter.sol
```diff
@@ -0,0 +1,143 @@
...
+
+    /// @notice function to settle the open delta of a currency
+    /// @dev The `_pay` function must implement necessary safety checks on the value of payer.
+    /// If approvals are used on this contract, do not allow `_pay` to pull tokens from any payer.
+    /// @dev The function is virtual and can be overrided if different parameters or logic are required.
+    function _settle(bytes memory params) internal virtual {
+        // TODO decode in calldata
+        (Currency currency, address payer) = abi.decode(params, (Currency, address));
+
+        int256 delta = poolManager.currencyDelta(address(this), currency);
```
snreynolds, line 86: I feel like the base contract should have no impl? Or what was the reasoning to have some impls for some functions and not others?

src/base/ActionsRouter.sol
```diff
@@ -0,0 +1,143 @@
...
+        (uint256[] memory actions, bytes[] memory params) = abi.decode(data, (uint256[], bytes[]));
+
+        uint256 numActions = actions.length;
+        if (numActions != params.length) revert LengthMismatch();
+
+        for (uint256 actionIndex = 0; actionIndex < numActions; actionIndex++) {
+            uint256 action = actions[actionIndex];
+
+            // separate actions on a pool, and settling deltas
+            if (action < 0x10) {
```
snreynolds, line 51: The one bummer is that this will be slightly less optimized since posm will only use INCREASE, DECREASE, TAKE basically.. and so the ordering of checks would just be different for different use cases

src/base/ActionsRouter.sol
snreynolds: But these functions dont need to be part of the "Actions"

src/base/ActionsRouter.sol
```diff
@@ -0,0 +1,143 @@
...
+                else if (action == Actions.BURN_POSITION) _burnPosition(params[actionIndex]);
+                else if (action == Actions.DONATE) _donate(params[actionIndex]);
+                else _handleAdditionalActions2(action, params[actionIndex]);
+            } else {
+                if (action == Actions.SETTLE) _settle(params[actionIndex]);
+                else if (action == Actions.TAKE) _take(params[actionIndex]);
+                else if (action == Actions.MINT_6909) _mint6909(params[actionIndex]);
+                else if (action == Actions.BURN_6909) _burn6909(params[actionIndex]);
+                else if (action == Actions.CLEAR_DELTA) _clearDelta(params[actionIndex]);
+                else if (action == Actions.SWEEP_ETH) _sweepETH(params[actionIndex]);
```
snreynolds, line 65: sweepEth is irrelevant - should be handled not inside the unlockCallback..

These should only be actions relevant to calling a function on v4

src/base/ActionsRouter.sol
```diff
@@ -0,0 +1,143 @@
...
+                else _handleAdditionalActions2(action, params[actionIndex]);
```
snreynolds, line 66: in which case I'd also opt to remove this call.. we can just include every kind of public/external function on v4

src/base/ActionsRouter.sol
hensha256: My reasoning was just that theres only really 1 way youre ever going to take/settle theres no variation there. The variation in `settle` comes down to the `pay` function which is abstract still

hensha256: It also enforces the `sync-pay-settle` pattern nicely

hensha256: But can delete if youd rather

src/base/ActionsRouter.sol
hensha256: itll be using settle too right? So settle and take are optimally positioned, its just liduidity commands that are put 1 lower. We can look at gas differences but I'd expect it to be <50 gas difference

src/libraries/BytesLib.sol
```diff
@@ -0,0 +1,53 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
hensha256, line 1: These functions have been copied in from the UniversalRouter. We cant import it from there otherwise we're going to have a circular dependency between V4-periphery and UR lol. I know they look ugly but theyre audited and live in UR

src/base/ActionsRouter.sol
hensha256: I've removed this but the reason I had it is so that people can add their own custom commands - like `SETTLE_PAIR` for adding liquidity etc

snreynolds:
fixes #172

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,127 @@
...
+    using SafeTransferLib for address;
+    using BytesLib for bytes;
+
+    /// @notice emitted when different numbers of parameters and actions are provided
+    error LengthMismatch();
+
+    /// @notice emitted when an inheriting contract does not support an action
+    error UnsupportedAction(uint256 action);
+
+    constructor(IPoolManager poolManager) SafeCallback(poolManager) {}
```
snreynolds, line 32: we do usually refer to the PoolManager as just `manager`.

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,127 @@
...
+
+    /// @notice internal function that triggers the execution of a set of actions on v4
+    /// TODO move the reentrancy lock modifier up a level
+    function _executeActions(bytes calldata params) internal isNotLocked {
```
snreynolds, line 36: comment about the inheriting contract exposing an external function that triggers the callback ?

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,127 @@
...
+    /// @notice emitted when different numbers of parameters and actions are provided
+    error LengthMismatch();
+
+    /// @notice emitted when an inheriting contract does not support an action
+    error UnsupportedAction(uint256 action);
+
+    constructor(IPoolManager poolManager) SafeCallback(poolManager) {}
+
+    /// @notice internal function that triggers the execution of a set of actions on v4
+    /// TODO move the reentrancy lock modifier up a level
```
snreynolds, line 35: to where?

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,127 @@
...
+    function _executeActions(bytes calldata params) internal isNotLocked {
+        poolManager.unlock(params);
+    }
+
+    /// @notice function that is called by the PoolManager through the SafeCallback.unlockCallback
+    function _unlockCallback(bytes calldata data) internal override returns (bytes memory) {
+        // TODO would it be better to use a struct
```
snreynolds, line 42: in UR we do not, I think enforcing same lengths is fine

test/utils/ActionsRouterPlanner.sol
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: UNLICENSED
```
snreynolds, line 1: Can we just adapt the Planner.sol file that was already written for posm? Its the same code! To me just another reason that we should get base posm in main.. cause it has some useful V4 integration patterns that hopefully this and router would use :)

src/base/SafeCallback.sol
```diff
@@ -1,20 +1,22 @@
...
 import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
 import {ImmutableState} from "./ImmutableState.sol";
 
 abstract contract SafeCallback is ImmutableState, IUnlockCallback {
-    error NotManager();
+    error NotPoolManager();
 
-    modifier onlyByManager() {
-        if (msg.sender != address(manager)) revert NotManager();
+    constructor(IPoolManager poolManager) ImmutableState(poolManager) {}
```
snreynolds, line 11: yeah same comment here.. in core we always reference it as `manager` which I honestly like. Why the change?

src/libraries/BytesLib.sol
snreynolds: Yeah we should really be importing :/ but maybe we can just make this library our own then? Cause I find these really hard to read.. when you do .toUint256Array(0) it seems like its initializing a 0 length uint256 array.

& its also unnecessarily operating on the full returned calldata twice?

I'd rather the syntax just be (uint256[] arr, bytes b) = data.decode(); 

cause we dont use `toUint256Array` or `toBytesArray` by themselves afaik

src/libraries/Actions.sol
```diff
@@ -0,0 +1,17 @@
...
+
+/// @notice Library to define different pool actions.
+library Actions {
+    // TODO can explore having these as an enum but i think constants might be cheaper
+
+    uint256 constant SWAP = 0x00;
+    uint256 constant INCREASE_LIQUIDITY = 0x01;
+    uint256 constant DECREASE_LIQUIDITY = 0x02;
+    uint256 constant DONATE = 0x03;
+    uint256 constant CLEAR_DELTA = 0x04;
```
snreynolds, line 12: nit: prefer just `CLEAR`? SETTLE/TAKE don't include `_DELTA`

Also I do think CLOSE & CLOSE_PAIR are needed in posm

snreynolds: So I think we should discuss... I'm also just wondering if we're going a little too general here to make it work for different use cases. Like generalization is good but now its less optimal for posm, and router bc it includes things both of those use cases shouldnt support. 

ie POSM wont need swap, donate & router wont need increase/decrease.

src/libraries/BytesLib.sol
hensha256: yeah nice idea, i've made a single function that does it all. Optimised for this usecase. Saves like another 300 gas woo

test/utils/ActionsRouterPlanner.sol
hensha256: yeah definitely! just made this because that PR isnt merged yet so needed one here too to test - theyll of course be 1 file whenever theyre both merged

src/base/SafeCallback.sol
hensha256: Changed it because I think `manager` is vague in periphery where we have the concept of both a "position manager" and a "pool manager"

src/base/BaseActionsRouter.sol
hensha256: I'm removing it from this contract. UR already has a reentrancy lock because it needs one wider across all commands. So I think we should just have the upper level contracts implement them (if they need them).

src/base/BaseActionsRouter.sol
hensha256: responded below as to why - i'm happy to change it but i do think its confusing given we have a "pool manager" and a "position manager". Its not clear what "manager" is when you call e.g. `manager.modifyLiquidity`

snreynolds: ah yeah that makes sense

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,59 @@
...
+        uint256 numActions = actions.length;
+        if (numActions != params.length) revert LengthMismatch();
+
+        for (uint256 actionIndex = 0; actionIndex < numActions; actionIndex++) {
+            uint256 action = actions[actionIndex];
+
+            _handleAction(action, params[actionIndex]);
+        }
+
+        // TODO do we want to return anything?
```
snreynolds, line 48: yes! In posm we return a `bytes[]` and each _handleAction return is a `bytes`. Just keeps it general so that all actions can return whatever they want

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /// @notice function to handle the parsing and execution of an action and its parameters
+    function _handleAction(uint256 action, bytes calldata params) internal virtual;
+
+    /// @notice function that returns address considered executer of the actions
+    /// @dev In many contracts this will be the address that calls the initial entry point that calls `_executeActions`
+    /// `msg.sender` shouldnt be used, as this will be the v4 pool manager contract that calls `unlockCallback`
+    /// If using ReentrancyLock.sol, this function can return Locker.get() - locker of the contract
```
snreynolds, line 57: should we put a note that we do not handle other context dependent variables (msg.data/msg.value) bc the unlockCallback actions should only use variables passed in from the main entry point OR _msgSender()

src/libraries/BytesLib.sol
```diff
@@ -0,0 +1,42 @@
...
+pragma solidity ^0.8.0;
+
+/// @title Library for Bytes Manipulation
+library BytesLib {
+    error SliceOutOfBounds();
+
```
snreynolds, line 7: natspec, what are these magic bytes? Is it the SliceOutOfBounds.selector?

src/libraries/BytesLib.sol
```diff
@@ -0,0 +1,42 @@
...
+    bytes4 constant SLICE_ERROR_SELECTOR = 0x3b99b53d;
+
+    /// @notice Performs the equivalent of `abi.decode(data, (uint256[], bytes[]))` in calldata
+    /// @param _bytes The input bytes string to extract input arrays from
+    /// @return actions The uint256 calldata array of actions
+    /// @return params The bytes calldata array of parameters
+    function decodeInCalldata(bytes calldata _bytes)
```
snreynolds, line 14: Hmm.. so I think it makes sense to keep the params in calldata before we call `unlock` but is there benefit to keeping them in calldata here? at some point they are still put in memory.. I'm just wondering if we are doing extra work by doing calldata -> decode but keep in calldata -> memory

and if it would actually be cheaper to just put these directly into memory now?

test/mocks/MockBaseActionsRouter.sol
```diff
@@ -0,0 +1,84 @@
...
+    uint256 public mintCount;
+    uint256 public burnCount;
+
+    constructor(IPoolManager poolManager) BaseActionsRouter(poolManager) {}
+
+    function executeAction(bytes calldata params) external isNotLocked {
+        _executeActions(params);
+    }
+
+    function _handleAction(uint256 action, bytes calldata params) internal override {
```
snreynolds, line 28: ooo nice ok!!! so now the implementer just chooses their actions, and they can keep this and override/revert with the default actions or roll their own.

Guess we should discuss for Router/Posm how we wanna do this? Ie if router is going to have an addLiq/removeLiq command then we can make them the same value

snreynolds (commented):
nice!! i feel really good about this - super flexible but if we do it right can get a lot of great shared code between posm + router

src/libraries/BytesLib.sol
hensha256: Its for gas efficiency. For some of them they will never need to go into memory - and for the others its cheapest to decode as late as possible so you dont pass around in memory

hensha256: And then when we do decode to memory it should be done in assembly like how UR does it - its muchhh cheaper than doing abi.decode

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,59 @@
...
+    constructor(IPoolManager poolManager) SafeCallback(poolManager) {}
+
+    /// @notice internal function that triggers the execution of a set of actions on v4
+    /// @dev inheriting contracts should call this function to trigger execution
+    function _executeActions(bytes calldata params) internal {
+        poolManager.unlock(params);
+    }
+
+    /// @notice function that is called by the PoolManager through the SafeCallback.unlockCallback
+    function _unlockCallback(bytes calldata data) internal override returns (bytes memory) {
```
snreynolds, line 35: i think this line introduces a compiler warning - I think @gretzke had a way to fix but not sure? Could just return empty bytes

src/base/BaseActionsRouter.sol
snreynolds: +1 to the above unless it just didn't load for me the updated comment?

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.24;
+
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {Currency} from "@uniswap/v4-core/src/types/Currency.sol";
```
snreynolds, line 5: unused import

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,59 @@
...
+import {TransientStateLibrary} from "@uniswap/v4-core/src/libraries/TransientStateLibrary.sol";
```
snreynolds, line 6: unused import

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,59 @@
...
+
+import {SafeTransferLib} from "solmate/utils/SafeTransferLib.sol";
+
+import {SafeCallback} from "./SafeCallback.sol";
+import {BytesLib} from "../libraries/BytesLib.sol";
+
+/// @notice Abstract contract for performing a combination of actions on Uniswap v4.
+/// @dev Suggested uint256 action values are defined in Actions.sol, however any definition can be used
+abstract contract BaseActionsRouter is SafeCallback {
+    using TransientStateLibrary for IPoolManager;
```
snreynolds, line 16: can remove

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.24;
+
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {Currency} from "@uniswap/v4-core/src/types/Currency.sol";
+import {TransientStateLibrary} from "@uniswap/v4-core/src/libraries/TransientStateLibrary.sol";
+
+import {SafeTransferLib} from "solmate/utils/SafeTransferLib.sol";
```
snreynolds, line 8: also this lol.. i think these just must all be leftover from old version

src/base/BaseActionsRouter.sol
```diff
@@ -0,0 +1,59 @@
...
+
+import {SafeCallback} from "./SafeCallback.sol";
+import {BytesLib} from "../libraries/BytesLib.sol";
+
+/// @notice Abstract contract for performing a combination of actions on Uniswap v4.
+/// @dev Suggested uint256 action values are defined in Actions.sol, however any definition can be used
+abstract contract BaseActionsRouter is SafeCallback {
+    using TransientStateLibrary for IPoolManager;
+    using SafeTransferLib for address;
```
snreynolds, line 17: and this

test/BaseActionsRouter.t.sol
```diff
@@ -0,0 +1,140 @@
...
+
+import {MockBaseActionsRouter} from "./mocks/MockBaseActionsRouter.sol";
+import {ActionsRouterPlanner, Plan} from "./utils/ActionsRouterPlanner.sol";
+import {Actions} from "../src/libraries/Actions.sol";
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {Test} from "forge-std/Test.sol";
+import {Deployers} from "@uniswap/v4-core/test/utils/Deployers.sol";
+import {GasSnapshot} from "forge-gas-snapshot/GasSnapshot.sol";
+
+contract BaseActionsRouterTest is Test, Deployers, GasSnapshot {
```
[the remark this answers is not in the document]
snreynolds: https://github.com/Uniswap/v4-periphery/issues/184

snreynolds (commented):
amazing ⭐  ⭐  ⭐  ⭐ 

left one comment about tests but we can add as a followup. merging this and dev-posm will have 2 planner.sol like files so just a note to combine later... ? i think it might be slightly easier to just fix this to use the other one cause theres only one test file here

snreynolds (approved):

PR #161, opened 2024-07-18
ERC721Permit cleanup

saucepoint:
some fixes from https://github.com/Uniswap/v4-periphery/pull/153

* OZ --> solmate
* cleaned up operator. we're using the base ERC721 getApproved as the operator. in v3 they were tightly packing operator and nonce, but since we're not doing that we can juse use the default ERC721 approved address
* ripped out a bunch of crap (BaseLiquidityManagemtn) 😭


contracts/NonfungiblePositionManager.sol
```diff
@@ -146,37 +145,55 @@ contract NonfungiblePositionManager is INonfungiblePositionManager, BaseLiquidit
...
+        // positionId = keccak256(abi.encodePacked(owner, tickLower, tickUpper, salt))
+        /// @solidity memory-safe-assembly
+        assembly {
+            mstore(0x26, salt) // [0x26, 0x46)
+            mstore(0x06, tickUpper) // [0x23, 0x26)
+            mstore(0x03, tickLower) // [0x20, 0x23)
+            mstore(0, owner) // [0x0c, 0x20)
+            positionId := keccak256(0x0c, 0x3a) // len is 58 bytes
+            mstore(0x26, 0) // rewrite 0x26 to 0
+        }
```
snreynolds, line 196: This can be a view function outside the main contract IMO.. ie we can have a LiquidityRouterState.sol that people import that just uses StateLibrary.sol - we can leave for now but I prefer all view functions to live outside the main contract

snreynolds (approved):

PR #182, opened 2024-07-21
checkpoint: use pool position

snreynolds:
TODO:

- do NOT merge into dev-posm, this will be a feature ontop to be merged into main
- left out updates to burn test
- this current design would allow us to  combine mint with increase & burn with decrease if we check tokenId == 0
- this current design is probably cheapest, but worst devX 

some other state options though that I'd like to talk about are:
- `tokenId => poolKey => range ` where range is just (tickUpper, tickLower)
- better for DEVX cause increase/decrease only have to pass in poolKey
- just check that the poolKey passed in results in non zero tickLower, tickUpper
- mint/burn can't be combined though
- but probably slightly more gas than what I have currently in PR.

ORRR `tokenId => poolKey => positionId`
- nice bc the positionId could be the same one that v4-core uses.. really nice for state library and looking up current liquidity of the position etc.. 
- mint/burn could probably still be combined bc u still need to pass in tL and tU
- this one is better for view functions/cleanliness with core and lookup of state based on a positionId
- also more expensive than what we have rn


.forge-snapshots/PositionManager_mint.snap
```diff
@@ -1 +1 @@
-384014
\ No newline at end of file
+337758
```
snreynolds, line 1: 🤠

snreynolds: -46K :))))

src/PositionManager.sol
```diff
@@ -83,10 +83,10 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
                 returnData[i] = _mint(params[i]);
             } else if (actions[i] == Actions.CLOSE_CURRENCY) {
                 returnData[i] = _close(params[i], sender);
-            } else if (actions[i] == Actions.BURN) {
-                // TODO: Burn will be coupled with decrease.
-                (uint256 tokenId) = abi.decode(params[i], (uint256));
-                burn(tokenId, sender);
+                // } else if (actions[i] == Actions.BURN) {
```
snreynolds, line 86: didnt wanna fix burn until i got some feedback on the draft.. cause mint/burn could change significantly with this

saucepoint:
to chime in -- i like both alternative options

leaning a bit towards `tokenId => PoolKey => positionId` since the positionId is marginally more useful than tickLower/tickUpper while also giving us the option of combining mint/burn with increase/decrease

src/PositionManager.sol
```diff
@@ -138,9 +140,9 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
         }
         _mint(owner, tokenId);
 
-        (BalanceDelta delta,) = _modifyLiquidity(range, liquidity.toInt256(), bytes32(tokenId), hookData);
+        (BalanceDelta delta,) = _modifyLiquidity(poolPos, liquidity.toInt256(), bytes32(tokenId), hookData);
 
-        tokenRange[tokenId] = range;
+        poolPositions[tokenId] = poolPos.toId();
```
hensha256, line 145: definitely need an event logging all of the fields here imo? Because all that people can look up is a hash

hensha256 (commented):
> ORRR tokenId => poolKey => positionId

This I feel like doesnt improve devx over what youve already got right? bc people still have to pass in both poolkey and tL/tU. So people still need to know those things. PLUS youre adding extra gas cost to calculate 2 extra hashes:
- the position ID
- the double nested mapping.

Imo this one doesnt gain much improvement for integrators? Theres no information that they no longer need to know as part of devx? Unless I'm missing something?

> tokenId => poolKey => range  where range is just (tickUpper, tickLower)

This I do think gains things for integrators at least because theyll only need to know the pool key (and tokenID).

So TLDR I think if you think that devx needs to be improved im pro option 1. If you dont think it needs to be improved im pro as-is. And I dont see any real pro of option 2 unless im missing something lol

snreynolds:
> > ORRR tokenId => poolKey => positionId
> 
> This I feel like doesnt improve devx over what youve already got right? bc people still have to pass in both poolkey and tL/tU. So people still need to know those things. PLUS youre adding extra gas cost to calculate 2 extra hashes:
> 
> * the position ID
> * the double nested mapping.
> 
> Imo this one doesnt gain much improvement for integrators? Theres no information that they no longer need to know as part of devx? Unless I'm missing something?
> 
> > tokenId => poolKey => range  where range is just (tickUpper, tickLower)
> 
> This I do think gains things for integrators at least because theyll only need to know the pool key (and tokenID).
> 
> So TLDR I think if you think that devx needs to be improved im pro option 1. If you dont think it needs to be improved im pro as-is. And I dont see any real pro of option 2 unless im missing something lol

So the reason Option 2 is nice is because the positionId could be the same as the one that core uses. It makes view functions super nice. Not really DevX improvements from input params, but you also could still have shared mint/burn with increase/decrease

snreynolds:
closing to continue convo on https://github.com/Uniswap/v4-periphery/pull/196

PR #186, opened 2024-07-23
CurrencySettler -> DeltaResolver

hensha256:
- Currently the currency settler is very generic which is adding gas. We often just pass `false` or `true` as the final parameter in meaning that really we should just be writing more efficient functions.
- For the rare cases when we truly want to pass in a boolean, the calling contract can handle that
- Adds safe transfer to make the library secure


src/PositionManager.sol
```diff
@@ -154,9 +154,9 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
 
         // the sender is the payer or receiver
         if (currencyDelta < 0) {
-            currency.settle(poolManager, sender, uint256(-int256(currencyDelta)), false);
+            poolManager.settle(currency, sender, uint256(-int256(currencyDelta)));
```
snreynolds, line 157: I kind of preferred doing currency.settle bc its very clear its not actually calling the manager contract but rather its using a library wrapping the manager.settle call

src/libraries/CurrencySettler.sol
```diff
@@ -0,0 +1,34 @@
...
+    /// @param currency Currency to settle
+    /// @param payer Address of the payer, the token sender
+    /// @param amount Amount to send
+    function settle(IPoolManager manager, Currency currency, address payer, uint256 amount) internal {
+        if (currency.isNative()) {
+            manager.settle{value: amount}();
+        } else {
+            manager.sync(currency);
+            if (payer != address(this)) {
+                ERC20(Currency.unwrap(currency)).safeTransferFrom(payer, address(manager), amount);
```
snreynolds, line 27: we should be using permit2 not safeTransferFrom

snreynolds (changes requested):

snreynolds (commented):
We should add tests! Bit odd this has no testing updates needed & since we're modifying it quite a bit from core it should have its own test file imo in periphery

src/libraries/CurrencySettler.sol
hensha256: Honestly imo it should have an abstract `pay` function so the implementing contract can figure out what to do... Thoughts if I make this an abstract contract?

hensha256: Like UR already has all its own payment stuff built in so would just implement `_pay` as a call to its own logic

snreynolds: ok! it would be nice tho for some of that to be shared with posm though

.forge-snapshots/PositionManager_multicall_initialize_mint.snap
```diff
@@ -1 +1 @@
-462111
\ No newline at end of file
+461714
```
hensha256, line 1: i wont claim i understand why its gone down 400 gas 😂

hensha256: i think because we're using safe transfer and that does token transfers in assembly

src/PositionManager.sol
snreynolds: i think this is still not fixed?

src/PositionManager.sol
```diff
@@ -155,12 +159,12 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
 
         // the sender is the payer or receiver
         if (currencyDelta < 0) {
-            currency.settle(poolManager, sender, uint256(-int256(currencyDelta)), false);
+            _settle(currency, sender, uint256(-int256(currencyDelta)));
```
snreynolds, line 162: this is very weird to me that only _settle is included in the DeltaResolver library? can we just define all functions that deal with closing deltas in 1 library?

for take its quite simple but then you dont need to pass poolManager

its just _take(currency, sender, delta)

and then later on we can also add mint/burn in that lib

src/base/DeltaResolver.sol
```diff
@@ -0,0 +1,31 @@
...
+            poolManager.settle();
+        }
+    }
+
+    /// @notice Abstract function for contracts to implement paying tokens to the poolManager
+    /// @dev The recipient of the payment should be the poolManager
+    /// @param token The token to settle. This is known not to be the native currency
+    /// @param payer The address who should pay tokens
+    /// @param amount The number of tokens to send
+    function _pay(Currency token, address payer, uint256 amount) internal virtual;
```
snreynolds, line 30: lol i wish we could call this send ...

src/base/DeltaResolver.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.24;
+
+import {Currency} from "@uniswap/v4-core/src/types/Currency.sol";
+import {ImmutableState} from "./ImmutableState.sol";
+
+/// @notice Abstract contract used to sync, send, and settle funds to the pool manager
+/// @dev Note that sync() is called before any erc-20 transfer in `settle`.
+abstract contract DeltaResolver is ImmutableState {
+    /// @notice Settle (pay) a currency to the PoolManager
```
snreynolds, line 10:
```suggestion
    /// @notice Pay and settle a currency to the PoolManager
```

src/base/DeltaResolver.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.24;
+
+import {Currency} from "@uniswap/v4-core/src/types/Currency.sol";
+import {ImmutableState} from "./ImmutableState.sol";
+
+/// @notice Abstract contract used to sync, send, and settle funds to the pool manager
+/// @dev Note that sync() is called before any erc-20 transfer in `settle`.
+abstract contract DeltaResolver is ImmutableState {
```
snreynolds, line 9: hm the inheriting structure in the base PositionManager contract now inherits two contracts(SafeCallback and this one) that both also inherit ImmutableState? Any way we can avoid that?

snreynolds: maybe we need to expose a virtual function poolManager() that is overridden in the main base contract ?

src/base/DeltaResolver.sol
snreynolds: note that this contract must be approved/permitted to pay on behalf of the payer?

src/PositionManager.sol
```diff
@@ -207,6 +211,12 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
         if (liquidity > 0) revert PositionMustBeEmpty();
     }
 
+    // implementation of abstract function DeltaResolver._pay
+    function _pay(Currency token, address payer, uint256 amount) internal override {
+        // TODO this should use Permit2
+        ERC20(Currency.unwrap(token)).safeTransferFrom(payer, address(poolManager), amount);
```
snreynolds, line 217: wondering if we should also support `transfer` not just transferFrom?

src/PositionManager.sol
snreynolds: i see that this got fixed.. still keft some comments tho abt some changes re DeltaResolver lib below!

src/PositionManager.sol
hensha256: oh this was just a draft function to get posm to functional-parity with where it was before this PR. I'm assuming youll redo the `_pay` implementation to use Permit2.
As for whether posm needs `transfer` (universal router does). That depends if posm will ever custody funds that need to go into another position? I dont think so because you just leave them in pool manager and the deltas handle it all. So I would imagine posm only needs transferFrom unless you have a counter example?

src/base/DeltaResolver.sol
hensha256: >note that this contract must be approved/permitted to pay on behalf of the payer?

not necessarily true though... it could be that this function chooses to pull tokens out of compound to pay the debt, or any other random way they want to pay.

src/PositionManager.sol
snreynolds: we may want to consider supporting transfer? but youre right we dont need it in this PR

snreynolds (commented):
Nice! Looks great

My only final comment is whether we should update core at all with these patterns? IK we have CurrencySettler in core which is mostly just used for testing purposes but I think it was nice that core/periphery had the same "infra" - I believe its why merging core into periphery became easier

snreynolds (approved):

PR #163, opened 2024-07-18
Enable native token settlement

gretzke:
## Description of changes

Allow sending native tokens when calling `modifyLiquidities` to enable settling with native tokens.

PR #189, opened 2024-07-24
posm: Native Token

### LINKED ISSUE #179: [posm]: support eth
### Component

Position Manager

support ETHHHHHHH

_No response_

### Describe alternatives.

_No response_

### Additional context.

_No response_

saucepoint:
## Related Issue
Closes #179 and replaces #163 

## Description of changes

Add native token settlement to posm


src/PositionManager.sol
```diff
@@ -155,6 +156,15 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
         // the sender is the payer or receiver
         if (currencyDelta < 0) {
             currency.settle(poolManager, sender, uint256(-int256(currencyDelta)), false);
+
+            // if there are native tokens left over after settling, return to sender
```
snreynolds, line 160: I feel like this code should live in a sweep function, and it should be on the caller to call sweep? thats what we do in UR

thoughts @hensha256 ?

src/PositionManager.sol
```diff
@@ -155,6 +156,15 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
...
+            if (currency.isNative()) {
+                uint256 nativeBalance = CurrencyLibrary.NATIVE.balanceOfSelf();
+                if (nativeBalance > 0) {
+                    address(sender).call{value: nativeBalance}("");
+                    // CurrencyLibrary.NATIVE.transfer(sender, nativeBalance);
```
snreynolds, line 165: do we not support native transfer in currency library?

snreynolds (commented):
I think we still need permit to be payable whenever we add it

ie if you want to batch call permit(USDC) and then modifyLiquidities(USDC, ETH) i think it will revert bc we are sending eth but permit is not payable

and maybe this needs a test as well?

src/PositionManager.sol
saucepoint, line 165: syntactically we can use either, i believe CurrencyLib has a marginal gas overhead because of if/else condition

snreynolds (commented):
one more request.. could we add mint, increase, decrease eth gas tests?

src/PositionManager.sol
snreynolds: if we don't use it then we should remove from the lib IMO

saucepoint: syntactically we can use either, i believe CurrencyLib has a marginal gas overhead because `.transfer` has an equivalence check + if/else condition

src/PositionManager.sol
saucepoint: from offline chats: we agreed to do it automatically with an internal function

the majority of operations will need to overpay the native token to account for slippage / price movements

snreynolds (approved):
nice! Could you just open up an issue to investigate off by 1s?

PR #193, opened 2024-07-24
Safe transfer ETH in posm

hensha256:
The current code doesnt revert if the ETH refund fails. Best practice to use safe transfer to handle this


.forge-snapshots/PositionManager_mint_nativeWithSweep.snap
```diff
@@ -1 +1 @@
-373221
\ No newline at end of file
+373220
```
marktoda, line 1: wao huge

src/PositionManager.sol
```diff
@@ -197,7 +199,7 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
     /// @param recipient the receiver of the excess native tokens. Should be the caller, the one that sent the native tokens
     function _sweepNativeToken(address recipient) internal {
         uint256 nativeBalance = address(this).balance;
-        if (nativeBalance > 0) address(recipient).call{value: nativeBalance}("");
+        if (nativeBalance > 0) recipient.safeTransferETH(nativeBalance);
```
marktoda, line 202: should we just revert if 0 instead?

marktoda (approved):

src/PositionManager.sol
hensha256: no if it’s 0 that means the sent in the perfect amount of ETH - this is just a refund

PR #164, opened 2024-07-19
Some posm tests

saucepoint:
- [x] restored commented-out tests
- [x] added ERC721 transfer tests for "mint, transfer, and __" flows
- [x] added "rebalance" (burn & mint) test

Note: left out 6909 tests as we're still deciding how to support it..

PR #194, opened 2024-07-24
some posm tests

saucepoint:
## Related Issue
Cherry-picked commits to replace #164 

## Description of changes

- [x] restored commented-out tests
- [x] added ERC721 transfer tests for "mint, transfer, and __" flows
- [x] added "rebalance" (burn & mint) test

Note: left out 6909 tests as we're still deciding how to support it..


test/position-managers/Execute.t.sol
```diff
@@ -119,7 +121,66 @@ contract ExecuteTest is Test, PosmTestSetup, LiquidityFuzzers {
     }
 
     // rebalance: burn and mint
-    function test_execute_rebalance() public {}
+    function test_execute_rebalance_perfect() public {
+        uint256 initialLiquidity = 100e18;
+
+        // mint a position on range [-300, 300]
+        BalanceDelta delta = mint(range, initialLiquidity, address(this), ZERO_BYTES);
+        uint256 tokenId = lpm.nextTokenId() - 1;
```
snreynolds, line 129: bleh.. this is a case for mint to return the tokenId

snreynolds: can you add a comment above that this tokenId is referencing the just minted token? or define it before the mint.. just bit confusing subtracting 1

test/position-managers/Execute.t.sol
```diff
@@ -119,7 +121,66 @@ contract ExecuteTest is Test, PosmTestSetup, LiquidityFuzzers {
...
+
+        bytes[] memory data = lpm.modifyLiquidities(calls, _deadline);
+        int256 delta0 = abi.decode(data[data.length - 2], (int256));
+        int256 delta1 = abi.decode(data[data.length - 1], (int256));
+
+        uint256 balance0After = currency0.balanceOfSelf();
+        uint256 balance1After = currency1.balanceOfSelf();
+
+        // TODO: use clear so user does not pay 1 wei
+        assertApproxEqAbs(delta0, 0, 1 wei);
```
snreynolds, line 158: is this currently 1 wei off? can you assert its exactly 1 wei off, that way this does not pass when we use clear

test/position-managers/Execute.t.sol
```diff
@@ -119,7 +121,66 @@ contract ExecuteTest is Test, PosmTestSetup, LiquidityFuzzers {
...
+        uint256 newTokenId = lpm.nextTokenId() - 1;
+        assertEq(lpm.ownerOf(newTokenId), address(this));
+
+        // new token has expected liquidity
+        positionId =
+            keccak256(abi.encodePacked(address(lpm), newRange.tickLower, newRange.tickUpper, bytes32(newTokenId)));
+        liquidity = manager.getPositionLiquidity(range.poolKey.toId(), positionId);
+        assertEq(liquidity, newLiquidity);
+    }
+
```
snreynolds, line 183: nice i really like this test

test/position-managers/FeeCollection.t.sol
```diff
@@ -68,6 +68,7 @@ contract FeeCollectionTest is Test, PosmTestSetup, LiquidityFuzzers {
         uint256 balance1Before = currency1.balanceOfSelf();
         BalanceDelta delta = collect(tokenId, ZERO_BYTES);
 
+        assertApproxEqAbs(uint256(int256(delta.amount1())), swapAmount.mulWadDown(FEE_WAD), 1 wei);
```
snreynolds, line 71: instead of using swapAmount.mulWadDown can you use getFeesOwed? I think its more precise.. and then try to do an exact assertion?

test/position-managers/FeeCollection.t.sol
```diff
@@ -136,6 +137,107 @@ contract FeeCollectionTest is Test, PosmTestSetup, LiquidityFuzzers {
         assertEq(manager.balanceOf(address(lpm), currency1.toId()), 0);
     }
 
+    function test_collect_donate() public {
+        LiquidityRange memory range = LiquidityRange({poolKey: key, tickLower: -120, tickUpper: 120});
+        mint(range, 10e18, address(this), ZERO_BYTES);
+        uint256 tokenId = lpm.nextTokenId() - 1;
```
snreynolds, line 143: again can you move tokenId declaration above mint, so we dont have to sub 1

test/position-managers/FeeCollection.t.sol
```diff
@@ -136,6 +137,107 @@ contract FeeCollectionTest is Test, PosmTestSetup, LiquidityFuzzers {
...
+        uint256 balance0Before = currency0.balanceOfSelf();
+        uint256 balance1Before = currency1.balanceOfSelf();
+        BalanceDelta delta = collect(tokenId, ZERO_BYTES);
+
+        assertApproxEqAbs(uint256(int256(delta.amount0())), feeRevenue, 1 wei);
+        assertApproxEqAbs(uint256(int256(delta.amount1())), feeRevenue, 1 wei);
+        assertEq(delta.amount0(), expectedFees.amount0());
+        assertEq(delta.amount1(), expectedFees.amount1());
+
+        assertEq(uint256(int256(delta.amount0())), currency0.balanceOfSelf() - balance0Before);
```
snreynolds, line 161: i feel like its more intuitive to me that we assert the balanceBefore + expectedFees equals the balance now

test/position-managers/FeeCollection.t.sol
```diff
@@ -136,6 +137,107 @@ contract FeeCollectionTest is Test, PosmTestSetup, LiquidityFuzzers {
...
+        LiquidityRange memory range = LiquidityRange({poolKey: key, tickLower: -120, tickUpper: 120});
+
+        // alice provisions 3x the amount of liquidity as bob
+        uint256 liquidityAlice = 3000e18;
+        uint256 liquidityBob = 1000e18;
+
+        vm.startPrank(alice);
+        mint(range, liquidityAlice, alice, ZERO_BYTES);
+        vm.stopPrank();
+        uint256 tokenIdAlice = lpm.nextTokenId() - 1;
```
snreynolds, line 176: tokenId up

test/position-managers/FeeCollection.t.sol
```diff
@@ -136,6 +137,107 @@ contract FeeCollectionTest is Test, PosmTestSetup, LiquidityFuzzers {
...
+            assertApproxEqAbs(
+                uint128(expectedFeesAlice.amount0()),
+                feeRevenue0.mulDivDown(liquidityAlice, liquidityAlice + liquidityBob),
+                1 wei
+            );
+            assertApproxEqAbs(
+                uint128(expectedFeesAlice.amount1()),
+                feeRevenue1.mulDivDown(liquidityAlice, liquidityAlice + liquidityBob),
+                1 wei
+            );
```
snreynolds, line 200: these are just asserts checking that expectedFees is calculated correctly? maybe these kinds of tests should live in a unit test for getFeesOwed?

would be cool to fuzz some of this

test/position-managers/PositionManager.t.sol
```diff
@@ -195,13 +199,55 @@ contract PositionManagerTest is Test, PosmTestSetup, LiquidityFuzzers {
...
+        uint256 balance0Before = currency0.balanceOfSelf();
+        uint256 balance1Before = currency1.balanceOfSelf();
+        BalanceDelta delta = decreaseLiquidity(tokenId, decreaseLiquidityDelta, ZERO_BYTES);
 
         bytes32 positionId =
             keccak256(abi.encodePacked(address(lpm), range.tickLower, range.tickUpper, bytes32(tokenId)));
         (uint256 liquidity,,) = manager.getPositionInfo(range.poolKey.toId(), positionId);
+        assertEq(liquidity, uint256(params.liquidityDelta) - decreaseLiquidityDelta);
+
+        assertEq(currency0.balanceOfSelf() - balance0Before, uint128(delta.amount0()));
```
snreynolds, line 211: also find this more intuitive to do current balance = old balance + delta

test/position-managers/PositionManager.t.sol
```diff
@@ -237,6 +283,145 @@ contract PositionManagerTest is Test, PosmTestSetup, LiquidityFuzzers {
         assertEq(currency1.balanceOfSelf() - balance1Before, uint256(int256(delta.amount1())));
     }
 
+    function test_mintTransferBurn() public {
+        LiquidityRange memory range = LiquidityRange({poolKey: key, tickLower: -600, tickUpper: 600});
+        uint256 liquidity = 100e18;
+        BalanceDelta mintDelta = mint(range, liquidity, address(this), ZERO_BYTES);
+        uint256 tokenId = lpm.nextTokenId() - 1;
```
snreynolds, line 290: just noting more tokenId changes

snreynolds: lol i'll stop commenting this but theres a few more places

snreynolds (commented):
love these tests!! just a few comments

test/position-managers/FeeCollection.t.sol
saucepoint: we're asserting delta == feesOwed in the line below, so gonna keep this line as-is since its a nice assertion that feesOwed ~= swapAmount * FEE_WAD

test/position-managers/FeeCollection.t.sol
saucepoint: keeping the assertion since its asserting the size/magnitude of collected fees is equal to the calculated fees and the actual fees transferred (_delta_)

but added `test_fuzz_getFeesOwed_donate` which fuzzes donation revenue against getFeesOwed. there's some imprecision with getFeesOwed 😅

snreynolds (approved):

PR #196, opened 2024-07-24
feat: store PoolPositionId

snreynolds:
## Related Issue

This introduces 2 fairly large changes:
- No longer store the PoolKey, tickLower, and tickUpper inside the position manager. Instead, store a hash of those same variables and force that any action (besides mint) passes in the correct parameters
- Does NOT separate mint and burn, but it does allow calling burn with non 0 position, which will automatically decrease the open position before burning

Note on why I don't think we should separate mint/burn:
Separating mint/burn is potentially possible but it introduces more if/else branching & I think would be less optimal so I am opting against that change. But open to feedback!


.forge-snapshots/PositionManager_mint.snap
```diff
@@ -1 +1 @@
-418214
\ No newline at end of file
+372003
```
snreynolds, line 1: 🔥

src/PositionManager.sol
```diff
@@ -168,67 +168,56 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
...
+            abi.decode(params, (uint256, PoolPosition, bytes));
+
+        _beforeModify(poolPos, tokenId, sender);
+        uint256 liquidity = _beforeBurn(poolPos, tokenId);
 
-        // Checks that the full position's liquidity has been removed and all tokens have been collected from tokensOwed.
-        _validateBurn(tokenId);
+        // Can only call modify if there is non zero liquidity.
+        BalanceDelta delta;
+        if (liquidity > 0) (delta,) = _modifyLiquidity(poolPos, -(liquidity.toInt256()), bytes32(tokenId), hookData);
```
hensha256, line 181: `_beforeBurn` actually returns a uint128. So you dont need to use safecast because liquidity cannot be larger than uint128. Also this negation can be unchecked.

```
uint256 liquidity = ...
int256 liquidityDelta;
unchecked {
    liquidityDelta = -int256(liquidity);
}
```

src/PositionManager.sol
```diff
@@ -168,67 +168,56 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
...
 
-    function _modifyLiquidity(LiquidityRange memory range, int256 liquidityChange, bytes32 salt, bytes memory hookData)
+    function _beforeModify(PoolPosition memory poolPos, uint256 tokenId, address sender) private view {
+        if (!_isApprovedOrOwner(sender, tokenId)) revert NotApproved(sender);
+        if (poolPositions[tokenId] != poolPos.toId()) revert IncorrectPoolPositionForTokenId(tokenId);
+    }
+
+    function _modifyLiquidity(PoolPosition memory poolPos, int256 liquidityChange, bytes32 salt, bytes memory hookData)
         internal
         returns (BalanceDelta liquidityDelta, BalanceDelta totalFeesAccrued)
```
hensha256, line 196: every call to this function only uses the first return parameter... so this internal function should probably just return 1 of them

hensha256 (changes requested):
Generally really like it! All my comments are naming requests or optimisations i think

src/PositionManager.sol
```diff
@@ -17,22 +17,23 @@ import {SafeCallback} from "./base/SafeCallback.sol";
...
 import {CurrencySettleTake} from "./libraries/CurrencySettleTake.sol";
-import {LiquidityRange} from "./types/LiquidityRange.sol";
+import {PoolPosition, PoolPositionLibrary} from "./libraries/PoolPosition.sol";
 
 contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Multicall, SafeCallback {
     using SafeTransferLib for address;
     using CurrencyLibrary for Currency;
     using CurrencySettleTake for Currency;
     using PoolIdLibrary for PoolKey;
+    using PoolPositionLibrary for PoolPosition;
```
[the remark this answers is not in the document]
snreynolds: `PositionConfig` it is

src/PositionManager.sol
```diff
@@ -85,9 +86,8 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
             } else if (actions[i] == Actions.CLOSE_CURRENCY) {
                 returnData[i] = _close(params[i], sender);
             } else if (actions[i] == Actions.BURN) {
-                // TODO: Burn will be coupled with decrease.
-                (uint256 tokenId) = abi.decode(params[i], (uint256));
-                burn(tokenId, sender);
+                // Will automatically decrease if the position is not empty.
```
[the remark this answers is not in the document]
snreynolds: ah thats a great clarification ty!

src/PositionManager.sol
```diff
@@ -168,67 +168,56 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
...
 
-        delete tokenRange[tokenId];
+        delete poolPositions[tokenId];
         // Burn the token.
         _burn(tokenId);
+        return abi.encode(delta);
     }
 
-    function _modifyLiquidity(LiquidityRange memory range, int256 liquidityChange, bytes32 salt, bytes memory hookData)
+    function _beforeModify(PoolPosition memory poolPos, uint256 tokenId, address sender) private view {
```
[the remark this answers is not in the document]
snreynolds: ok! I got ahead of myself bc I think there will be a _afterModifyLiq() for staking purposes

src/interfaces/IPositionManager.sol
```diff
@@ -18,12 +18,13 @@ interface IPositionManager {
...
     error UnsupportedAction();
     error PositionMustBeEmpty();
+    error IncorrectPoolPositionForTokenId(uint256 tokenId);
 
-    // TODO: This will just return a positionId
-    function tokenRange(uint256 tokenId)
-        external
-        view
-        returns (PoolKey memory poolKey, int24 tickLower, int24 tickUpper);
+    /// @notice Maps the ERC721 tokenId to a positionId, which is a keccak256 hash of the positions pool key, range (tickLower, tickUpper)
```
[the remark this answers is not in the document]
snreynolds: changed to configId

src/libraries/LiquidityNegation.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.24;
+
+/// @notice Library for negating a uint256 value.
+/// Used for decreasing liquidity on a pool.
+/// @dev It is safe to negate in an unchecked block because liquidity is always postive (uint256)
+library LiquidityNegation {
+    function toNegativeInt256(uint256 liquidity) internal pure returns (int256 liquidityDelta) {
+        unchecked {
+            liquidityDelta = -int256(liquidity);
```
hensha256, line 10: it still needs to us SafeCast though - the `uint256` -> `int256` isnt safe. Only the negation is

hensha256: Honestly imo not worth having a whole library for an unchecked subtraction. Can just go back to how it was if you prefer (not unchecked) , I jumping to another function call will probably undo the benefit.

src/PositionManager.sol
```diff
@@ -168,73 +172,71 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
         return abi.encode(currencyDelta);
     }
 
-    function burn(uint256 tokenId, address sender) internal {
-        _requireApprovedOrOwner(tokenId, sender);
+    /// note: this is overloaded with ERC721Permit._burn
```
hensha256, line 175: @ dev ?

src/libraries/LiquidityNegation.sol
snreynolds: a jump is pretty cheap I thought? I was just trying to get rid of the ugliness with inline unchecked

test/position-managers/PositionManager.t.sol
```diff
@@ -184,6 +188,58 @@ contract PositionManagerTest is Test, PosmTestSetup, LiquidityFuzzers {
...
+            TickMath.getSqrtPriceAtTick(params.tickLower),
+            TickMath.getSqrtPriceAtTick(params.tickUpper),
+            uint128(int128(params.liquidityDelta))
+        );
+
+        // burn liquidity
+        uint256 balance0BeforeBurn = currency0.balanceOfSelf();
+        uint256 balance1BeforeBurn = currency1.balanceOfSelf();
+
+        // No decrease will happen on the burn call so the delta will be 0.
```
hensha256, line 221: Incorrect comment

hensha256 (approved):

PR #204, opened 2024-07-25
Reentrancy lock posm

### LINKED ISSUE #170: [posm]: use transient reentrancy lock
### Component

Position Manager

We need to lock calls to `modifyLiquidities` because some commands query the full delta balance on the pool manager. this assumes that all checks on the balance have been done throughout the duration of the call, so reentering and updating that balance would break this assumption.

_No response_

### Describe alternatives.

_No response_

### Additional context.

_No response_

hensha256:
Closes #170


src/PositionManager.sol
```diff
@@ -49,75 +50,66 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
...
 
     function _unlockCallback(bytes calldata payload) internal override returns (bytes memory) {
-        // TODO: Fix double encode/decode
-        (bytes memory unlockData, address sender) = abi.decode(payload, (bytes, address));
+        (Actions[] memory actions, bytes[] memory params) = abi.decode(payload, (Actions[], bytes[]));
 
-        (Actions[] memory actions, bytes[] memory params) = abi.decode(unlockData, (Actions[], bytes[]));
-
-        bytes[] memory returnData = _dispatch(actions, params, sender);
+        bytes[] memory returnData = _dispatch(actions, params);
```
snreynolds, line 64: love

src/PositionManager.sol
```diff
@@ -147,29 +139,29 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
     }
 
     /// @param params is an encoding of the Currency to close
-    /// @param sender is the msg.sender encoded by the `modifyLiquidities` function before the `unlockCallback`.
-    /// @return an encoding of int256 the balance of the currency being settled by this call
-    function _close(bytes memory params, address sender) internal returns (bytes memory) {
+    /// @return btyes an encoding of int256 the balance of the currency being settled by this call
```
snreynolds, line 142:
```suggestion
    /// @return bytes an encoding of int256 the balance of the currency being settled by this call
```

src/PositionManager.sol
```diff
@@ -147,29 +139,29 @@ contract PositionManager is IPositionManager, ERC721Permit, PoolInitializer, Mul
...
+    function _close(bytes memory params) internal returns (bytes memory) {
         (Currency currency) = abi.decode(params, (Currency));
         // this address has applied all deltas on behalf of the user/owner
         // it is safe to close this entire delta because of slippage checks throughout the batched calls.
         int256 currencyDelta = poolManager.currencyDelta(address(this), currency);
 
-        // the sender is the payer or receiver
+        // the locker is the payer or receiver
+        address caller = _getLocker();
```
snreynolds, line 150: with BaseActionsRouter this should be _msgSender() yes?

test/position-managers/PositionManager.t.sol
```diff
@@ -54,6 +56,23 @@ contract PositionManagerTest is Test, PosmTestSetup, LiquidityFuzzers {
         lpm.modifyLiquidities(abi.encode(planner.actions, badParams), block.timestamp + 1);
     }
 
+    function test_modifyLiquidities_reverts_reentrancy() public {
+        // Create a reentrant token and initialize the pool
+        ReentrantToken reentrantToken = new ReentrantToken(lpm);
+        (currency0, currency1) = (address(reentrantToken) < Currency.unwrap(currency1))
```
snreynolds, line 62: could we underscore these vars.. just cause we also have them defined at the top level contract & same with `key`

src/PositionManager.sol
hensha256: yeah youll implement
```
function _msgSender {
   return _getLocker()
}
```

test/position-managers/PositionManager.t.sol
hensha256: oh i was overriding them on purpose ... they reset their value at the beginning of every test 
but happy to bring in new vars if you want

snreynolds: yeah its just more readable in the sense that we know they are different than the global params? but honestly its fine as is, its not a harmful test

snreynolds (approved):

PR #207, opened 2024-07-25
remove isApprovedOrOwner from increase

dianakocsis:
## Related Issue
anyone can increase anyone else's position - they don't need to be the owner or approved

## Description of changes
remove isApprovedOrOwner check from increase


test/position-managers/IncreaseLiquidity.t.sol
```diff
@@ -154,7 +154,8 @@ contract IncreaseLiquidityTest is Test, PosmTestSetup, Fuzzers {
         uint256 balance0BeforeAlice = currency0.balanceOf(alice);
         uint256 balance1BeforeAlice = currency1.balanceOf(alice);
 
-        vm.startPrank(alice);
+        // bob can increase liquidity for alice even though he is not the owner / not approved
```
snreynolds, line 157: could you add a new test for this? so just keep this test the exact same and add a new one, it could do probably even simpler asserts (ie just that alices liquyidity goes up by the exact amount) but pranks bob

snreynolds (dismissed):

test/position-managers/IncreaseLiquidity.t.sol
```diff
@@ -164,6 +165,53 @@ contract IncreaseLiquidityTest is Test, PosmTestSetup, Fuzzers {
         assertApproxEqAbs(balance1BeforeAlice, currency1.balanceOf(alice), tolerance);
     }
 
+    function test_increaseLiquidity_withUnapprovedCaller() public {
+        // Alice and Bob provide liquidity on the range
+        // Alice uses her exact fees to increase liquidity (compounding)
```
hensha256, line 170: ?? does she?

hensha256: i feel like this is quite a complicated test to just test something thats not as complicated?
might be easier to have just 1 person providing liquidity not 2.
Just like
- Call 1: Alice mints herself a position
- Call 2: Bob increases liquidity in alices position
No need for Bob to have a position in the same range that feels like its confusing it a bit

hensha256 (approved):

PR #208, opened 2024-07-26
feat: Use permit2 transferFrom

snreynolds:
Couple thoughts:
- Need to test with signature based approvals on permit2 allowance transfer contract
- Need to add a selfPermitWithPermit2 function that accepts a signature to permit posm on permit2. (Similar to SelfPermit on v3 nft manager i believe) but basically this just lets you batch an approval with whatever add/remove liq you want to do
- Do we need to support sig transfer?
- Possibly still support vanilla transfer in pay


snreynolds:
@hensha256 Not sure why CI is failing, it all passes locally looks like its an issue with imports / doesn't know what snapLastCall is in v4-core test 😢

test/shared/PosmTestSetup.sol
```diff
@@ -9,30 +9,60 @@ import {Deployers} from "@uniswap/v4-core/test/utils/Deployers.sol";
...
+        _approvePermit2AsASpender();
+        // 2. Then, the caller must approve POSM as a spender of permit2. TODO: This could also be a signature.
+        _approvePosmAsASpender();
+    }
+
+    function approvePosmNative() public {
+        // Assumes currency0 is the native token so only execute approvals for currency1.
+        IERC20(Currency.unwrap(currency1)).approve(address(permit2), type(uint256).max);
+        permit2.approve(Currency.unwrap(currency1), address(lpm), type(uint160).max, type(uint48).max);
     }
```
saucepoint, line 49: Should probably be named something more descriptive.. `approvePosmCurrency1()`?

saucepoint (dismissed):
small nit, but otherwise lgtm

as for the open questions -- i think supporting signature transfer for UX (and vanilla transfer for gas) would be nice

the design might be ugly where we need to support multiple CLOSE commands, each using different payment flows?

test/shared/PosmTestSetup.sol
```diff
@@ -9,30 +9,60 @@ import {Deployers} from "@uniswap/v4-core/test/utils/Deployers.sol";
...
+        _approvePosmAsASpender();
         vm.stopPrank();
     }
+
+    function _approvePermit2AsASpender() internal {
+        IERC20(Currency.unwrap(currency0)).approve(address(permit2), type(uint256).max);
+        IERC20(Currency.unwrap(currency1)).approve(address(permit2), type(uint256).max);
+    }
+
+    function _approvePosmAsASpender() internal {
```
dianakocsis, line 64: small nit: `_approvePosmAsSpenderOfPermit2()`

dianakocsis (changes requested):

src/PositionManager.sol
```diff
@@ -227,7 +231,7 @@ contract PositionManager is
 
     // implementation of abstract function DeltaResolver._pay
     function _pay(Currency token, address payer, uint256 amount) internal override {
-        // TODO this should use Permit2
-        ERC20(Currency.unwrap(token)).safeTransferFrom(payer, address(poolManager), amount);
+        // TODO: Should we also support direct transfer?
```
dianakocsis, line 234: in v3 i think it was necessary because of exact input multihop case. what would be a use case for it now?

test/shared/PosmTestSetup.sol
```diff
@@ -9,30 +9,60 @@ import {Deployers} from "@uniswap/v4-core/test/utils/Deployers.sol";
...
         IERC20(Currency.unwrap(currency1)).transfer(to, STARTING_USER_BALANCE);
     }
 
     function approvePosm() public {
-        IERC20(Currency.unwrap(currency0)).approve(address(lpm), type(uint256).max);
-        IERC20(Currency.unwrap(currency1)).approve(address(lpm), type(uint256).max);
+        // Because POSM uses permit2, we must execute 2 permits/approvals.
+        // 1. First, the caller must approve permit2 on the token.
+        _approvePermit2AsASpender();
+        // 2. Then, the caller must approve POSM as a spender of permit2. TODO: This could also be a signature.
```
dianakocsis, line 41: also test signature batch

src/PositionManager.sol
dianakocsis: also in UR we dont support sig transfer so maybe we dont need to here

test/shared/PosmTestSetup.sol
```diff
@@ -9,30 +9,60 @@ import {Deployers} from "@uniswap/v4-core/test/utils/Deployers.sol";
...
         vm.stopPrank();
     }
+
+    function _approvePermit2AsASpender() internal {
+        IERC20(Currency.unwrap(currency0)).approve(address(permit2), type(uint256).max);
+        IERC20(Currency.unwrap(currency1)).approve(address(permit2), type(uint256).max);
+    }
+
+    function _approvePosmAsASpender() internal {
+        permit2.approve(Currency.unwrap(currency0), address(lpm), type(uint160).max, type(uint48).max);
```
dianakocsis, line 65: will then probably be switched/added to calling selfPermitWithPermit2

snreynolds: I'm gonna do selfPermitWithPermit2 in a sep PR ! so will add sig based tests also in that PR

test/shared/PosmTestSetup.sol
snreynolds: Will test this with selfPermitWithPermit2

src/PositionManager.sol
snreynolds: I think we do want to support transfer. Then UR can do 

removeLiquidity on v3 but send amounts directly to posm
and then call settle with transfer & not have to use permit2

if UR uses permit2 it has to approve permit2 which is not ideal i think

dianakocsis: got it! currently right now the payer is always the locker
```
        // the locker is the payer or receiver
        address caller = _getLocker();
        if (currencyDelta < 0) {
            _settle(currency, caller, uint256(-currencyDelta));
```
maybe there could be an extra param that specifies whether the payer is the address or the contract

dianakocsis: this was an issue in v3 periphery where the payer always had to be the msg.sender when adding liquidity
```abi.encode(MintCallbackData({poolKey: poolKey, payer: msg.sender}))```
https://github.com/Uniswap/v3-periphery/blob/main/contracts/base/LiquidityManagement.sol#L85

Comment (hensha256) [reply src/PositionManager.sol:234]: yeah imo this will be 2 separate commands in posm but itll be interesting to see how they integrate into deltaresolver together... will you just use the same internal _pay function for both?
but either way that doesnt matter on this PR, just on the one where you do it haha

File: test/position-managers/PositionManager.t.sol [hunk 6]
```diff
@@ -71,8 +72,8 @@ contract PositionManagerTest is Test, PosmTestSetup, LiquidityFuzzers {
         PositionConfig memory config = PositionConfig({poolKey: key, tickLower: 0, tickUpper: 60});
         bytes memory calls = getMintEncoded(config, 1e18, address(this), "");
 
-        // SafeTransferLib does not bubble the ContractLocked error and instead reverts with its own error
-        vm.expectRevert("TRANSFER_FROM_FAILED");
+        // Permit2.transferFrom does not bubble the ContractLocked error and instead reverts with its own error
+        vm.expectRevert(abi.encodeWithSelector(IAllowanceTransfer.AllowanceExpired.selector, 0));
```
hensha256, line 76: vm.expectRevert("TRANSFER_FROM_FAILED");
mmm this is no longer reverting for reentrancy problems if its reventing with allowance expired. To test reentrancy you need to give permit2 allowance to the reentrant token - so that permit2 doesnt prevent the transfer. It should still be failing with `TRANSFER_FROM_FAILED` from solmate

hensha256: If you run this test with -vvvvv youll see in the stack trace that theres never reentrancy because permit2 rejects the transfer before it reenters

test/shared/PosmTestSetup.sol
```diff
@@ -9,30 +9,60 @@ import {Deployers} from "@uniswap/v4-core/test/utils/Deployers.sol";
...
+        // 1. First, the caller must approve permit2 on the token.
+        _approvePermit2AsASpender();
+        // 2. Then, the caller must approve POSM as a spender of permit2. TODO: This could also be a signature.
+        _approvePosmAsASpenderOfPermit2();
+    }
+
+    function approvePosmCurrency1() public {
+        // Assumes currency0 is the native token so only execute approvals for currency1.
+        IERC20(Currency.unwrap(currency1)).approve(address(permit2), type(uint256).max);
+        permit2.approve(Currency.unwrap(currency1), address(lpm), type(uint160).max, type(uint48).max);
```
hensha256, line 48: feels like this file would be neater with a function thats like
```
approvePosmCurrency(Currency currency) {
    IERC20.approve
    permit2.approve
}
```

then `approvePosmCurrency1` can just call it with currency1, and `approvePosm` can call it with both currency0 and currency1? Currently theres 3 different lines of code for approving permit2

hensha256 (commented):
couple of small things

test/position-managers/PositionManager.t.sol
hensha256: https://github.com/Uniswap/v4-periphery/pull/208/commits/d18599ef5bc41f3af76b29f48003fe2b0b687968

proof of fix:
Screenshot 2024-07-29 at 18 50 54

src/base/PositionManagerImmutableState.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.19;
+
+import {IAllowanceTransfer} from "permit2/src/interfaces/IAllowanceTransfer.sol";
+
+contract PositionManagerImmutableState {
```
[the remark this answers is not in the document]
snreynolds: Was just following a pattern but can do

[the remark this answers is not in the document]
snreynolds: rn ImmutableState also only has 1 var tho..

dianakocsis (approved):

hensha256 (approved):

PR #205, opened 2024-07-25
[Draft] ERC721 Permit Changes

saucepoint:
## Related Issue
Which issue does this pull request resolve?

## Description of changes

PR #210, opened 2024-07-26
ERC721Permit

### LINKED ISSUE #167: [posm]: Add Permit functionality to ERC721
### Component

Position Manager

Permit should: 
- [x] Use unordered nonces
- [x] Not be payable (in V3 it is payable)
- [ ] Use relevant libraries and base contracts (ie a PermitHash [library](https://github.com/Uniswap/v4-periphery/pull/153#discussion_r1681402782)) and a EIP1271 base contract that caches the domain separator. (see comments [here](https://github.com/Uniswap/v4-periphery/pull/153#discussion_r1681374875).
- [x] Constants above functions
- [x] Test: Ensure PERMIT_TYPEHASH is equal to expected hash
- [x] Use custom errors
- [x] replace magic constant with IERC1271.isValidSignature.selector
- [x] [this](https://github.com/Uniswap/v4-periphery/pull/141#discussion_r1684211300) comment

Ensure that all comments related to permit from [this PR](https://github.com/Uniswap/v4-periphery/pull/153) and [this PR](https://github.com/Uniswap/v4-periphery/pull/141) are resolved in the final feature PR.

_No response_

### Describe alternatives.

_No response_

### Additional context.

_No response_

saucepoint:
## Related Issue
Closes #167 and replaces https://github.com/Uniswap/v4-periphery/pull/205

## Description of changes

Adds ERC721Permit with

- [x] Solmate ERC721
- [x] OZ EIP712.sol
- [x] Unordered Nonces


src/base/ERC721Permit.sol
```diff
@@ -2,19 +2,93 @@
...
+        view
+        returns (bytes32 digest)
+    {
+        digest = keccak256(
+            abi.encodePacked(
+                "\x19\x01",
+                DOMAIN_SEPARATOR(),
+                keccak256(abi.encode(PERMIT_TYPEHASH, spender, tokenId, _nonce, deadline))
+            )
+        );
```
saucepoint, line 87:
```suggestion
        digest =  _hashTypedDataV4(keccak256(abi.encode(PERMIT_TYPEHASH, spender, tokenId, _nonce, deadline)));
```

using OZ's EIP712 internal function..

src/base/ERC721Permit.sol
```diff
@@ -2,19 +2,79 @@
...
+            if (recoveredAddress == address(0)) revert InvalidSignature();
+            if (recoveredAddress != owner) revert Unauthorized();
+        }
+
+        _useUnorderedNonce(owner, nonce);
+        _approve(owner, spender, tokenId);
+    }
+
+    function approve(address spender, uint256 id) public override {
+        // override Solmate's ERC721 approve so approve() and permit() share the same code paths
```
hensha256, line 62:
```suggestion
        // override Solmate's ERC721 approve so approve() and permit() share the _approve method
```

src/PositionManager.sol
```diff
@@ -59,6 +59,11 @@ contract PositionManager is
         _;
     }
 
+    // TODO: implement
```
hensha256, line 62: given this wont be implemented before audits i think its worth instead putting it in the constructor parameters? or a clearer comment about when it will be implemented?

src/base/ERC721Permit.sol
```diff
@@ -2,19 +2,79 @@
...
+            address recoveredAddress = ecrecover(digest, v, r, s);
+            if (recoveredAddress == address(0)) revert InvalidSignature();
+            if (recoveredAddress != owner) revert Unauthorized();
+        }
+
+        _useUnorderedNonce(owner, nonce);
+        _approve(owner, spender, tokenId);
+    }
+
+    function approve(address spender, uint256 id) public override {
```
hensha256, line 61: natspec?

src/base/ERC721Permit.sol
hensha256: I'd also put this ocmment above the function as an `@ dev`

src/base/UnorderedNonce.sol
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.24;
+
+contract UnorderedNonce {
```
hensha256, line 4: natspec

test/shared/LiquidityOperations.sol
```diff
@@ -93,6 +93,31 @@ abstract contract LiquidityOperations is CommonBase {
         return abi.decode(result[0], (BalanceDelta));
     }
 
+    // TODO: organize somewhere else, or rename this file to NFTLiquidityHelpers?
```
hensha256, line 96: i think just rename! I had a similar thing with routing so i just had 1 file called `RoutingTestHelpers` i think do something similar! `PosmTestHelpers`?

test/shared/LiquidityOperations.sol
```diff
@@ -93,6 +93,31 @@ abstract contract LiquidityOperations is CommonBase {
...
+    function permit(address signer, uint256 privateKey, uint256 tokenId, address operator, uint256 nonce) internal {
+        bytes32 digest = getDigest(operator, tokenId, 1, block.timestamp + 1);
+
+        (uint8 v, bytes32 r, bytes32 s) = vm.sign(privateKey, digest);
+
+        vm.prank(signer);
```
hensha256, line 102: imo the whole point in a permit is that the signer often isnt the person who sends it. I think you should have a parameter of this function `caller` and do `vm.prank(caller)`. Then some tests you can pass in `signer` and others you can explicitly test that someone else can pass it in

hensha256: e.g. a test where the `caller` is the `operator`!

test/position-managers/PositionManager.multicall.t.sol
```diff
@@ -67,4 +81,35 @@ contract PositionManagerMulticallTest is Test, PosmTestSetup, LiquidityFuzzers {
...
+        bytes32 digest = getDigest(bob, tokenId, nonce, block.timestamp + 1);
+        (uint8 v, bytes32 r, bytes32 s) = vm.sign(alicePK, digest);
+
+        // bob gives himself permission and decreases liquidity
+        bytes[] memory calls = new bytes[](2);
+        calls[0] = abi.encodeWithSelector(
+            PositionManager(lpm).permit.selector, bob, tokenId, block.timestamp + 1, nonce, v, r, s
+        );
+        uint256 liquidityToRemove = 0.4444e18;
+        bytes memory actions = getDecreaseEncoded(tokenId, config, 0.4444e18, ZERO_BYTES);
```
hensha256, line 104:
```suggestion
        bytes memory actions = getDecreaseEncoded(tokenId, config, liquidityToRemove, ZERO_BYTES);
```

test/shared/LiquidityOperations.sol
hensha256: ok i see youre already testing that - just not in this helper function. Thats fair. I _think_ I'd still rather this was `operator` not `signer` pranking just because the `signer` wouldnt do this - they'd just call `approve` instead

test/position-managers/Permit.t.sol
```diff
@@ -0,0 +1,313 @@
...
+        seedBalance(bob);
+
+        approvePosmFor(alice);
+        approvePosmFor(bob);
+
+        // define a reusable range
+        config = PositionConfig({poolKey: key, tickLower: -300, tickUpper: 300});
+    }
+
+    function test_permitTypeHash() public view {
```
hensha256, line 60: would love a similar test to this but for the domain separator

test/position-managers/Permit.t.sol
```diff
@@ -0,0 +1,313 @@
...
+
+    function test_permit_decreaseLiquidity() public {
+        uint256 liquidityAlice = 1e18;
+        vm.prank(alice);
+        mint(config, liquidityAlice, alice, ZERO_BYTES);
+        uint256 tokenIdAlice = lpm.nextTokenId() - 1;
+
+        // alice gives bob operator permissions
+        permit(alice, alicePK, tokenIdAlice, bob, 1);
+
```
hensha256, line 103: I feel like this file is missing specific tests of _permit_. Its testing "if you permit, X action works", but i feel like the actual state changes of permit are missing? Like a test that does a permit and checks

- The nonce wasnt set before, and then is set afterwards
- Bob wasnt approved, and then is afterwards
- Bob didnt get approved for all
- if you do another permit afterwards the old approval is wiped

You've overwritten the `approve` function, so that function needs testing too. Or if youd rather keep it simple you could remove the override of `approve` and just use the `_approve` function for `permit`

src/base/UnorderedNonce.sol
```diff
@@ -0,0 +1,30 @@
...
+    error NonceAlreadyUsed();
+
+    mapping(address owner => mapping(uint256 word => uint256 bitmap)) public nonces;
+
+    /// @notice Consume a nonce, reverting if its already been used
+    /// @param owner address, the owner/signer of the nonce
+    /// @param nonce uint256, the nonce to consume
+    function _useUnorderedNonce(address owner, uint256 nonce) internal {
```
hensha256, line 12: would love tests of this function in particular - does it only set the bit in question, does it revert if the nonce was already taken?
if you look in core at `test_fuzz_flipTick`, something like that to check that `_useUnorderedNonce` always sets the right bit, doesnt set others, and reverts always when it was already set?

hensha256 (changes requested):
Looks great! Mostly just want more testing!

src/base/UnorderedNonce.sol
snreynolds: I think there are these exact fuzz tests in permit2 if you want to also transfer them over @saucepoint

src/base/ERC721Permit.sol
```diff
@@ -2,19 +2,79 @@
...
+        payable
+        override
+    {
+        if (block.timestamp > deadline) revert DeadlineExpired();
+
+        address owner = ownerOf(tokenId);
+        if (spender == owner) revert NoSelfPermit();
+
+        bytes32 digest = _hashTypedDataV4(keccak256(abi.encode(PERMIT_TYPEHASH, spender, tokenId, nonce, deadline)));
+
```
snreynolds, line 43: I think we could really improve readability here a lot by moving this logic into a signature verification library..

really all this should look like in this code path is

```
bytes memory signature = ...
bytes32 permitHash = Permit.hash(spender, tokenId, nonce, deadline)
signature.verify(_hashTypedDataV4(permitHash), owner);  // also does all the support for IERC1271, and the ecrecover logic etc..
```

can reference also the permit2 signature library as well should be pretty similar

test/position-managers/Permit.t.sol
```diff
@@ -0,0 +1,313 @@
...
+import {IERC721Permit} from "../../src/interfaces/IERC721Permit.sol";
+import {ERC721Permit} from "../../src/base/ERC721Permit.sol";
+import {UnorderedNonce} from "../../src/base/UnorderedNonce.sol";
+
+import {PositionConfig} from "../../src/libraries/PositionConfig.sol";
+import {IPositionManager} from "../../src/interfaces/IPositionManager.sol";
+
+import {PosmTestSetup} from "../shared/PosmTestSetup.sol";
+
+contract PermitTest is Test, PosmTestSetup {
```
snreynolds, line 24: can we call this PermitERC721 ? Cause we will also have a permit for permit2 exposed

test/position-managers/Permit.t.sol
```diff
@@ -0,0 +1,313 @@
...
+    using FixedPointMathLib for uint256;
```
snreynolds, line 25: & also would be great to have a suite of unit tests for approve

test/position-managers/Permit.t.sol
saucepoint: will rename this at the end before merge to preserve reviews / diffs

src/base/ERC721Permit.sol
```diff
@@ -2,19 +2,73 @@
...
+import {UnorderedNonce} from "./UnorderedNonce.sol";
+
+/// @title ERC721 with permit
+/// @notice Nonfungible tokens that support an approve via signature, i.e. permit
+abstract contract ERC721Permit is ERC721, IERC721Permit, EIP712, UnorderedNonce {
+    using SignatureVerification for bytes;
+    /// @inheritdoc IERC721Permit
+    /// @dev Value is equal to keccak256("Permit(address spender,uint256 tokenId,uint256 nonce,uint256 deadline)");
+
+    bytes32 public constant override PERMIT_TYPEHASH = ERC721PermitHashLibrary.PERMIT_TYPEHASH;
```
snreynolds, line 20: do we need to declare this here? don't see it used internally

src/base/ERC721Permit.sol
```diff
@@ -2,19 +2,73 @@
 pragma solidity ^0.8.24;
 
 import {ERC721} from "solmate/src/tokens/ERC721.sol";
+import {EIP712} from "@openzeppelin/contracts/utils/cryptography/EIP712.sol";
+import {Address} from "@openzeppelin/contracts/utils/Address.sol";
```
snreynolds, line 6: remove

src/base/ERC721Permit.sol
```diff
@@ -2,19 +2,73 @@
...
+import {IERC721Permit} from "../interfaces/IERC721Permit.sol";
+import {UnorderedNonce} from "./UnorderedNonce.sol";
+
+/// @title ERC721 with permit
+/// @notice Nonfungible tokens that support an approve via signature, i.e. permit
+abstract contract ERC721Permit is ERC721, IERC721Permit, EIP712, UnorderedNonce {
+    using SignatureVerification for bytes;
+    /// @inheritdoc IERC721Permit
+    /// @dev Value is equal to keccak256("Permit(address spender,uint256 tokenId,uint256 nonce,uint256 deadline)");
+
```
snreynolds, line 19: random space?

src/base/ERC721Permit.sol
```diff
@@ -2,19 +2,73 @@
...
+    bytes32 public constant override PERMIT_TYPEHASH = ERC721PermitHashLibrary.PERMIT_TYPEHASH;
+
+    /// @notice Computes the nameHash and versionHash
+    constructor(string memory name_, string memory symbol_, string memory version_)
+        ERC721(name_, symbol_)
+        EIP712(name_, version_)
+    {}
+
+    /// @inheritdoc IERC721Permit
+    function DOMAIN_SEPARATOR() public view returns (bytes32) {
```
snreynolds, line 29: imo this should live in the `EIP712` contract

src/base/ERC721Permit.sol
```diff
@@ -2,19 +2,73 @@
...
+        _approve(owner, spender, tokenId);
+    }
+
+    /// @notice Change or reaffirm the approved address for an NFT
+    /// @dev override Solmate's ERC721 approve so approve() and permit() share the _approve method
+    /// The zero address indicates there is no approved address
+    /// Throws error unless `msg.sender` is the current NFT owner,
+    /// or an authorized operator of the current owner.
+    /// @param spender The new approved NFT controller
+    /// @param id The tokenId of the NFT to approve
```
snreynolds, line 56: inheritdoc?

snreynolds (commented):
ok thoughts on not using openzeppelin contracts at all here.. i think so far its now just EIP712 which gives me a headache looking at... can we just replace w our own?

src/base/ERC721Permit.sol
saucepoint: it was part of IERC721Permit, from v3

not sure if backend/interface depends on it 🤔

hensha256: i feel like its not worth an rpc call to look this up we can just calculate it offchain? i say remove

src/base/UnorderedNonce.sol
```diff
@@ -0,0 +1,32 @@
...
+        (uint256 wordPos, uint256 bitPos) = bitmapPositions(nonce);
+        uint256 bit = 1 << bitPos;
```
hensha256, line 16: technically we have a gas optimised version of this in core if it interests you
https://github.com/Uniswap/v4-core/blob/799dd2cb980319a8d3b827b6a7aa59a606634553/src/libraries/TickBitmap.sol#L48-L52

test/ERC721Permit.t.sol
```diff
@@ -0,0 +1,292 @@
...
+        assertEq(erc721Permit.getApproved(tokenId), address(0));
+        vm.startPrank(operator);
+        vm.expectEmit(true, true, true, true, address(erc721Permit));
+        emit IERC721.Approval(address(this), spender, tokenId);
+        erc721Permit.approve(spender, tokenId);
+        vm.stopPrank();
+        assertEq(erc721Permit.getApproved(tokenId), spender);
+    }
+
+    function test_approve_unauthorizedRevert(address caller) public {
```
hensha256, line 54: these tests al need `_fuzz` in the name

test/UnorderedNonce.t.sol
```diff
@@ -0,0 +1,113 @@
...
+            unorderedNonce.spendNonce(address(this), second);
+        }
+    }
+
+    function testInvalidateNoncesRandomly(uint248 wordPos, uint256 mask) public {
+        unorderedNonce.invalidateUnorderedNonces(wordPos, mask);
+        assertEq(mask, unorderedNonce.nonces(address(this), wordPos));
+    }
+
+    function testInvalidateTwoNoncesRandomly(uint248 wordPos, uint256 startBitmap, uint256 mask) public {
```
hensha256, line 99: i feel like this test doesnt actually check the unordered nonce contract? it just tests the `invalidateUnorderedNonces` function which doesnt use UnorderedNonce it just implements something custom?

e.g. there can easily be overlap between `startBitmap` and `mask` having the same bit set... and that doesnt revert because its not actually using the nonce checking.

Honestly i just think this test and the one above can be deleted? The test `testInvalidateNonzeroWord` had a nice use of this function to then check that used nonces reverted

hensha256 (commented):
lots of really nice testing in here 👯

src/base/UnorderedNonce.sol
saucepoint: took at crack at this, so please review closely ❤️

src/base/ERC721Permit.sol
saucepoint: changed it to external, only reason why this (and PERMIT_TYPEHASH) were public because of IERC721Permit

if we're in agreement, I can change the underlying interface

test/mocks/MockUnorderedNonce.sol
```diff
@@ -0,0 +1,15 @@
...
+
+import {UnorderedNonce} from "../../src/base/UnorderedNonce.sol";
+
+contract MockUnorderedNonce is UnorderedNonce {
+    function spendNonce(address owner, uint256 nonce) external {
+        _useUnorderedNonce(owner, nonce);
+    }
+
+    /// @dev Bulk-spend nonces on a single word. FOR TESTING ONLY
+    function invalidateUnorderedNonces(uint256 wordPos, uint256 mask) external {
```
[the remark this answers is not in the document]
hensha256: wait sorry i meant _this_ function feels like a `batchSpendNonces`? the one above is a `spendNonce` it just uses one right?

src/interfaces/IERC721Permit.sol
```diff
@@ -18,10 +16,8 @@ interface IERC721Permit {
     /// @param spender The account that is being approved
     /// @param tokenId The ID of the token that is being approved for spending
     /// @param deadline The deadline timestamp by which the call must be mined for the approve to work
-    /// @param v Must produce valid secp256k1 signature from the holder along with `r` and `s`
-    /// @param r Must produce valid secp256k1 signature from the holder along with `v` and `s`
-    /// @param s Must produce valid secp256k1 signature from the holder along with `r` and `v`
-    function permit(address spender, uint256 tokenId, uint256 deadline, uint256 nonce, uint8 v, bytes32 r, bytes32 s)
+    /// @param signature Concatenated data from a valid secp256k1 signature from the holder, i.e. abi.encodePacked(r, s, v)
+    function permit(address spender, uint256 tokenId, uint256 deadline, uint256 nonce, bytes calldata signature)
```
hensha256, line 20: maybe worth an @dev to say why its payable lol

hensha256: sorry again kenny

src/base/UnorderedNonce.sol
hensha256: it looks great!!! 👯

src/interfaces/IERC721Permit.sol
snreynolds: wait I added this in my other PR

snreynolds: where u left a similar comment haah, so i got this one

src/base/UnorderedNonce.sol
```diff
@@ -12,21 +12,42 @@ contract UnorderedNonce {
...
+            mstore(0x20, bitmap.slot)
+            let slot := keccak256(0, 0x40)
+
+            // uint256 previousBits = bitmap[wordPos]
+            let previousBits := sload(slot)
+
+            // revert if it's already been used
+            if eq(and(previousBits, bit), bit) {
+                mstore(0, 0x1fb09b80) // 4 bytes of NonceAlreadyUsed.selector
+                revert(0x1c, 0x04)
```
snreynolds, line 46: if we're going to use .revertWith lib from core this will need to be changed

src/base/UnorderedNonce.sol
```diff
@@ -12,21 +12,42 @@ contract UnorderedNonce {
...
-    /// @notice Returns the index of the bitmap and the bit position within the bitmap. Used for unordered nonces
-    /// @param nonce The nonce to get the associated word and bit positions
-    /// @return wordPos The word position or index into the nonceBitmap
-    /// @return bitPos The bit position
-    /// @dev The first 248 bits of the nonce value is the index of the desired bitmap
-    /// @dev The last 8 bits of the nonce value is the position of the bit in the bitmap
-    function bitmapPositions(uint256 nonce) private pure returns (uint256 wordPos, uint256 bitPos) {
-        wordPos = uint248(nonce >> 8);
-        bitPos = uint8(nonce);
+    function _flipBit(mapping(uint256 => uint256) storage bitmap, uint256 nonce) private {
```
snreynolds, line 20: imo this does more than flip the bit so we should maybe split out this code.

1 - use the nonce to extract the wordPos/the bitmap to operate on 
2- flip bit

flipBit(nonce) makes it seem like the nonce thats being passed in is the bit thats being flipped which is not really how the data is used

src/base/UnorderedNonce.sol
```diff
@@ -12,21 +12,42 @@ contract UnorderedNonce {
...
+        // equivalent to:
```
snreynolds, line 21: can we fuzz test equivalence? this is a lot of assembly :P

src/base/UnorderedNonce.sol
snreynolds: are we doing that? lol

snreynolds: we can put that in the error pr just a call out tho

test/mocks/MockUnorderedNonce.sol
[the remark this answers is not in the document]
saucepoint: oh lol makes sense. got confused since spendNonce can technically "batch" spend multiple bits

src/base/UnorderedNonce.sol
```diff
@@ -1,53 +1,20 @@
...
-            if eq(and(previousBits, bit), bit) {
-                mstore(0, 0x1fb09b80) // 4 bytes of NonceAlreadyUsed.selector
-                revert(0x1c, 0x04)
-            }
-
-            // bitmap[wordPos] = (previousBits | bit)
-            sstore(slot, or(previousBits, bit))
-        }
+        (uint256 wordPos, uint256 bitPos) = nonce.getBitmapPositions();
+        nonces[owner].flipBit(wordPos, bitPos);
```
snreynolds, line 18: is it more expensive if we did nonces[owner][wordPos].flipBit(bitPos)? can you see? i think i just like that abstraction a little better

saucepoint: its not as friendly since `flipBit` would calculate the new bitmap and return it so the caller can set it in storage

the alternative/cleaner syntax might be `nonces[owner].spendNonce(nonce)` where `spendNonce` does the wordPos/bitPos logic

snreynolds: hmmm

hensha256: > nonces[owner][wordPos].flipBit(bitPos)

yeah generally nested mappings are much more expensive becaus eyou do more hashing (and hashing is on the verge of getting even more expensive, theres a change coming)

however i like `nonces[owner].spendNonce(nonce)` this!!

hensha256: ok i misunderstood - now i see it lol
i agree that doing it on the word would be messy becaus eyou have to write
```
nonces[owner][wordPos] = nonces[owner][wordPos].flipBit(x)
```

snreynolds: i think i still prefer the abstraction where we were to change it to 

nonces[owner][wordPos].flipTick bc the functions become more reusable (to extract bitPos and wordPos) from nonce, then to get the new bitmap...

buttt i get it if its too expensive

maybe we could still have nonces[owner].spendNonce() that still use other helpers for those two things? im just noticing that we have to rewrite helpers in tests to do the same thing weve written here/ I could see those functions being helpful generally in a library

snreynolds: Can you see how much gas is saved not doing it this way? @saucepoint 

like is it significantly more expsenive to do

nonces[owner][wordPos] = nonces[owner][wordPos].flipTick(bitPos) ?

snreynolds: i will also say tho.. this is not a super heavily used codepath so we could just go back to the solidity version and be done w it, theres no need to hyper optimize this path

src/base/ERC721Permit.sol
saucepoint: note for self: leaving DOMAIN_SEPARATOR and will move to EIP712.sol if/when its written

src/base/UnorderedNonce.sol
hensha256: yeah im happy with that 🤷 no strong opinions

src/base/UnorderedNonce.sol
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.24;
+
+/// @title Unordered Nonce
+/// @notice Contract state and methods for using unordered nonces in signatures
+contract UnorderedNonce {
+    error NonceAlreadyUsed();
+
+    mapping(address owner => mapping(uint256 word => uint256 bitmap)) public nonces;
```
snreynolds, line 9: natspec that the word is actually max type(uint248).max

test/position-managers/Permit.t.sol
```diff
@@ -0,0 +1,277 @@
...
+                    keccak256("Uniswap V4 Positions NFT"), // storage is private on EIP712.sol so we need to hardcode these
+                    keccak256("1"),
+                    block.chainid,
+                    address(lpm)
+                )
+            )
+        );
+    }
+
+    function test_permit_increaseLiquidity() public {
```
snreynolds, line 76: you dont actually need permissions to increase, so can probably remove this?

snreynolds: or you can leave it to show you dont need a permit

test/position-managers/Permit.t.sol
```diff
@@ -0,0 +1,277 @@
...
+        assertGt(balance0BobBefore, currency0.balanceOf(bob));
+        assertGt(balance1BobBefore, currency1.balanceOf(bob));
+    }
+
+    function test_permit_decreaseLiquidity() public {
+        uint256 liquidityAlice = 1e18;
+        vm.prank(alice);
+        mint(config, liquidityAlice, alice, ZERO_BYTES);
+        uint256 tokenIdAlice = lpm.nextTokenId() - 1;
+
```
snreynolds, line 109: can you add an expect revert bob trying to adjust alices position before the permit

saucepoint: tested in `test_noPermit_decreaseLiquidityRevert`

test/position-managers/Permit.t.sol
saucepoint: removing

the test `test_increaseLiquidity_withUnapprovedCaller` shows that users can increase without approval

snreynolds (approved):

PR #218, opened 2024-07-27
Add transfer support

snreynolds:
Decide what the final "resolve" actions are & share the FULL_DELTA logic with router
opening into actions router since this builds on that


src/PositionManager.sol
```diff
@@ -217,9 +237,35 @@ contract PositionManager is
         if (nativeBalance > 0) recipient.safeTransferETH(nativeBalance);
     }
 
+    /// @param params an encoding of Currency, address
+    function _sweepERC20To(bytes calldata params) internal {
```
hensha256, line 241: i think using CurrencyLib this should be easy to make into a `_sweepCurrencyTo` and then you can just do `CurrencyLib.balanceOfSelf` and CurrencyLib.transfer`

hensha256: (and then you can remove that eth sweep from the settle function...)

src/PositionManager.sol
```diff
@@ -217,9 +237,35 @@ contract PositionManager is
...
+        if (tokenBalance > 0) currency.transfer(to, tokenBalance);
+    }
+
     // implementation of abstract function DeltaResolver._pay
-    function _pay(Currency token, address payer, uint256 amount) internal override {
-        // TODO this should use Permit2
-        ERC20(Currency.unwrap(token)).safeTransferFrom(payer, address(poolManager), amount);
+    function _pay(Currency currency, address payer, uint256 amount) internal override {
+        if (payer == address(this)) {
+            // TODO: This transfer no eth check. This is guaranteed to not be eth.
```
hensha256, line 250: > This is guaranteed to not be eth.

say why! also i dont understand what the todo is?

src/interfaces/IPositionManager.sol
```diff
@@ -6,6 +6,9 @@ import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
 interface IPositionManager {
     error NotApproved(address caller);
     error DeadlinePassed();
+    error NothingToSettle();
```
hensha256, line 9: unused?

test/position-managers/IncreaseLiquidity.t.sol
```diff
@@ -335,4 +335,41 @@ contract IncreaseLiquidityTest is Test, PosmTestSetup, Fuzzers {
...
+        vm.prank(alice);
+        bytes[] memory results = lpm.modifyLiquidities(calls, _deadline);
+        uint256 amount0 = abi.decode(results[1], (uint256));
+        uint256 amount1 = abi.decode(results[2], (uint256));
+
+        // The balances were swept back to this address.
+        assertEq(IERC20(Currency.unwrap(currency0)).balanceOf(address(lpm)), 0);
+        assertEq(IERC20(Currency.unwrap(currency1)).balanceOf(address(lpm)), 0);
+
+        assertEq(currency0.balanceOf(address(this)), balanceBefore0 - amount0);
```
hensha256, line 372: e.g. here without return values i feel like you can do
```
// you do this already above
assert(balance(lpm), 0)
// new check without return values available
assertEq(changeInPMBalance + changeInThisBalance, 100e18)
```

src/PositionManager.sol
snreynolds: Because by the time we call _pay from inside the DeltaResolver settle function we have already checked if its eth. eth follows a diff path in settle

test/position-managers/IncreaseLiquidity.t.sol
snreynolds: I can get return values now in my other pr so will leave it & fix during merge confs

src/PositionManager.sol
```diff
@@ -158,9 +158,6 @@ contract PositionManager is
         address caller = _msgSender();
         if (currencyDelta < 0) {
             _settle(currency, caller, uint256(-currencyDelta));
-
-            // if there are native tokens left over after settling, return to locker
-            if (currency.isNative()) _sweepNativeToken(caller);
```
hensha256, line 152: yes i feel good about this

src/PositionManager.sol
```diff
@@ -50,6 +50,8 @@ contract PositionManager is
 
     IAllowanceTransfer public immutable permit2;
 
+    uint256 public constant FULL_DELTA = type(uint256).max;
```
hensha256, line 53: I do think youll find if you make this `0` gas should go down noticeably because of removing non-0 calldata bytes.
It _should_ go down by `(16-4) * 32) = 384`

hensha256 (commented):
just 1 thought but generally looks great

src/PositionManager.sol
snreynolds: lemme check!

snreynolds: went down by ~700

snreynolds: https://github.com/Uniswap/v4-periphery/pull/218/commits/b0b1b7321e55e689f0b24262f746db6c02e424c1

src/PositionManager.sol
```diff
@@ -202,16 +212,28 @@ contract PositionManager is
...
+    function _pay(Currency currency, address payer, uint256 amount) internal override {
+        if (payer == address(this)) {
+            // TODO: currency is guaranteed to not be eth so the native check in transfer is not optimal.
+            currency.transfer(address(poolManager), amount);
+        } else {
+            permit2.transferFrom(payer, address(poolManager), uint160(amount), Currency.unwrap(currency));
+        }
+    }
+
+    function _getFullSettleAmount(Currency currency) private view returns (uint256 amount) {
```
hensha256, line 233: how do you feel about putting this in DeltaResolver? 👀 i also have this exact code lol

snreynolds: sure!

test/position-managers/PositionManager.gas.t.sol
```diff
@@ -376,7 +376,13 @@ contract PosMGasTest is Test, PosmTestSetup, GasSnapshot {
 
     function test_gas_mint_native_excess() public {
         uint256 liquidityToAdd = 10_000 ether;
-        bytes memory calls = getMintEncoded(configNative, liquidityToAdd, address(this), ZERO_BYTES);
```
hensha256, line 379: if you wanted to keep using the liquidity operations helpers you could always have it check if one of the currencies is native and add a sweep automatically? like in `finalizeML` if its a mint/increase

hensha256: but dont feel strongly about it if you dnt want to

snreynolds: eh its fine for now, can maybe add it later

src/V4Router.sol
```diff
@@ -66,12 +65,11 @@ abstract contract V4Router is IV4Router, BaseActionsRouter, DeltaResolver {
                     recipient := calldataload(add(params.offset, 0x20))
                 }
 
-                int256 delta = poolManager.currencyDelta(address(this), currency);
```
hensha256, line 69: i think you can now remove the transient state library from this file

hensha256: and probably from posm too?

snreynolds: removed from router, we need in posm for close

test/position-managers/IncreaseLiquidity.t.sol
```diff
@@ -338,4 +338,42 @@ contract IncreaseLiquidityTest is Test, PosmTestSetup, Fuzzers {
             );
         }
     }
+
+    function test_mint_settleWithBalance() public {
```
dianakocsis, line 342: can you also test `test_increaseLiquidity_settleWithBalance()`

src/libraries/Actions.sol
```diff
@@ -32,4 +32,8 @@ library Actions {
     // mint + burn ERC721 position
     uint256 constant MINT_POSITION = 0x22;
```
dianakocsis, line 33: thoughts on moving these to the pools action section? since they also modify liquidity

hensha256 (approved):

PR #220, opened 2024-07-27
test SafeCallback

### LINKED ISSUE #183: test SafeCallback
### Component

General design optimization (improving efficiency, cleanliness, or developer experience)

A lot of our peripheral contracts use SafeCallback so we should add tests. specifically noticed its missing a revert test

_No response_

### Describe alternatives.

_No response_

### Additional context.

_No response_

saucepoint:
## Related Issue
Closes #183 

## Description of changes

Added a mock contract and tested that `SafeCallback.unlockCallback()` is only accessible from PoolManager


test/SafeCallback.t.sol
```diff
@@ -0,0 +1,29 @@
...
+    function test_unlock(uint256 num) public {
+        bytes memory result = safeCallback.unlockManager(num);
+        assertEq(num, abi.decode(result, (uint256)));
+    }
+
+    function test_unlockRevert(address caller, bytes calldata data) public {
+        vm.startPrank(caller);
+        if (caller != address(manager)) vm.expectRevert(SafeCallback.NotPoolManager.selector);
+        safeCallback.unlockCallback(data);
+        vm.stopPrank();
```
saucepoint, line 27: this is pretty neat from foundry, when fuzzing for addresses the fuzzer includes deployed contracts from setup! 😲

test/SafeCallback.t.sol
```diff
@@ -0,0 +1,29 @@
...
+
+contract SafeCallbackTest is Test, Deployers {
+    MockSafeCallback safeCallback;
+
+    function setUp() public {
+        deployFreshManager();
+        safeCallback = new MockSafeCallback(manager);
+    }
+
+    function test_unlock(uint256 num) public {
```
snreynolds, line 18: can we also just add an assert that the pool manager in SafeCallback is the address we expect/the pool manager address we initialized it with

snreynolds (changes requested):

snreynolds (approved):

Issue #199, opened 2024-07-25
[posm, routing] gas optimise abi decode/encode


hensha256:
### Component

Position Manager

### Describe the suggested feature and problem it solves.

Decode in assembly

hensha256:
This hsould include a helper library that both posm and routing can use

PR #226, opened 2024-07-30
Decode posm params in calldata

hensha256:
https://github.com/Uniswap/v4-periphery/issues/199


src/PositionManager.sol
```diff
@@ -92,22 +94,22 @@ contract PositionManager is
...
     }
 
-    /// @param params is an encoding of uint256 tokenId, PositionConfig memory config, uint256 liquidity, bytes hookData
+    /// @param params is an encoding of uint256 tokenId, PositionConfig calldata config, uint256 liquidity, bytes hookData
     /// @dev Calling increase with 0 liquidity will credit the caller with any underlying fees of the position
-    function _increase(bytes memory params) internal {
-        (uint256 tokenId, PositionConfig memory config, uint256 liquidity, bytes memory hookData) =
-            abi.decode(params, (uint256, PositionConfig, uint256, bytes));
+    function _increase(bytes calldata params) internal {
+        (uint256 tokenId, PositionConfig calldata config, uint256 liquidity, bytes calldata hookData) =
```
snreynolds, line 100: why is PositionConfig still calldata? it gets translated to memory when we do toId() no? So I feel like this is doing more work than it needs to?

snreynolds (commented):
I took a quick look, generally I like. Although I would like @saucepoint to take a look bc he made another comment on a pr about decoding params before we actually do the internal call.

src/PositionManager.sol
hensha256: it does but it means its only brought to memory once instead of twice - i've optimised `toId` too though

src/libraries/CalldataDecoder.sol
```diff
@@ -40,4 +45,99 @@ library CalldataDecoder {
...
+        pure
+        returns (uint256 tokenId, PositionConfig calldata config, uint256 liquidity, bytes calldata hookData)
+    {
+        assembly ("memory-safe") {
+            tokenId := calldataload(params.offset)
+            config := add(params.offset, 0x20)
+            liquidity := calldataload(add(params.offset, 0x100))
+        }
+        hookData = params.toBytes(9);
+    }
```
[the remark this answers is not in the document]
saucepoint: (applies to the other functions too)

saucepoint (commented):
looks great as-is. we should probably decide on decoding before vs within the _internalFunctions

really like the tests too

saucepoint (approved):

PR #229, opened 2024-07-30
posm: add staking through subscribers

snreynolds:
TODO:
- [x] Test ETH, I think we may want to make these functions payable to enable multicall with them
- [x] Test malicious subscriber, token can still unsubscribe


src/base/Notifier.sol
```diff
@@ -0,0 +1,51 @@
...
+
+        ISubscriber(newSubscriber).notifySubscribe(tokenId, config);
+        emit Subscribed(tokenId, address(newSubscriber));
+    }
+
+    /// @dev Must always allow a user to unsubscribe. In the case of a malicious subscriber, a user can always unsubscribe safely, ensuring liquidity is always modifiable.
+    function _unsubscribe(uint256 tokenId, PositionConfig memory config) internal {
+        ISubscriber _subscriber = subscriber[tokenId];
+
+        if (gasleft() < subscriberGasLimit) revert SubscriberCannotBeNotified();
```
hensha256, line 37: i feel like this revert is unnecessary - maybe this subscriber only needs like 30% of the total subscriberGasLimit, but the user has to send WAY more just to get this line to pass. If it actually runs out of gas then that will cause the failure?

hensha256: i feel like this is true in core too... like what does that check gain

src/base/Notifier.sol
```diff
@@ -0,0 +1,51 @@
...
+        try _subscriber.notifyUnsubscribe{gas: subscriberGasLimit}(tokenId, config) {} catch {}
```
hensha256, line 38: is this subject to return data bombing? i feel like yes? we might have to do this call in assembly to avoid it 👀

src/libraries/PositionConfig.sol
```diff
@@ -3,17 +3,61 @@ pragma solidity ^0.8.24;
 
 import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
 
-// A PositionConfig is the input for creating and modifying a Position in core, set per tokenId
+// A PositionConfig is the input for creating and modifying a Position in core, whos truncated hash is set per tokenId
```
hensha256, line 6:
```suggestion
// A PositionConfig is the input for creating and modifying a Position in core, whose truncated hash is set per tokenId
```

src/libraries/PositionConfig.sol
```diff
@@ -24,7 +68,7 @@ library PositionConfigLibrary {
             mstore(add(fmp, 0x14), calldataload(add(config, 0x20))) // currency1: [0x20, 0x34)
             mstore(fmp, calldataload(config)) // currency0: [0x0c, 0x20)
 
-            id := keccak256(add(fmp, 0x0c), 0x48) // len is 72 bytes
+            id := shr(1, keccak256(add(fmp, 0x0c), 0x48)) // len is 72 bytes, truncate upper bit of the hash
```
hensha256, line 71: re the comment - technically it truncates the LOWER bit of the hash because it shifts right no?
if a hash was `10110011` and you shift right one it becomes `01011001` so youve lost the lower bit of the hash. And the _upper_ bit is guaranteed to be 0

src/libraries/PositionConfig.sol
```diff
@@ -3,17 +3,61 @@ pragma solidity ^0.8.24;
...
+// A PositionConfig is the input for creating and modifying a Position in core, whos truncated hash is set per tokenId
 struct PositionConfig {
     PoolKey poolKey;
     int24 tickLower;
     int24 tickUpper;
 }
 
-/// @notice Library for computing the configId given a PositionConfig
+/// @notice Library to get and set the PositionConfigId and subscriber status for a given tokenId
 library PositionConfigLibrary {
```
hensha256, line 14: needs a comment explaining the setup - that the msb is the 0/1 for subscriber status, and the other 255 bits are the most significant 255 bits of the hash

src/libraries/PositionConfig.sol
```diff
@@ -3,17 +3,61 @@ pragma solidity ^0.8.24;
...
+    }
+
+    function setSubscribe(mapping(uint256 => bytes32) storage positionConfigs, uint256 tokenId) internal {
+        bytes32 _config = positionConfigs[tokenId];
+        positionConfigs[tokenId] = _config | DIRTY_UPPER_BIT;
+    }
+
+    function setUnsubscribe(mapping(uint256 => bytes32) storage positionConfigs, uint256 tokenId) internal {
+        bytes32 _config = positionConfigs[tokenId];
+        positionConfigs[tokenId] = _config & MASK_UPPER_BIT;
```
hensha256, line 45: probably cheaper not to have an intermediate var?
`positionConfigs[tokenId] &= MASK_UPPER_BIT`
and above
`positionConfigs[tokenId] |= DIRTY_UPPER_BIT;`

hensha256 (changes requested):

src/PositionManager.sol
```diff
@@ -48,12 +50,13 @@ contract PositionManager is
...
     /// @inheritdoc IPositionManager
-    mapping(uint256 tokenId => bytes32 configId) public positionConfigs;
+    mapping(uint256 tokenId => bytes32 config) public positionConfigs;
 
     IAllowanceTransfer public immutable permit2;
 
-    constructor(IPoolManager _poolManager, IAllowanceTransfer _permit2)
+    constructor(IPoolManager _poolManager, IAllowanceTransfer _permit2, uint256 _subscriberGasLimit)
         BaseActionsRouter(_poolManager)
+        Notifier(_subscriberGasLimit)
```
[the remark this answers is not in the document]
snreynolds: yeah sure, happy to do that. But should it still be a constructor param (the percentage?)

src/PositionManager.sol
```diff
@@ -75,6 +88,25 @@ contract PositionManager is
         _executeActions(unlockData);
     }
 
+    function subscribe(uint256 tokenId, PositionConfig calldata config, address subscriber)
+        external
+        onlyIfApproved(msg.sender, tokenId)
+        onlyValidConfig(tokenId, config)
+    {
+        if (positionConfigs.getConfigId(tokenId) != config.toId()) revert IncorrectPositionConfigForTokenId(tokenId);
```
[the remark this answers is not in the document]
snreynolds: oh oops! thanks

src/PositionManager.sol
```diff
@@ -75,6 +88,25 @@ contract PositionManager is
...
+        _subscribe(tokenId, config, subscriber);
+        positionConfigs.setSubscribe(tokenId);
```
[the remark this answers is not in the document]
snreynolds: ah nice, good call out!

src/base/Notifier.sol
snreynolds: i think youre right, what I thought it was preventing was a user sending too little gas to evade the notification to subscriber but I think it would still just revert later

src/base/Notifier.sol
snreynolds: i think no, because we call it in solidity and it expects to return data (spoke about this in the thread with daniel)

src/PositionManager.sol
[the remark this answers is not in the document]
hensha256: oh i think it would be to remove constructor params entirely - if we dont care about the constructor having params then keep it as is!

src/base/Notifier.sol
hensha256: yeah same... i feel like its wrong in core too

src/base/Notifier.sol
hensha256: ok, worth writing a test for one with a huge return data then to show that that isnt true

src/libraries/PositionConfig.sol
hensha256: Right, yes, the bit that is signalling is the "upper" bit, I agree.
But the bit that is truncated off the hash is the lowest bit of the hash.

So "truncate upper bit of the hash" is incorrect

hensha256: I dont mind "upper" i just think the comment is wrong

src/libraries/PositionConfig.sol
hensha256: weeeeeird its in [the docs](https://docs.soliditylang.org/en/latest/types.html#compound-and-increment-decrement-operators)

ok so either way i'd say get rid of `_config` from memory and just do 
```
positionConfigs[tokenId] = positionConfigs[tokenId] & MASK_UPPER_BIT;
```

src/base/Notifier.sol
snreynolds: Let me open an issue in core as well then?

snreynolds: https://github.com/Uniswap/v4-core/issues/814

src/libraries/PositionConfig.sol
snreynolds: ohh I see what you mean! yes that makes sense

src/libraries/PositionConfig.sol
snreynolds: wait I'm sorry - it does! I must have been doing something differently yesterday 💤  i was tired.. it was 1030 lol

src/PositionManager.sol
```diff
@@ -61,11 +64,31 @@ contract PositionManager is
...
         if (block.timestamp > deadline) revert DeadlinePassed();
         _;
     }
 
+    /// @notice Reverts if the caller is not the owner or approved for the ERC721 token
+    /// @param sender The address of the caller
+    /// @param tokenId the unique identifier of the ERC721 token
+    /// @dev either msg.sender or _msgSender() is passed in as the sender
+    /// _msgSender() should ONLY be used if this is being called from within the unlockCallback
+    modifier onlyIfApproved(address sender, uint256 tokenId) {
```
hensha256, line 79: nit `sender` -> `caller` ?

src/base/Notifier.sol
```diff
@@ -0,0 +1,61 @@
...
+    ISubscriber private constant NO_SUBSCRIBER = ISubscriber(address(0));
+
+    // a percentage of the block.gaslimit denoted in BPS, used as the gas limit for subscriber calls
+    // 100 bps is 1%
+    // at 30M gas, the limit is 300K
+    uint256 private constant BLOCK_LIMIT_BPS = 100;
+
+    mapping(uint256 tokenId => ISubscriber subscriber) public subscriber;
+
+    constructor() {}
```
hensha256, line 28: remove?

src/PositionManager.sol
```diff
@@ -77,6 +100,28 @@ contract PositionManager is
...
     }
 
+    /// @inheritdoc INotifier
+    function subscribe(uint256 tokenId, PositionConfig calldata config, address subscriber)
+        external
+        payable
+        onlyIfApproved(msg.sender, tokenId)
+        onlyValidConfig(tokenId, config)
+    {
+        positionConfigs.setSubscribe(tokenId);
```
hensha256, line 110: i think a comment here and in unsubscribe could be helpful, like
`// call to _subscribe will revert if the user already has a subscriber`
or something like that?

src/base/Notifier.sol
```diff
@@ -0,0 +1,61 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+pragma solidity ^0.8.24;
+
+import {ISubscriber} from "../interfaces/ISubscriber.sol";
+import {PositionConfig} from "../libraries/PositionConfig.sol";
+import {GasLimitCalculator} from "../libraries/GasLimitCalculator.sol";
+
+import "../interfaces/INotifier.sol";
+
+abstract contract Notifier is INotifier {
```
hensha256, line 10: natspec

src/base/Notifier.sol
```diff
@@ -0,0 +1,61 @@
...
+        emit Subscribed(tokenId, address(newSubscriber));
+    }
+
+    /// @dev Must always allow a user to unsubscribe. In the case of a malicious subscriber, a user can always unsubscribe safely, ensuring liquidity is always modifiable.
+    function _unsubscribe(uint256 tokenId, PositionConfig memory config) internal {
+        ISubscriber _subscriber = subscriber[tokenId];
+
+        uint256 subscriberGasLimit = BLOCK_LIMIT_BPS.toGasLimit();
+
+        bytes memory data = abi.encodeWithSelector(ISubscriber.notifyUnsubscribe.selector, tokenId, config);
```
hensha256, line 46: remove this line?

src/base/Notifier.sol
```diff
@@ -0,0 +1,61 @@
...
+        if (_subscriber != NO_SUBSCRIBER) revert AlreadySubscribed(address(_subscriber));
+        subscriber[tokenId] = ISubscriber(newSubscriber);
+
+        ISubscriber(newSubscriber).notifySubscribe(tokenId, config);
+        emit Subscribed(tokenId, address(newSubscriber));
+    }
+
+    /// @dev Must always allow a user to unsubscribe. In the case of a malicious subscriber, a user can always unsubscribe safely, ensuring liquidity is always modifiable.
+    function _unsubscribe(uint256 tokenId, PositionConfig memory config) internal {
+        ISubscriber _subscriber = subscriber[tokenId];
```
hensha256, line 42: should it revert if they arent already subscribed? I believe currently if you call `_unsubscribe` and youre not subscribed it would succeed? Because contract calls to EOAs (like address(0)) succeed. Maybe add a test to check

src/interfaces/INotifier.sol
```diff
@@ -0,0 +1,18 @@
...
+
+import {PositionConfig} from "../libraries/PositionConfig.sol";
+
+/// @notice This interface is used to opt in to sending updates to external contracts about position modifications or transfers
+interface INotifier {
+    /// @notice Enables the subscriber to receive notifications for a respective position
+    /// @param tokenId the ERC721 tokenId
+    /// @param config the corresponding PositionConfig for the tokenId
+    /// @param subscriber the address to notify
+    function subscribe(uint256 tokenId, PositionConfig calldata config, address subscriber) external payable;
```
hensha256, line 12: `@dev` saying why its payable?
same on `permit` lol, to save people having the same confusion we did

src/PositionManager.sol
```diff
@@ -277,4 +320,10 @@ contract PositionManager is
             permit2.transferFrom(payer, address(poolManager), uint160(amount), Currency.unwrap(currency));
         }
     }
+
+    /// @dev overrides solmate transferFrom in case a notification to subscribers is needed
+    function transferFrom(address from, address to, uint256 id) public override {
```
hensha256, line 325: is it right that ERC721 doesnt have a `transfer` function?! Just `transferFrom`?? How am I just learning this

src/PositionManager.sol
```diff
@@ -277,4 +320,10 @@ contract PositionManager is
...
+        super.transferFrom(from, to, id);
+        if (positionConfigs.hasSubscriber(id)) _notifyTransfer(id, from, to);
```
hensha256, line 327: should transferring auto-unsubscribe?

test/mocks/MockBadSubscribers.sol
```diff
@@ -0,0 +1,61 @@
...
+        _;
+    }
+
+    function notifySubscribe(uint256 tokenId, PositionConfig memory config) external onlyByPosm {
+        notifySubscribeCount++;
+    }
+
+    function notifyUnsubscribe(uint256 tokenId, PositionConfig memory config) external onlyByPosm {
+        notifyUnsubscribeCount++;
+        uint256 _memPtr = memPtr;
```
hensha256, line 38: isnt memPtr always 0? i dont see a way it can be set?

test/position-managers/NativeToken.t.sol
```diff
@@ -402,4 +407,35 @@ contract PositionManagerTest is Test, PosmTestSetup, LiquidityFuzzers {
...
+        calls[1] = abi.encodeWithSelector(lpm.subscribe.selector, tokenId, config, sub);
+
+        lpm.multicall{value: 10e18}(calls);
+
+        bytes32 positionId =
+            Position.calculatePositionKey(address(lpm), config.tickLower, config.tickUpper, bytes32(tokenId));
+        (uint256 liquidity,,) = manager.getPositionInfo(config.poolKey.toId(), positionId);
+
+        // assertEq(liquidity, 100e18);
+        // assertEq(sub.notifySubscribeCount(), 1);
```
hensha256, line 439: uncomment?

test/mocks/MockBadSubscribers.sol
hensha256: can you explain what this subscriber does thats bad?

src/PositionManager.sol
snreynolds: yeah

src/PositionManager.sol
snreynolds: Hm.. I actually dont think so

snreynolds: the new owner can always unsubscribe

src/PositionManager.sol
snreynolds: I added in INotifier

src/interfaces/INotifier.sol
snreynolds: put in INotifier and IERC721

test/mocks/MockBadSubscribers.sol
snreynolds: notify unsubscribe returns too much data

i call setReturnDataSize to a large size and so the subscriber's notifyUnsubscribeCount does not get set bc it runs OOG

src/PositionManager.sol
```diff
@@ -50,7 +53,7 @@ contract PositionManager is
     uint256 public nextTokenId = 1;
 
     /// @inheritdoc IPositionManager
-    mapping(uint256 tokenId => bytes32 configId) public positionConfigs;
+    mapping(uint256 tokenId => bytes32 config) public positionConfigs;
```
hensha256, line 56: should this be private? now that the id is a mixture of information? and have a getter or 2?

src/interfaces/IERC721Permit.sol
```diff
@@ -21,6 +21,7 @@ interface IERC721Permit {
     /// @param v Must produce valid secp256k1 signature from the holder along with `r` and `s`
     /// @param r Must produce valid secp256k1 signature from the holder along with `v` and `s`
     /// @param s Must produce valid secp256k1 signature from the holder along with `r` and `v`
+    /// @dev payable so it can be multicalled with NATIVE related acitons
```
hensha256, line 24:
```suggestion
    /// @dev payable so it can be multicalled with NATIVE related actions
```

src/libraries/PositionConfig.sol
```diff
@@ -3,17 +3,58 @@ pragma solidity ^0.8.24;
...
+    /// @notice returns the truncated hash of the PositionConfig for a given tokenId
+    function getConfigId(mapping(uint256 => bytes32) storage positionConfigs, uint256 tokenId)
+        internal
+        view
+        returns (bytes32 configId)
+    {
+        configId = positionConfigs[tokenId] & MASK_UPPER_BIT;
+    }
+
+    function setConfigId(
```
hensha256, line 29: should this be called in `mint`? or removed? it currently isnt called anywhere

test/position-managers/PositionManager.notifier.t.sol
```diff
@@ -0,0 +1,283 @@
...
+        lpm.approve(address(this), tokenId);
+        vm.stopPrank();
+
+        lpm.subscribe(tokenId, config, address(badSubscriber));
+
+        MockReturnDataSubscriber(badSubscriber).setReturnDataSize(0x600000);
+        lpm.unsubscribe(tokenId, config);
+
+        // the subscriber contract call failed bc it used too much gas
+        assertEq(MockReturnDataSubscriber(badSubscriber).notifyUnsubscribeCount(), 0);
```
hensha256, line 201: check that the tokenId no longer has a subscriber to show it was successful

test/position-managers/PositionManager.notifier.t.sol
```diff
@@ -0,0 +1,283 @@
...
+        // approve this contract to operate on alices liq
+        vm.startPrank(alice);
+        lpm.approve(address(this), tokenId);
+        vm.stopPrank();
+
+        lpm.subscribe(tokenId, config, address(sub));
+
+        lpm.unsubscribe(tokenId, config);
+
+        assertEq(sub.notifyUnsubscribeCount(), 1);
```
hensha256, line 183: `assertEq(address(lpm.subscriber(tokenId)), address(0));`

test/position-managers/PositionManager.notifier.t.sol
hensha256: `assertEq(address(lpm.subscriber(tokenId)), address(sub));`

hensha256: and check that the 1/0 in config is set correctly too

hensha256: throughout this file i'd love to see checks that the `subscriber[]` and the bit are setting and unsetting when they should

test/position-managers/NativeToken.t.sol
```diff
@@ -402,4 +407,35 @@ contract PositionManagerTest is Test, PosmTestSetup, LiquidityFuzzers {
         assertEq(currency0.balanceOfSelf() - balance0Before, uint128(delta.amount0()));
         assertEq(currency1.balanceOfSelf() - balance1Before, uint128(delta.amount1()));
     }
+
+    // this test fails unless subscribe is payable
+    function test_mutlicall_mint_subscribe_native() public {
```
hensha256, line 412:
```suggestion
    function test_multicall_mint_subscribe_native() public {
```

src/libraries/PositionConfig.sol
snreynolds: oh ya.. I think merge conflicts got the best of me here

src/base/Notifier.sol
```diff
@@ -0,0 +1,58 @@
...
+import {PositionConfig} from "../libraries/PositionConfig.sol";
+import {GasLimitCalculator} from "../libraries/GasLimitCalculator.sol";
+
+import "../interfaces/INotifier.sol";
+
+/// @notice Notifier is used to opt in to sending updates to external contracts about position modifications or transfers
+abstract contract Notifier is INotifier {
+    using GasLimitCalculator for uint256;
+
+    error SubscriberCannotBeNotified();
```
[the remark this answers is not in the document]
snreynolds: actually we dont even throw this anymore, removed

test/position-managers/PositionManager.notifier.t.sol
snreynolds: good call, added

src/PositionManager.sol
```diff
@@ -338,4 +337,12 @@ contract PositionManager is
         super.transferFrom(from, to, id);
         if (positionConfigs.hasSubscriber(id)) _notifyTransfer(id, from, to);
     }
+
+    function getPositionConfigId(uint256 tokenId) external view returns (bytes32) {
```
hensha256, line 341: `@inheritdoc` and both these 2 functions in interface with natspec

hensha256: oh theyre in with natspec - just inheritdoc!

hensha256 (approved):

PR #230, opened 2024-07-31
Decode router params calldata

hensha256:
#199


test/libraries/CalldataDecoder.t.sol
```diff
@@ -83,6 +87,56 @@ contract CalldataDecoderTest is Test {
         _assertEq(_config, config);
     }
 
+    function test_fuzz_decodeSwapExactInParams(IV4Router.ExactInputParams calldata _swapParams) public view {
```
snreynolds, line 90: Ok hm I just had this thought which I think relates to all of these tests... Basically we encode the fuzz, and decode it with the library. But we never decode it regularly... shouldnt we be testing it against abi.decode9) not just the raw passed in data? 

ie 

_encodedParams = abi.encode(_swapParams)
_params = abi.decode(_encodedParams, IV4Router.ExactInputParams)

and then all the asserts are against _params

but lmk if you dont agree

snreynolds (approved):
lgtm, 1 little comment but if you disagree then feel free to merge

test/libraries/CalldataDecoder.t.sol
hensha256: #251

PR #234, opened 2024-07-31
Optimise actions encoding

hensha256:
Currently all our actions are well within 1 byte, but are being encoded as an array of 256 bit numbers - very wasteful on calldata. Now theyre packed as 1 byte commands like in the UR


src/base/BaseActionsRouter.sol
```diff
@@ -27,13 +27,13 @@ abstract contract BaseActionsRouter is SafeCallback {
     /// @notice function that is called by the PoolManager through the SafeCallback.unlockCallback
     function _unlockCallback(bytes calldata data) internal override returns (bytes memory) {
         // abi.decode(data, (uint256[], bytes[]));
```
snreynolds, line 29: update the comment

also can you add a @param tag explaining the encoding for actions

src/base/BaseActionsRouter.sol
```diff
@@ -27,13 +27,13 @@ abstract contract BaseActionsRouter is SafeCallback {
...
-        (uint256[] calldata actions, bytes[] calldata params) = data.decodeActionsRouterParams();
+        (bytes calldata actions, bytes[] calldata params) = data.decodeActionsRouterParams();
 
         uint256 numActions = actions.length;
         if (numActions != params.length) revert InputLengthMismatch();
 
         for (uint256 actionIndex = 0; actionIndex < numActions; actionIndex++) {
-            uint256 action = actions[actionIndex];
+            uint256 action = uint256(uint8(actions[actionIndex]));
```
snreynolds, line 36: would this line be incorrect to do `uint256(actions[actionIndex])` ?

snreynolds (commented):
this is super cool, didn't even think to do this/didn't realize it would save so much gas

src/base/BaseActionsRouter.sol
hensha256: wont compile because `actions[actionIndex]` is a `bytes1` so you have to cast it via `uint8`

hensha256: you always have to cast to a different type of the _same size_ before you change sizes. 
e.g. if you have int128 and you want uint256 you have to do `uint256(uint128(int128))`
so in this case you have ot go to `uint8` because its the same size as `bytes1`

snreynolds (approved):

PR #235, opened 2024-07-31
Router settle with balance

hensha256:
Allow settle to be funds from the router


test/shared/implementation/V4RouterImplementation.sol
```diff
@@ -27,7 +27,11 @@ contract V4RouterImplementation is V4Router, ReentrancyLock {
     }
 
     function _pay(Currency token, address payer, uint256 amount) internal override {
-        ERC20(Currency.unwrap(token)).safeTransferFrom(payer, address(poolManager), amount);
+        if (payer == address(this)) {
+            ERC20(Currency.unwrap(token)).safeTransfer(address(poolManager), amount);
```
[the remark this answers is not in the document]
hensha256: 🤷‍♀️ its a test so didnt think about it lol, will do it now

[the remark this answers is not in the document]
hensha256: done! i also moved this contract to `/mocks` and deleted `/implementations`

test/mocks/MockV4Router.sol
```diff
@@ -27,7 +28,11 @@ contract V4RouterImplementation is V4Router, ReentrancyLock {
     }
 
     function _pay(Currency token, address payer, uint256 amount) internal override {
-        ERC20(Currency.unwrap(token)).safeTransferFrom(payer, address(poolManager), amount);
+        if (payer == address(this)) {
+            token.transfer(address(poolManager), amount);
```
snreynolds, line 32: so I think we both have the same problem in pay where by this point (if its been called from DeltaResolver) it definitely wont be NATIVE currency, but we still check if its native inside currency. I wonder if we should write a currency.transferERC20() without that native check.. anyways food for thought.

src/V4Router.sol
```diff
@@ -44,11 +44,11 @@ abstract contract V4Router is IV4Router, BaseActionsRouter, DeltaResolver {
...
             if (action == Actions.SETTLE_ALL) {
                 Currency currency = params.decodeCurrency();
-                uint256 amount = _getFullSettleAmount(currency);
-
-                // TODO support address(this) paying too
                 // TODO should it have a maxAmountOut added slippage protection?
-                _settle(currency, _msgSender(), amount);
+                _settle(currency, _msgSender(), _getFullSettleAmount(currency));
+            } else if (action == Actions.SETTLE_WITH_BALANCE) {
+                Currency currency = params.decodeCurrency();
```
snreynolds, line 50: i like how you do this here, Im going to change posm to do this as well

snreynolds (approved):

test/mocks/MockV4Router.sol
hensha256: this contract is actually just a mock for tests - in the universal router (where i implement `_pay`) it doesnt check if its ETH. But its a good thought for posm

snreynolds (approved):

PR #236, opened 2024-08-01
multicall: bubble up revert reason

### LINKED ISSUE #175: [posm]: Bubble up reverts
### Component

Position Manager

Like v4-core we should bubble up reverts. Needed when we use multicall in posm

_No response_

### Describe alternatives.

_No response_

### Additional context.

_No response_

saucepoint:
## Related Issue
Closes #175 

## Description of changes

Modify Multicall.sol to bubble up custom reverts with arguments


src/base/Multicall.sol
```diff
@@ -20,7 +26,7 @@ abstract contract Multicall is IMulticall {
                     }
                 }
                 // Next 5 lines from https://ethereum.stackexchange.com/a/83577
-                if (result.length < 68) revert();
+                if (result.length < 68) CallFailed.selector.bubbleUpAndRevertWith();
```
snreynolds: not sure about this 68 length check.. but why wouldnt we want to bubble up if > 68 ?

src/base/Multicall.sol
```diff
@@ -1,30 +1,28 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity ^0.8.19;
 
+import {CustomRevert} from "@uniswap/v4-core/src/libraries/CustomRevert.sol";
+
 import {IMulticall} from "../interfaces/IMulticall.sol";
 
+import "forge-std/console2.sol";
```
snreynolds, line 8: remove?

snreynolds (commented):
i think once you merge the other bubble reverts into this PR would be nice to test with those Wrapped reverts to make sure we're surfacing those properly in the multicall

test/Multicall.t.sol
```diff
@@ -106,4 +107,112 @@ contract MulticallTest is Test {
...
+        calls[0] = abi.encodeWithSelector(MockMulticall(multicall).revertWithBytes.selector, data);
+
+        // revert is caught
+        vm.expectRevert(abi.encodeWithSelector(MockMulticall.ErrorBytes.selector, data));
+        multicall.multicall(calls);
+
+        // confirm expected length of the revert
+        try multicall.revertWithBytes(data) {}
+        catch (bytes memory reason) {
+            // errors with 0 bytes are by default 64 bytes of data (length & pointer?) + 4 bytes of selector
```
snreynolds, line 183: errors with 0 bytes of data outside the selector? thats super interesting.. didnt know that i guess that explains the 68 number

snreynolds (approved):
noice looks good to me

src/base/Multicall.sol
hensha256: because in old-solidity errors couldnt be less than 68 bytes, so it meant there was no error to bubble

PR #237, opened 2024-08-01
add posm view lib

snreynolds:
trying to clean up base PositionManager.sol contract, moving internal func to a lib & cleaning up some tests


src/libraries/PosmLiquidityLibrary.sol
```diff
@@ -0,0 +1,26 @@
...
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {StateLibrary} from "@uniswap/v4-core/src/libraries/StateLibrary.sol";
+import {Position} from "@uniswap/v4-core/src/libraries/Position.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+import {PoolIdLibrary} from "@uniswap/v4-core/src/types/PoolId.sol";
+
+import {PositionConfig} from "./PositionConfig.sol";
+
+/// @notice A library for reading posm liquidity balances from core, wraps core's StateLibrary
+library PosmLiquidityLibrary {
```
hensha256, line 13: I do wonder if we should be keeping the name `posm` out of the final code ... i feel like no one else knows what it means?

test/shared/PosmTestSetup.sol
```diff
@@ -75,4 +82,13 @@ contract PosmTestSetup is Test, Deployers, DeployPermit2, LiquidityOperations {
             delta = delta + hook.deltas(i);
         }
     }
+
+    function _getPositionLiquidity(uint256 tokenId, PositionConfig memory _config)
```
hensha256, line 86: this PR introduces a library that does exactly this, feels wrong that our tests that integrate then cant access it and youve had to copy the same code into the tests?
maybe its worth exposing a public function on position manager called `getPositionLiquidity(tokenId, config)` that returns the liquidity? and then tests/integrations can use that?
thoughts?

src/libraries/PosmLiquidityLibrary.sol
hensha256: oh also i dont think it needs to be `positionmanager` specific at all? really any contract that adds liquidity could use this library right?
i feel like it should be called `LiquidityLibrary` and `bytes32 salt` should be passed in instead of `tokenId`?
And then its just a generic library for liquidity providing contracts to use for their liquidity look ups?

src/libraries/PosmLiquidityLibrary.sol
```diff
@@ -0,0 +1,26 @@
...
+
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {StateLibrary} from "@uniswap/v4-core/src/libraries/StateLibrary.sol";
+import {Position} from "@uniswap/v4-core/src/libraries/Position.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+import {PoolIdLibrary} from "@uniswap/v4-core/src/types/PoolId.sol";
+
+import {PositionConfig} from "./PositionConfig.sol";
+
+/// @notice A library for reading posm liquidity balances from core, wraps core's StateLibrary
```
hensha256, line 12:
```suggestion
/// @notice A library for reading liquidity balances owned by the current contract from the PoolManager
```

test/shared/PosmTestSetup.sol
hensha256: mmm im not sure i like that given its to be used internally? like we dont use the StateLibrary internally in PM? but i dont feel strongly

saucepoint:
replaced by #270

PR #238, opened 2024-08-01
allow batching sig based approvals through p2 forwarder

snreynolds:
## Related Issue

## Description of changes


src/base/Permit2Forwarder.sol
```diff
@@ -0,0 +1,29 @@
...
+/// TODO: Does it need to be enforced?
+contract Permit2Forwarder {
+    IAllowanceTransfer public immutable permit2;
+
+    constructor(IAllowanceTransfer _permit2) {
+        permit2 = _permit2;
+    }
+
+    /// @notice allows forwarding a single permit to permit2
+    function selfPermit(address owner, IAllowanceTransfer.PermitSingle calldata permitSingle, bytes calldata signature)
```
hensha256, line 17: This is just to allow it to be added into multicall right? but in reality they could be permitting anyone, it doesnt have to be `self`?

If so I'd propose just calling it `permit`? `Permit2Forwarder.permit` makes sense to me

or `permitSingle` and `permitBatch`?

hensha256: oh just seen your `@dev` above!
No i dont think it needs to be enforced, I think just rename to remove the word `self` and keep the comment that thats the intended use case

src/base/Permit2Forwarder.sol
```diff
@@ -0,0 +1,32 @@
...
+/// @dev This contract does not enforce the spender to be this contract, but that is the intended use case
+contract Permit2Forwarder {
+    IAllowanceTransfer public immutable permit2;
+
+    constructor(IAllowanceTransfer _permit2) {
+        permit2 = _permit2;
+    }
+
+    /// @notice allows forwarding a single permit to permit2
+    /// @dev this function is payable to allow multicall with NATIVE bsed actions
```
hensha256, line 16:
```suggestion
    /// @dev this function is payable to allow multicall with NATIVE based actions
```

src/base/Permit2Forwarder.sol
```diff
@@ -0,0 +1,32 @@
...
+    function permit(address owner, IAllowanceTransfer.PermitSingle calldata permitSingle, bytes calldata signature)
+        external
+        payable
+    {
+        permit2.permit(owner, permitSingle, signature);
+    }
+
+    /// @notice allows forwarding batch permits to permit2
+    /// @dev this function is payable to allow multicall with NATIVE bsed actions
```
hensha256, line 25:
```suggestion
    /// @dev this function is payable to allow multicall with NATIVE based actions
```

test/position-managers/PositionManager.multicall.t.sol
```diff
@@ -71,4 +102,111 @@ contract PositionManagerMulticallTest is Test, PosmTestSetup, LiquidityFuzzers {
...
+        lpm.modifyLiquidities(mintCall, _deadline);
+
+        // 3. encode a permit for that revoked token
+        IAllowanceTransfer.PermitSingle memory permit =
+            defaultERC20PermitAllowance(Currency.unwrap(currency0), permitAmount, permitExpiration, permitNonce);
+        permit.spender = address(lpm);
+        bytes memory sig = getPermitSignature(permit, bobPrivateKey, PERMIT2_DOMAIN_SEPARATOR);
+
+        bytes[] memory calls = new bytes[](2);
+        calls[0] = abi.encodeWithSelector(Permit2Forwarder(lpm).permit.selector, bob, permit, sig);
```
hensha256, line 137: why isnt this just `Permit2Forwarder.permit.selector` ? why does it need a parameter?

snreynolds: Oh i can do that!

snreynolds: i really wanted it to be possible to always do IPositionManager._.selector  but we cant do that bc not all external functions are defined on IPositionManager

so then I tried lpm.selector but that also doesnt work.. its super weird


if u know how we can fix this so we dont always have to use the specific contract that would be the best but fixed for now like this

PR #243, opened 2024-08-01
posm: Clear

saucepoint:
## Related Issue
Adds support for `poolManager.clear()` to posm

## Description of changes

Expose support for `Actions.CLEAR`, which requires a maximum allowable clear limit.

i.e. For cases where the user will not know the amount (autocompounding), a user may be _willing to forfiet up to 100 wei of dust_


src/PositionManager.sol
```diff
@@ -161,6 +164,14 @@ contract PositionManager is
         }
     }
 
+    /// @dev integrators may elect to forfeit positive deltas with clear
+    /// provide a safety check that amount-to-clear is less than a user-provided maximum
+    function _clear(Currency currency, uint256 amountMax) internal {
+        int256 currencyDelta = poolManager.currencyDelta(address(this), currency);
+        if (amountMax < uint256(currencyDelta)) revert ClearExceedsMaxAmount(currency, currencyDelta, amountMax);
```
snreynolds, line 171: this is preference but it reads easier to me if we flip the vars

if (currencyDelta > amountMax) revert..

test/libraries/CalldataDecoder.t.sol
```diff
@@ -72,6 +72,14 @@ contract CalldataDecoderTest is Test {
         assertEq(Currency.unwrap(currency), Currency.unwrap(_currency));
     }
 
+    function test_fuzz_decodeCurrencyAndAddress(Currency _currency, uint256 _amount) public view {
```
snreynolds, line 75:
```suggestion
    function test_fuzz_decodeCurrencyAndUint256(Currency _currency, uint256 _amount) public view {
```

test/position-managers/IncreaseLiquidity.t.sol
```diff
@@ -458,4 +464,40 @@ contract IncreaseLiquidityTest is Test, PosmTestSetup, Fuzzers {
         assertEq(currency0.balanceOf(address(this)), balanceBefore0 - amount0);
         assertEq(currency1.balanceOf(address(this)), balanceBefore1 - amount1);
     }
+
+    function test_increaseLiquidity_clearRevert() public {
```
snreynolds, line 468: can you add one test where someone tries to clear a negative balance, it should revert in core

test/position-managers/PositionManager.gas.t.sol
```diff
@@ -170,13 +170,60 @@ contract PosMGasTest is Test, PosmTestSetup, GasSnapshot {
...
+        // Alice elects to forfeit unclaimed tokens
+        Plan memory planner = Planner.init();
+        planner.add(Actions.INCREASE_LIQUIDITY, abi.encode(tokenIdAlice, config, liquidityDelta, ZERO_BYTES));
+        planner.add(Actions.CLEAR, abi.encode(config.poolKey.currency0, halfTokensOwedAlice + 1 wei));
+        planner.add(Actions.CLEAR, abi.encode(config.poolKey.currency1, halfTokensOwedAlice + 1 wei));
+        bytes memory calls = planner.encode();
+
+        vm.prank(alice);
+        lpm.modifyLiquidities(calls, _deadline);
+        snapLastCall("PositionManager_increase_autocompound_clearExcess");
```
snreynolds, line 224: can you keep the same gas test but with take, not clear

test/position-managers/IncreaseLiquidity.t.sol
```diff
@@ -164,14 +165,19 @@ contract IncreaseLiquidityTest is Test, PosmTestSetup, Fuzzers {
...
-        vm.startPrank(alice);
-        increaseLiquidity(tokenIdAlice, config, liquidityDelta, ZERO_BYTES);
-        vm.stopPrank();
+        Plan memory planner = Planner.init();
+        planner.add(Actions.INCREASE_LIQUIDITY, abi.encode(tokenIdAlice, config, liquidityDelta, ZERO_BYTES));
+        planner.add(Actions.CLEAR, abi.encode(config.poolKey.currency0, 1 wei)); // alice is willing to forfeit 1 wei
+        planner.add(Actions.CLEAR, abi.encode(config.poolKey.currency1, 1 wei));
+        bytes memory calls = planner.encode();
 
-        // alice barely spent any tokens
```
snreynolds, line 171: also can we leave this test with take, and just add the clear one as well

so we have:

test_increaseLiquidity_withExactFees_take
test_increaseLiquidity_withExactFees_clear

test/position-managers/PositionManager.gas.t.sol
saucepoint: we have one, where its using close (take)

`test_gas_autocompound_excessFeesCredit / PositionManager_increase_autocompoundExcessFeesCredit` (169,831)

snreynolds (approved):
lgtm but lets also get @hensha256 to review

src/libraries/CalldataDecoder.sol
```diff
@@ -129,6 +129,17 @@ library CalldataDecoder {
         }
     }
 
+    function decodeCurrencyAndUint256(bytes calldata params)
```
hensha256, line 132:
```
 /// @dev equivalent to: abi.decode ...
```

src/PositionManager.sol
```diff
@@ -204,6 +207,14 @@ contract PositionManager is
         }
     }
 
+    /// @dev integrators may elect to forfeit positive deltas with clear
+    /// provide a safety check that amount-to-clear is less than a user-provided maximum
```
hensha256, line 211:
```suggestion
    /// provides a safety check that amount-to-clear is less than a user-provided maximum
```

src/PositionManager.sol
```diff
@@ -204,6 +207,14 @@ contract PositionManager is
...
+    function _clear(Currency currency, uint256 amountMax) internal {
+        int256 currencyDelta = poolManager.currencyDelta(address(this), currency);
+        if (uint256(currencyDelta) > amountMax) revert ClearExceedsMaxAmount(currency, currencyDelta, amountMax);
```
hensha256, line 214: more a stylistic thing - but I wonder if the behaviour should be that

```
if (uint256(currencyDelta) > amountMax) _take(currency, amount)
```

i feel like the natural behaviour is "if its less than X its cheaper for me to leave it, but if its more than X I want it" ?

saucepoint: from offline: we'll introduce CLEAR_OR_TAKE as a command. we want to try to map commands ~1:1 with pool manager

hensha256 (approved):

PR #244, opened 2024-08-01
[POC] Demo: Standard Error Wrapping 

saucepoint:
## Related Issue
Exploring the ergonomics of wrapped error handling

## Description of changes


src/base/BaseActionsRouter.sol
```diff
@@ -16,12 +16,22 @@ abstract contract BaseActionsRouter is SafeCallback {
...
+    error Wrap__CurrencyNotSettled(address subcontext, bytes reason);
+
     constructor(IPoolManager _poolManager) SafeCallback(_poolManager) {}
 
     /// @notice internal function that triggers the execution of a set of actions on v4
     /// @dev inheriting contracts should call this function to trigger execution
     function _executeActions(bytes calldata unlockData) internal {
-        poolManager.unlock(unlockData);
+        try poolManager.unlock(unlockData) {}
+        catch (bytes memory error) {
```
gretzke, line 28: I would suggest renaming `error` because it's a reserved keyword

snreynolds: we usually use `reason`... and it will also match the input in the bubble up error definitions

snreynolds (commented):
this is just bubbling up from modify? are we not doing multicall as well>

src/base/BaseActionsRouter.sol
```diff
@@ -16,12 +16,22 @@ abstract contract BaseActionsRouter is SafeCallback {
...
+            if (bytes4(error) == IPoolManager.CurrencyNotSettled.selector) {
+                revert Wrap__CurrencyNotSettled(address(poolManager), error);
```
hensha256, line 30: why are we calling out this one in particular as different? even though its being bubbled too?

saucepoint:
upon further discussions, we agreed that a contract should only `Wrap__` and revert if the call was made to an arbitrary/non-static address

taking a look at posm, we are not making any calls to arbitrary addresses until subscriber is merged in. closing for now, but will reference in the future

PR #246, opened 2024-08-02
Recipient mapping!

### LINKED ISSUE #202: [routing] support router paying for swaps
### Component

Pool Interaction, Swaps

Currenly only _msgSender pays for swaps, in UR we will have multi-hop swaps that go through multiple protocols. For this v4 will need to be able to pay with funds already in the router

_No response_

### Describe alternatives.

_No response_

### Additional context.

_No response_

hensha256:
closes #202 
mapping recipients with constants for 2 reasons
- saves gas for the 2 most common use-cases because you pass in lots of 0-calldata which is cheaper
- makes SDKs much easier to write because the SDK does not need to know the address of the contract, nor the address of the user. Its agnostic to those facts because it can use these constants instead


src/base/BaseActionsRouter.sol
```diff
@@ -52,4 +53,17 @@ abstract contract BaseActionsRouter is SafeCallback {
     /// `msg.sender` shouldnt be used, as this will be the v4 pool manager contract that calls `unlockCallback`
     /// If using ReentrancyLock.sol, this function can return _getLocker()
     function _msgSender() internal view virtual returns (address);
+
+    /// @notice Calculates the address for a action
+    /// @param recipient The address or address-flag for the action
+    /// @return output The resultant address for the action
+    function map(address recipient) internal view returns (address) {
```
hensha256, line 60: wouldve been nice in a library but sadly it needs `_msgSender` to work. Open to suggestions about where the function should live, and where the constants should live

.forge-snapshots/Payments_swap_settleWithBalance_takeAllToSpecifiedAddress.snap
```diff
@@ -0,0 +1 @@
+127508
```
hensha256, line 1: Having a recipient of a specific address (not a helper constant) is now ~90 more expensive than before. But taking to the `msg.sender` is now ~40 cheaper than before.
And most importantly - our sdks will be much easier to write haha

hensha256: previously taking to any address: 127419
now taking to specific address: 127508
now taking to msg.sender: 127370

gretzke (approved):

PR #242, opened 2024-08-01
Settle pair

dianakocsis:
## Related Issue
add settle_pair command

PR #247, opened 2024-08-02
Optimised currency deltas

hensha256:
Its very expensive to make arrays in memory so its actually cheaper to do 2 external calls to exttload than create the array


.forge-snapshots/CurrencyDeltas_10iterations.snap
```diff
@@ -0,0 +1 @@
+189116
```
hensha256, line 1: Down ~1000 gas (tested locally)

snreynolds:
closing since we remove the lib in #242

PR #249, opened 2024-08-02
Pay with balance, and use delta as amount in

hensha256:


test/mocks/MockV4Router.sol
```diff
@@ -38,4 +38,6 @@ contract MockV4Router is V4Router, ReentrancyLock {
     function _msgSender() internal view override returns (address) {
         return _getLocker();
     }
+
+    receive() external payable {}
```
snreynolds, line 42: whats the motivation for this?

src/V4Router.sol
```diff
@@ -48,7 +50,7 @@ abstract contract V4Router is IV4Router, BaseActionsRouter, DeltaResolver {
                 _settle(currency, _msgSender(), _getFullSettleAmount(currency));
             } else if (action == Actions.SETTLE_WITH_BALANCE) {
                 Currency currency = params.decodeCurrency();
-                _settle(currency, address(this), _getFullSettleAmount(currency));
+                _settle(currency, address(this), currency.balanceOfSelf());
```
snreynolds, line 53: this change is specifically for FOT support yes? we need this then in posm too..

snreynolds: just noting that the behavior for this command is changing here and i want to keep the same commands that are needed in posm and routing behaving the same.

Im also confused are we still adding a command settle that takes in an amount?? or take with an amount?? 

ie in final we will have?

settle(curreny,amount) -> amount can be the constant OPEN_DELTA which means query it
settle_all -> current behavior
settleWithBalance(currency) -> amount is derived from the contract balance and payer is address(this)

src/V4Router.sol
```diff
@@ -63,12 +65,11 @@ abstract contract V4Router is IV4Router, BaseActionsRouter, DeltaResolver {
     }
 
     function _swapExactInputSingle(IV4Router.ExactInputSingleParams memory params) private {
+        uint128 amountIn = (params.amountIn == OPEN_DELTA)
```
snreynolds, line 68: can this be moved to a library? this is for sure shared logic / will be shared logic

src/V4Router.sol
```diff
@@ -78,8 +79,9 @@ abstract contract V4Router is IV4Router, BaseActionsRouter, DeltaResolver {
             // Caching for gas savings
             uint256 pathLength = params.path.length;
             uint128 amountOut;
-            uint128 amountIn = params.amountIn;
             Currency currencyIn = params.currencyIn;
+            uint128 amountIn =
```
snreynolds, line 83: lib could be used here too
so we dont have to repeat amount == OPEN_DELTA ? __ : ___

src/V4Router.sol
hensha256: discussed on slack but this is not (solely) for FOT although it doesnt make a sub-optimal FOT option possible. This is to enable multi-hop multi-protocol swaps like
USDC-v2->DAI-v4->UNI
where the DAI gets left in the contract and then passed in to the second swap using SETTLE_WITH_BALANCE and OPEN_DELTA

separately i will add a settle that takes an amount for FOT to be optimal

test/mocks/MockV4Router.sol
hensha256: just so that the mock contract can receive ETH to test SETTLE_WITH_BALANCE with ETH as the currency. The mock didnt previously have a way to receive ETH to do so

src/V4Router.sol
```diff
@@ -50,9 +52,9 @@ abstract contract V4Router is IV4Router, BaseActionsRouter, DeltaResolver {
                 Currency currency = params.decodeCurrency();
                 // TODO should it have a maxAmountOut added slippage protection?
                 _settle(currency, _msgSender(), _getFullSettleAmount(currency));
-            } else if (action == Actions.SETTLE_WITH_BALANCE) {
-                Currency currency = params.decodeCurrency();
-                _settle(currency, address(this), _getFullSettleAmount(currency));
+            } else if (action == Actions.SETTLE) {
+                (Currency currency, uint256 amount, bool payerIsUser) = params.decodeCurrencyUint256AndBool();
+                _settle(currency, payerIsUser ? _msgSender() : address(this), _mapAmount(amount, currency));
```
snreynolds, line 57: mapRecipient?

snreynolds (commented):
i like, also just my comments about shared lib from before

src/base/BaseActionsRouter.sol
```diff
@@ -11,6 +11,10 @@ import {Actions} from "../libraries/Actions.sol";
 abstract contract BaseActionsRouter is SafeCallback {
     using CalldataDecoder for bytes;
 
+    /// @notice used to signal that the recipient of an action should be the _msgSender of address(this)
+    address internal constant MSG_SENDER = address(1);
```
snreynolds, line 15: can we just have a shared constants file with all of these?

msg sender, address this, contract balance, open delta?

test/BaseActionsRouter.t.sol
```diff
@@ -138,11 +138,11 @@ contract BaseActionsRouterTest is Test, Deployers, GasSnapshot {
...
 
-    function test_fuzz_map(address recipient) public view {
-        address mappedRecipient = router.map(recipient);
-        if (recipient == Actions.MSG_SENDER) {
+    function test_fuzz_mapRecipient(address recipient) public view {
+        address mappedRecipient = router.mapRecipient(recipient);
+        if (recipient == router.MSG_SENDER_()) {
             assertEq(mappedRecipient, address(0xdeadbeef));
-        } else if (recipient == Actions.ADDRESS_THIS) {
+        } else if (recipient == router.ADDRESS_THIS_()) {
```
snreynolds, line 145: i would prefer this is Constants.ADDRESS_THIS/ Constants.MSG_SENDER 

esp for msg sender its confusing bc you can actually query the true msg sender on the router/posm contract with router.msgSender()

src/base/DeltaResolver.sol
```diff
@@ -59,4 +62,24 @@ abstract contract DeltaResolver is ImmutableState {
...
+        if (amount == Constants.CONTRACT_BALANCE) {
+            return currency.balanceOfSelf();
+        } else if (amount == Constants.OPEN_DELTA) {
+            return _getFullSettleAmount(currency);
+        }
+        return amount;
+    }
+
+    /// @notice Calculates the amount for a swap action
+    function _mapSwapAmount(uint128 amount, Currency currency) internal view returns (uint128) {
```
snreynolds, line 77: nit: Since I believe we will use this function we can use a more neutral name like mapInputAmount

snreynolds (approved):

PR #256, opened 2024-08-03
Replace OZ EIP712

saucepoint:
## Related Issue
Replace OZ EIP712.sol with our own because UR uses OZ v4.7 and cannot find the import

## Description of changes

Took Permit2's EIP712.sol:
* turned `_HASHED_NAME` into an immutable
* ~~Added checks for _CACHED_THIS == address(this) (mirrors OZ's behavior)~~ removed since its a delegate protection, see comment below
* Added `_HASHED_VERSION` to DOMAIN_SEPARATOR (mirrors OZ's behavior)

---

RYO cryptography at 10pm on a friday, what could possibly go wrong 🤷


.forge-snapshots/PositionManager_permit.snap
```diff
@@ -1 +1 @@
-79585
\ No newline at end of file
+79660
```
saucepoint, line 1: image

src/base/EIP712.sol
```diff
@@ -0,0 +1,44 @@
...
+contract EIP712 is IEIP712 {
+    // Cache the domain separator as an immutable value, but also store the chain id that it
+    // corresponds to, in order to invalidate the cached domain separator if the chain id changes.
+    bytes32 private immutable _CACHED_DOMAIN_SEPARATOR;
+    uint256 private immutable _CACHED_CHAIN_ID;
+    bytes32 private immutable _HASHED_NAME;
+    bytes32 private immutable _HASHED_VERSION;
+
+    bytes32 private constant _TYPE_HASH =
+        keccak256("EIP712Domain(string name,string version,uint256 chainId,address verifyingContract)");
```
snreynolds, line 19: do we rlly need version? whats the gas with no version?

saucepoint:
upon additional reading from https://github.com/Uniswap/permit2/pull/51 and https://github.com/OpenZeppelin/openzeppelin-contracts/pull/2852

we do not need to check `address(this) == _CACHED_THIS` in `DOMAIN_SEPARATOR`

1) it saves 36 gas

2) it breaks ERC721Permit tokens looking to delegatecall posm. we dont expect the community to tap into posm for its ERC721Permit implementation. and if they do, they're bad devs 👿

src/base/EIP712.sol
saucepoint: increased first-cold-permit by 12 gas but saved 12 gas if its warm already

snreynolds (dismissed):
st,small,507x507-pad,600x600,f8f8f8 u1

src/base/EIP712.sol
```diff
@@ -0,0 +1,43 @@
...
+/// @dev Reference: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/utils/cryptography/EIP712.sol
+contract EIP712 is IEIP712 {
+    // Cache the domain separator as an immutable value, but also store the chain id that it
+    // corresponds to, in order to invalidate the cached domain separator if the chain id changes.
+    bytes32 private immutable _CACHED_DOMAIN_SEPARATOR;
+    uint256 private immutable _CACHED_CHAIN_ID;
+    bytes32 private immutable _HASHED_NAME;
+
+    bytes32 private constant _TYPE_HASH =
+        keccak256("EIP712Domain(string name,uint256 chainId,address verifyingContract)");
```
hensha256, line 19: why dont we want a version string? i thought part of the idea was that we might have multiple versions? @snreynolds

src/base/EIP712.sol
```diff
@@ -0,0 +1,43 @@
...
+
+    /// @notice Returns the domain separator for the current chain.
+    /// @dev Uses cached version if chainid is unchanged from construction.
+    function DOMAIN_SEPARATOR() public view override returns (bytes32) {
+        return block.chainid == _CACHED_CHAIN_ID ? _CACHED_DOMAIN_SEPARATOR : _buildDomainSeparator();
+    }
+
+    /// @notice Builds a domain separator using the current chainId and contract address.
+    function _buildDomainSeparator() private view returns (bytes32) {
+        return keccak256(abi.encode(_TYPE_HASH, _HASHED_NAME, block.chainid, address(this)));
```
hensha256, line 36: is this save if we are delegate calling? using `address(this)`? we build the domain separator in the constructor so itll use this contract but then when it delegate calls to `permit` wont it be in the wrong contract?

hensha256: oh our multicall contract is the same address as this contract so i guess its fine

hensha256: weird tho

src/base/EIP712.sol
saucepoint: it shaves a bit of gas and sara thinks we can version in name

hensha256: it shouldnt shave anything off gas except in the constructor? after that its not calculating the typehash right?

hensha256: oh you should hardcode this hash - solidity calculates constants at runtime not compile time so itll be recalculating every time

src/base/EIP712.sol
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    /// @notice Creates an EIP-712 typed data hash
+    function _hashTypedData(bytes32 dataHash) internal view returns (bytes32) {
```
hensha256, line 40: ok i went through OZ's contracts to figure out why theirs is cheaper lol. If you rewrite this function as:
```
    function _hashTypedData(bytes32 dataHash) internal view returns (bytes32 digest) {
        bytes32 domainSeparator = DOMAIN_SEPARATOR();
        assembly("memory-safe") {
            let fmp := mload(0x40)
            mstore(fmp, hex"19_01")
            mstore(add(fmp, 0x02), domainSeparator)
            mstore(add(fmp, 0x22), dataHash)
            digest := keccak256(fmp, 0x42)
        }
    }
```

our gas will be in line with theirs again! Based on [this function](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/24a641d9c9e0137093592a466c5496315626d98d/contracts/utils/cryptography/MessageHashUtils.sol#L76)

.forge-snapshots/PositionManager_permit.snap
```diff
@@ -1 +1 @@
-79585
\ No newline at end of file
+79566
```
saucepoint, line 1: image

src/interfaces/IEIP712.sol
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface IEIP712 {
```
hensha256, line 4: I think IPositionManager should inherit this?

hensha256 (approved):

hensha256 (approved):

PR #261, opened 2024-08-03
Take command in router

hensha256:


src/PositionManager.sol
```diff
@@ -128,69 +128,75 @@ contract PositionManager is
     }
 
     function _handleAction(uint256 action, bytes calldata params) internal virtual override {
-        if (action == Actions.INCREASE_LIQUIDITY) {
```
hensha256, line 131: this looks bad but i just split the command into a nested if for gas. If you view it in codespace itll make it much easier to see what i did lol.
its like: 
```
if (action > settle) {
   ... all liquidity / mint / burn commands ...
} else {
   ... all payment commands ...
}
```

src/base/DeltaResolver.sol
```diff
@@ -82,16 +82,14 @@ abstract contract DeltaResolver is ImmutableState {
     /// @notice Calculates the amount for a take action
     function _mapTakeAmount(uint256 amount, Currency currency) internal view returns (uint256) {
         if (amount == Constants.OPEN_DELTA) {
-            return _getFullCredit(currency).toUint128();
```
hensha256, line 85: i think this was accidentally copied from the  `mapInputAmount` function which returns a uint128

src/base/DeltaResolver.sol
```diff
@@ -82,16 +82,14 @@ abstract contract DeltaResolver is ImmutableState {
...
 
     /// @notice Calculates the amount for a swap action
     function _mapInputAmount(uint128 amount, Currency currency) internal view returns (uint128) {
-        if (amount == Constants.CONTRACT_BALANCE) {
-            return currency.balanceOfSelf().toUint128();
-        } else if (amount == Constants.OPEN_DELTA) {
+        if (amount == Constants.OPEN_DELTA) {
             return _getFullCredit(currency).toUint128();
         }
         return amount;
```
saucepoint, line 95: this function is now exactly the same as `_mapTakeAmount()` (minus casting)

1) why did we get rid of `amount == Constants.CONTRACT_BALANCE`

2) can we natspec why `_getFullCredit` is used for fetching input amounts? (i'm assuming its for multihop -- the inputs of an intermediate swap is the output (credit) of the previous step)

hensha256: > why did we get rid of amount == Constants.CONTRACT_BALANCE

I just cant see a time it will be used. The entire point in this function is for 1 usecase:
- USDC traded on v2 for UNI. UNI is left in the router
- Now we want to trade UNI for PEPE on v4 how do we do that?

Previously i thought we would do `swap with input Constants.CONTRACT_BALANCE`... and then do `settle OPEN_DELTA`. But Ive realised we're duplicating logic because you can do `settle CONTRACT_BALANCE` and then `swap OPEN_DELTA`.

Maybe its worth having both? idk just feels like a lot of options

saucepoint (approved):

PR #267, opened 2024-08-04
Align constants with UR

hensha256:


src/libraries/FlagConstants.sol
saucepoint, line 1: I think `ActionConstants` is a bit more descriptive

hensha256: YES love that

saucepoint (approved):

PR #273, opened 2024-08-05
Wrap reverts thrown by subscribers

gretzke:
## Related Issue

## Description of changes


test/mocks/MockBadSubscribers.sol
```diff
@@ -56,3 +56,45 @@ contract MockReturnDataSubscriber is ISubscriber {
...
     }
 }
+
+/// @notice A subscriber contract that returns values from the subscriber entrypoints
+contract MockRevertSubscriber is ISubscriber {
+    PositionManager posm;
+
+    error NotAuthorizedNotifer(address sender);
+
+    error TestRevert(string);
```
saucepoint, line 66: would be nice to see different error types like string errors / requires and simple errors (parameter-less)

saucepoint (approved):
looks good to me, thinking we should have some test around different error types but im confident it will handle them without problems

so send it and we can improve test later 😝

PR #277, opened 2024-08-05
Use custom revert

snreynolds:
## Related Issue

## Description of changes


src/base/Notifier.sol
```diff
@@ -56,17 +59,33 @@ abstract contract Notifier is INotifier {
...
+
+        if (!success) {
+            Wrap__TransferNotificationReverted.selector.bubbleUpAndRevertWith(address(_subscriber));
+        }
+    }
+
+    function _call(address target, bytes memory encodedCall) internal returns (bool success) {
+        assembly ("memory-safe") {
+            success := call(gas(), target, 0, add(encodedCall, 0x20), mload(encodedCall), 0, 0)
         }
```
saucepoint, line 89: wonder if this should be standardized in core's CustomRevert.sol?

src/base/Notifier.sol
```diff
@@ -35,10 +34,14 @@ abstract contract Notifier is INotifier {
...
-        try ISubscriber(newSubscriber).notifySubscribe(tokenId, config, data) {}
-        catch (bytes memory reason) {
-            revert Wrap__SubsciptionReverted(newSubscriber, reason);
+        bool success = _call(
+            address(newSubscriber), abi.encodeWithSelector(ISubscriber.notifySubscribe.selector, tokenId, config, data)
+        );
+
+        if (!success) {
+            Wrap__SubsciptionReverted.selector.bubbleUpAndRevertWith(address(newSubscriber));
         }
```
saucepoint, line 43: just my opinion -- but i dont how i feel about the manual call handler just so we can standardize around `.bubbleUpAndRevertWith`

like it feels less readable now

saucepoint (approved):

gretzke (approved):

src/base/Notifier.sol
snreynolds: yeahhh i get that

PR #279, opened 2024-08-05
add mint position event

snreynolds:
while this does add gas, we saved ~30k/40k gas not storing the PositionConfig so we're still net negative with that design change, and this is necessary to inform offchain integrations of a tokenId's associated PositionConfig

made it super minimal (no liquidity param), as anything related to getting liquidity value should be picked up from core events


saucepoint (approved):
looks good and makes sense why we need it

---

also worth mentioning `_modifyLiquidity` cannot re-enter posm to mint a position so we wont have out-of-order events 👍

snreynolds:
> looks good and makes sense why we need it
> 
> also worth mentioning `_modifyLiquidity` cannot re-enter posm to mint a position so we wont have out-of-order events 👍

I guess if a hook is the initial caller they could..

saucepoint (approved):

PR #282, opened 2024-08-06
Provide feesAccrued to subscriber.notifyModifyLiquidity

saucepoint:
## Related Issue

Provide `feesAccrued` to subscribers. Considering subscribers are staking modules, this will be may be valuable information

## Description of changes

* Pass ` feesAccrued` to subscribers in `notifyModifyLiquidity`

**Resolved Question**:
* ~~Do we need to provide `liquidityChange` AND `liquidityDelta`, it's possible to convert between the two via `LiquidityAmounts`~~
* liquidityChange and liquidityDelta are interchangeable assuming the hook did not adjust deltas. Regardless, we believe liquidityChange to be the more valuable information for staking modules


src/PositionManager.sol
```diff
@@ -345,7 +346,7 @@ contract PositionManager is
         );
 
         if (positionConfigs.hasSubscriber(uint256(salt))) {
-            _notifyModifyLiquidity(uint256(salt), config, liquidityChange);
+            _notifyModifyLiquidity(uint256(salt), config, liquidityChange, liquidityDelta, feesAccrued);
```
snreynolds, line 349: I am leaning on taking out liquidityDelta?

snreynolds (commented):
i think looks good to me, i am thinking we dont need to pass the liquidityDelta but either way Im ok with

snreynolds (approved):

PR #285, opened 2024-08-07
fix: slippage checks

snreynolds:
fixes slippage to use safe cast for decrease/burn
fixes slippage on mint to check for negative deltas
subtracts out the fee amounts before slippage checks

edge case: on an autocompounding add liq, the slippage must be the fee amounts


src/libraries/SlippageCheck.sol
```diff
@@ -2,33 +2,33 @@
...
-    /// @notice Revert if one or both deltas exceeds a maximum input
-    /// @dev When increasing liquidity, delta can be positive when excess fees need to be collected
-    /// in those cases, slippage checks are not required
-    function validateMaxInNegative(BalanceDelta delta, uint128 amount0Max, uint128 amount1Max) internal pure {
+        // Called on mint or increase, where we expect the returned delta to be negative.
+        // However, in both mint or increase, a delta could be positive on specific hooked pools.
+        // Thus we only cast the delta when the delta is negative. And that value we check against the amountMaxs.
+        // If the delta is positive, we do not need to check against amount0Maxs as the min value is 0.
         if (
             delta.amount0() < 0 && amount0Max < uint128(-delta.amount0())
```
hensha256, line 33: `uint128(-delta.amount0())`
because of custom accounting you can't do this casting

hensha256: we either need slippage vars as an `int` to enable support for custom accounting, or safe cast (which might risk locking liquidity from custom accounting pools in the pool manager)

test/position-managers/IncreaseLiquidity.t.sol
```diff
@@ -119,7 +119,10 @@ contract IncreaseLiquidityTest is Test, PosmTestSetup, Fuzzers {
 
         Plan memory planner = Planner.init();
         planner.add(
-            Actions.INCREASE_LIQUIDITY, abi.encode(tokenIdAlice, config, liquidityDelta, 0 wei, 0 wei, ZERO_BYTES)
+            Actions.INCREASE_LIQUIDITY,
+            abi.encode(
+                tokenIdAlice, config, liquidityDelta, feesOwedAlice.amount0(), feesOwedAlice.amount1(), ZERO_BYTES
```
hensha256, line 124: would love some tests where fee amount pays for the entire increase and more

hensha256: so the user has to take the excess fees at the end

src/PositionManager.sol
```diff
@@ -270,10 +276,13 @@ contract PositionManager is
         uint256 liquidity = uint256(getPositionLiquidity(tokenId, config));
 
         BalanceDelta liquidityDelta;
+        BalanceDelta feesAccrued;
```
hensha256, line 279: why are these defined outside of the if statement?

test/position-managers/IncreaseLiquidity.t.sol
```diff
@@ -30,6 +30,8 @@ import {FeeMath} from "../shared/FeeMath.sol";
 import {PosmTestSetup} from "../shared/PosmTestSetup.sol";
 import {ActionConstants} from "../../src/libraries/ActionConstants.sol";
 
+import "forge-std/console2.sol";
```
hensha256, line 33: remove :)

src/PositionManager.sol
```diff
@@ -217,8 +217,10 @@ contract PositionManager is
         bytes calldata hookData
     ) internal onlyValidConfig(tokenId, config) {
         // Note: The tokenId is used as the salt for this position, so every minted position has unique storage in the pool manager.
-        BalanceDelta liquidityDelta = _modifyLiquidity(config, liquidity.toInt256(), bytes32(tokenId), hookData);
-        liquidityDelta.validateMaxInNegative(amount0Max, amount1Max);
+        (BalanceDelta liquidityDelta, BalanceDelta feesAccrued) =
+            _modifyLiquidity(config, liquidity.toInt256(), bytes32(tokenId), hookData);
+        // Slippage checks should be done on the principle liquidityDelta which is the liquidityDelta - feesAccrued
```
hensha256, line 222:
```suggestion
        // Slippage checks should be done on the principal delta which is the liquidityDelta - feesAccrued
```

for all occurrences

hensha256: principle vs principal - for a loan etc its principal

src/libraries/SlippageCheck.sol
```diff
@@ -2,33 +2,33 @@
...
-            revert MaximumAmountExceeded();
-        }
-    }
-
-    /// @notice Revert if one or both deltas exceeds a maximum input
-    /// @dev When increasing liquidity, delta can be positive when excess fees need to be collected
-    /// in those cases, slippage checks are not required
-    function validateMaxInNegative(BalanceDelta delta, uint128 amount0Max, uint128 amount1Max) internal pure {
+        // Called on mint or increase, where we expect the returned delta to be negative.
+        // However, in both mint or increase, a delta could be positive on specific hooked pools.
```
hensha256, line 29:
```suggestion
        // However, in both mint or increase, a delta could be positive on pools where hooks can return deltas on modify liquidity.
```

src/libraries/SlippageCheck.sol
```diff
@@ -2,33 +2,33 @@
...
+        // Thus we only cast the delta when the delta is negative. And that value we check against the amountMaxs.
+        // If the delta is positive, we do not need to check against amount0Maxs as the min value is 0.
```
hensha256, line 31:
```suggestion
        // Given a positive delta means the hook is giving tokens to the user for increasing liquidity, we do not check slippage
```

hensha256: something like that?

src/PositionManager.sol
snreynolds: moved!

src/PositionManager.sol
snreynolds: principle vs principal is just something i will never get right i swear

src/libraries/SlippageCheck.sol
snreynolds: ok i fixed a bunch of natspec, tried to be as detailed as possible lmk what you think

src/PositionManager.sol
```diff
@@ -270,10 +276,13 @@ contract PositionManager is
         uint256 liquidity = uint256(getPositionLiquidity(tokenId, config));
 
         BalanceDelta liquidityDelta;
+        BalanceDelta feesAccrued;
         // Can only call modify if there is non zero liquidity.
         if (liquidity > 0) {
```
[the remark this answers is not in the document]
snreynolds: fixed

saucepoint (approved):

PR #286, opened 2024-08-07
OZ: Remove contract balance swap input

hensha256:
This branch was unreachable. Given I didn't have any use for this branch in the first place I opted to remove it instead of fixing it.


snreynolds (approved):
i think i prefer this version https://github.com/Uniswap/v4-periphery/pull/286/commits/e124d76eda9fc5150764d6226726590dab067921 

I tried some things out to consolidate some of the DeltaResolver code and couldn't get to a version that was better than just casting in router.... so will leave it up to you

PR #287, opened 2024-08-07
move sub unsub

snreynolds:
## Related Issue

## Description of changes


saucepoint (commented):
looks good! but will defer to @hensha256 for the final approval

src/libraries/PositionConfigId.sol
```diff
@@ -0,0 +1,43 @@
...
+pragma solidity ^0.8.24;
+
+import {PositionConfig, PositionConfigLibrary} from "./PositionConfig.sol";
+
+/// @notice A configId is set per tokenId
+/// The lower 255 bits are used to store the truncated hash of the corresponding PositionConfig
+/// The upper bit is used to signal if the tokenId has a subscriber
+struct PositionConfigId {
+    bytes32 id;
+}
```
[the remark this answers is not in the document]
hensha256: this is to allow it to be passed around in `storage` which you cannot do with types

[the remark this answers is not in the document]
hensha256: and this whole library has to operate over storage

src/libraries/PositionConfigId.sol
```diff
@@ -0,0 +1,43 @@
...
+    bytes32 constant DIRTY_UPPER_BIT = 0x8000000000000000000000000000000000000000000000000000000000000000;
+
+    /// @notice returns the truncated hash of the PositionConfig for a given tokenId
+    function getConfigId(PositionConfigId storage _configId) internal view returns (bytes32 configId) {
+        configId = _configId.id & MASK_UPPER_BIT;
+    }
+
+    /// @dev We only set the config on mint, guaranteeing that the most significant bit is unset, so we can just assign the entire 32 bytes to the id.
+    function setConfigId(PositionConfigId storage configId, PositionConfig calldata config) internal {
+        configId.id = config.toId();
```
hensha256, line 26: only question for me is whether this function should be

```
setConfigId(PositionConfigId storage configId, bytes32 id)
configId.id = id;
```

and then mint would do
```
positionConfigs[tokenId].setConfigId(config.toId());
```

thoughts? its not a strong opinion just a nit thought

hensha256: and then it would remove the imports of `PositionConfig` and `PositionConfigLibrary` in this file

hensha256 (dismissed):

src/libraries/PositionConfigId.sol
snreynolds: yeahh I thought about that to.. I can change it

hensha256 (dismissed):

hensha256 (approved):

PR #299, opened 2024-08-13
add script

snreynolds:
## Related Issue

## Description of changes


script/parameters/sepolia.json
saucepoint, line 1: thoughts on adding foundry-devops as a dependency?

let's us read contract addresses from previous broadcast artifacts

https://github.com/Cyfrin/foundry-devops

PR #309, opened 2024-08-28
OZ-L06: try-catch permit2Forwarder to avoid exposing DOS on multicall

saucepoint:
## Related Issue
OZ-L06: allow permit2Forwarder's `permit`/`permitBatch` to fail via try-catch.

Prevents DOS'ing multicalls containing `permit`/`permitBatch` where an external party front runs the permit2Forwarder call

## Description of changes

try-catch'd permit2Forwarder's `permit`/`permitBatch`


src/base/Permit2Forwarder.sol
```diff
@@ -27,6 +27,6 @@ contract Permit2Forwarder {
         external
         payable
     {
-        permit2.permit(owner, _permitBatch, signature);
+        try permit2.permit(owner, _permitBatch, signature) {} catch {}
```
hensha256, line 30: maybe a quick comment that explains why its in a try catch?

test/position-managers/PositionManager.multicall.t.sol
```diff
@@ -321,4 +331,76 @@ contract PositionManagerMulticallTest is Test, Permit2SignatureHelpers, PosmTest
...
+            defaultERC20PermitAllowance(Currency.unwrap(currency1), permitAmount, permitExpiration, permitNonce);
+        permit1.spender = address(lpm);
+        bytes memory sig1 = getPermitSignature(permit1, charliePK, PERMIT2_DOMAIN_SEPARATOR);
+
+        // bob front-runs the permits
+        vm.startPrank(bob);
+        lpm.permit(charlie, permit0, sig0);
+        lpm.permit(charlie, permit1, sig1);
+        vm.stopPrank();
+
```
hensha256, line 359: check that the permit has succeeded by checking permit2 has stored the permitNonce has been taken i think?

test/position-managers/PositionManager.multicall.t.sol
```diff
@@ -321,4 +331,76 @@ contract PositionManagerMulticallTest is Test, Permit2SignatureHelpers, PosmTest
...
+        // charlie tries to mint an LP token with multicall(permit, permit, mint)
+        bytes[] memory calls = new bytes[](3);
+        calls[0] = abi.encodeWithSelector(Permit2Forwarder(lpm).permit.selector, charlie, permit0, sig0);
+        calls[1] = abi.encodeWithSelector(Permit2Forwarder(lpm).permit.selector, charlie, permit1, sig1);
+        bytes memory mintCall = getMintEncoded(config, 10e18, charlie, ZERO_BYTES);
+        calls[2] = abi.encodeWithSelector(IPositionManager.modifyLiquidities.selector, mintCall, _deadline);
+
+        uint256 tokenId = lpm.nextTokenId();
+        lpm.multicall(calls);
```
hensha256, line 368: ideally this test would check
```
bytes[] memory results = lpm.multicall(calls);
assert(results[0] is correct revert reason) 
assert(results[1] is correct revert reason) 
```

i guess that revert reason is that the nonce is taken

test/position-managers/PositionManager.multicall.t.sol
```diff
@@ -321,4 +331,76 @@ contract PositionManagerMulticallTest is Test, Permit2SignatureHelpers, PosmTest
...
+        tokens[1] = Currency.unwrap(currency1);
+
+        IAllowanceTransfer.PermitBatch memory permit =
+            defaultERC20PermitBatchAllowance(tokens, permitAmount, permitExpiration, permitNonce);
+        permit.spender = address(lpm);
+        bytes memory sig = getPermitBatchSignature(permit, charliePK, PERMIT2_DOMAIN_SEPARATOR);
+
+        // bob front-runs the permits
+        vm.prank(bob);
+        lpm.permitBatch(charlie, permit, sig);
```
dianakocsis, line 393: stop prank

test/position-managers/PositionManager.multicall.t.sol
```diff
@@ -321,4 +331,76 @@ contract PositionManagerMulticallTest is Test, Permit2SignatureHelpers, PosmTest
...
+        bytes[] memory calls = new bytes[](3);
+        calls[0] = abi.encodeWithSelector(Permit2Forwarder(lpm).permit.selector, charlie, permit0, sig0);
+        calls[1] = abi.encodeWithSelector(Permit2Forwarder(lpm).permit.selector, charlie, permit1, sig1);
+        bytes memory mintCall = getMintEncoded(config, 10e18, charlie, ZERO_BYTES);
+        calls[2] = abi.encodeWithSelector(IPositionManager.modifyLiquidities.selector, mintCall, _deadline);
+
+        uint256 tokenId = lpm.nextTokenId();
+        lpm.multicall(calls);
+
+        assertEq(lpm.ownerOf(tokenId), charlie);
```
dianakocsis, line 370: could you also assert that charlie was not the owner before the call? and in the other test too

dianakocsis (changes requested):
looks great! just 2 things

test/position-managers/PositionManager.multicall.t.sol
hensha256: you dont need to stop after `prank`
`vm.startPrank` pranks until you call `vm.stopPrank`
if you just use `vm.prank` it only pranks 1 call and theres no need to stop it

test/position-managers/PositionManager.multicall.t.sol
saucepoint: mm good idea now thats try-catch'd

test/position-managers/PositionManager.multicall.t.sol
saucepoint: can you take a look now? im trying to do the Wrap__ error handling

cc @snreynolds

snreynolds (commented):
hmm to me I feel like we should allow the caller to specify whether or not they want to continue the multicall if there is an error

also if permit errors (even for a different reason) the rest of the code will still execute in this case

saucepoint:
> hmm to me I feel like we should allow the caller to specify whether or not they want to continue the multicall if there is an error

the idea being we expose another generic `multicall(bytes calldata calls, bool revert)` where `revert = true` will cause a halt?

hensha256 (approved):

dianakocsis (approved):

PR #310, opened 2024-08-28
feat: add reverse mapping for pool key, store tL, tU

snreynolds:
motivation was to have better interface standardization for ERC721s
(functions dont require PositionConfig to be passed in)

- & you get an onchain lookup of poolKey, tickLower, tickUpper
- & position descriptor can read from config

I personally was against this for gas reasons (mint is 40-50K more expensive) but that should go down with mints on the same pool, and overhead for other ops is around 4-5K.


.forge-snapshots/PositionManager_mint_native.snap
```diff
@@ -1 +1 @@
-341353
\ No newline at end of file
+370766
```
saucepoint, line 1: i regret to inform yall that we have, unfortunately, lost sara to ultrasound money gang 😭 

pouring one out for all my mainnet retail LPs 🙏

snreynolds: 🪦

hensha256:
>(mint is 40-50K more expensive

so its like 8k more for mints on the same pool right? which is the "normal" case?

src/libraries/PositionConfig.sol
```diff
@@ -9,27 +9,3 @@ struct PositionConfig {
     int24 tickLower;
     int24 tickUpper;
 }
-
```
hensha256, line 12: now that this is just a struct shall it just go into IPositionManager?

src/interfaces/ISubscriber.sol
```diff
@@ -7,23 +7,15 @@ import {BalanceDelta} from "@uniswap/v4-core/src/types/BalanceDelta.sol";
...
     /// @param tokenId the token ID of the position
-    /// @param config details about the position
     /// @param data additional data passed in by the caller
-    function notifySubscribe(uint256 tokenId, PositionConfig memory config, bytes memory data) external;
+    function notifySubscribe(uint256 tokenId, bytes memory data) external;
     /// @param tokenId the token ID of the position
-    /// @param config details about the position
     /// @param data additional data passed in by the caller
-    function notifyUnsubscribe(uint256 tokenId, PositionConfig memory config, bytes memory data) external;
+    function notifyUnsubscribe(uint256 tokenId, bytes memory data) external;
```
hensha256, line 14: i feel like the subscriber would still want this information? imo it should still be sent

src/PositionManager.sol
```diff
@@ -110,23 +110,32 @@ contract PositionManager is
...
+    // TODO: Move to custom type, packed memory.
+    struct PositionInfo {
+        // lower 25 bytes of the poolId
+        bytes25 poolId;
+        int24 tickLower;
+        int24 tickUpper;
+        bool hasSubscriber;
+    }
+
+    mapping(uint256 tokenId => PositionInfo info) internal positionInfo;
```
hensha256, line 133: public?

hensha256: unless im missing something? how will position descriptor call to fetch this information?

src/PositionManager.sol
```diff
@@ -110,23 +110,32 @@ contract PositionManager is
...
+    mapping(bytes32 poolId => PoolKey poolKey) internal poolKeys;
```
hensha256, line 134: bytes25?

hensha256: public?

src/PositionManager.sol
```diff
@@ -432,8 +430,7 @@ contract PositionManager is
         liquidity = poolManager.getPositionLiquidity(config.poolKey.toId(), positionId);
     }
 
-    /// @inheritdoc IPositionManager
-    function getPositionConfigId(uint256 tokenId) external view returns (bytes32) {
-        return positionConfigs[tokenId].getConfigId();
+    function _toConfig(PositionInfo memory info) internal view returns (PositionConfig memory config) {
```
hensha256, line 433: should PositionInfo by a library with this function in it?
im open

src/interfaces/ISubscriber.sol
snreynolds: subscriber can just read it tho so if they dont need it its worse. i think we should just remove it and now its consistent with notifyTransfer

src/PositionManager.sol
snreynolds: yeah was gonna move some things around just wanted to check people liked the basic idea

snreynolds:
> > (mint is 40-50K more expensive
> 
> so its like 8k more for mints on the same pool right? which is the "normal" case?

not sure I can add a test but sounds about right it definitely wont be 40-50K

src/interfaces/ISubscriber.sol
snreynolds: also it means more standardized to ERC721s not just PositionManager which is I think the point of this whole change no?

src/PositionManager.sol
```diff
@@ -259,37 +236,56 @@ contract PositionManager is
...
     function _increase(
         uint256 tokenId,
-        PositionConfig calldata config,
         uint256 liquidity,
         uint128 amount0Max,
         uint128 amount1Max,
         bytes calldata hookData
-    ) internal onlyIfApproved(msgSender(), tokenId) onlyValidConfig(tokenId, config) {
+    ) internal onlyIfApproved(msgSender(), tokenId) {
+        (PositionInfo memory info, PoolKey memory poolKey) = getPoolPositionInfo(tokenId);
```
saucepoint, line 244: are there gas savings if we dont load the entirety of PositionInfo? we're not using the truncated poolId obviously, and the ticks are just repacked to IPoolManager.ModifyLiquidityParams

something like this could be quite cool
```solidity
(
    PoolKey memory key,
    IPoolManager.ModifyLiquidityParams memory liquidityParams,
    bool hasSubscriber
) = veryAwesomeFunction(tokenId);

liquidityParams.liquidityDelta = liquidityChange;
```

theres also ways to pack it further since PoolKey is 528 bits, so we can fit tL/tU/hasSubscriber in the remaining bits of the 3rd word

saucepoint (commented):
despite the massive diff, i quite like this

something about not checking subscriber on mint feels nice, same with maybe making burn only call unsubscribe

src/PositionManager.sol
```diff
@@ -304,37 +300,50 @@ contract PositionManager is
...
 
-        delete positionConfigs[tokenId];
+        delete positionInfo[tokenId];
         // Burn the token.
         _burn(tokenId);
+
+        // TODO: This will be changed from audit issue..
+        if (info.hasSubscriber) {
+            _notifyModifyLiquidity(uint256(tokenId), -(liquidity.toInt256()), feesAccrued);
+        }
```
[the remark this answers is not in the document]
snreynolds: I guess if burn does call modifyLiquidity then we should

src/PositionManager.sol
```diff
@@ -415,25 +402,47 @@ contract PositionManager is
...
+        poolKey = poolKeys[info.poolId];
     }
 
     /// @inheritdoc IPositionManager
-    function getPositionConfigId(uint256 tokenId) external view returns (bytes32) {
-        return positionConfigs[tokenId].getConfigId();
+    function getPositionLiquidity(uint256 tokenId) public view returns (uint128 liquidity) {
+        (PositionInfo memory info, PoolKey memory poolKey) = getPoolPositionInfo(tokenId);
+        liquidity = _getLiquidity(tokenId, poolKey, info.tickLower, info.tickUpper);
+    }
```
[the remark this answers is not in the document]
snreynolds: external is a good call

test/erc721Permit/ERC721Permit.permit.t.sol
```diff
@@ -229,7 +229,8 @@ contract ERC721PermitTest is Test {
         assertEq(erc721Permit.nonces(alice, wordPos) & (1 << bitPos), 0);
     }
 
-    function test_fuzz_erc721Permit_SignatureDeadlineExpired(address spender) public {
+    function test_fuzz_erc721Permit_SignatureDeadlineExpired() public {
+        address spender = 0xf30b182eA406CC5A71b8A44e2FE50cC80383F914;
```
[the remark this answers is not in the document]
snreynolds: oh i was trying something out.. via-ir breaks this test. even on main

src/PositionManager.sol
snreynolds: yeah was thinking the same thing.. I think I'm actually happy with it now that I have a version where the PositionInfo is packed (uint256).. since its the minimum we can basically return

src/libraries/PositionConfig.sol
```diff
@@ -3,33 +3,9 @@ pragma solidity ^0.8.24;
 
 import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
 
-// A PositionConfig is the input for creating and modifying a Position in core, whose truncated hash is set per tokenId
+// A PositionConfig is the input for creating and modifying a Position in core
```
snreynolds, line 6: Keeping this for now to reduce merge conflicts

later can move to test utils

src/interfaces/ISubscriber.sol
hensha256: yeah thats all fair, agreed

src/base/Notifier.sol
```diff
@@ -30,27 +29,23 @@ abstract contract Notifier is INotifier {
...
     modifier onlyIfApproved(address caller, uint256 tokenId) virtual;
-    modifier onlyValidConfig(uint256 tokenId, PositionConfig calldata config) virtual;
-
-    function _positionConfigs(uint256 tokenId) internal view virtual returns (PositionConfigId storage);
+    function _setUnsubscribe(uint256 tokenId) internal virtual;
+    function _setSubscribe(uint256 tokenId) internal virtual;
 
     /// @inheritdoc INotifier
-    function subscribe(uint256 tokenId, PositionConfig calldata config, address newSubscriber, bytes calldata data)
+    function subscribe(uint256 tokenId, address newSubscriber, bytes calldata data)
```
hensha256, line 36: i found the flow of setting subscription/unsubscription very hard to follow

so we have 2 ways of showing that someone is subscribed (as previously), both in the new PositionInfo and in the mapping

The mapping `subscriber` is easy to follow
- look it up in the mapping
- check theyre not subscribed already
- set the new subscriber

To check that position info boolean is subscribed
- external function `subscribe`
- calls virtual function `_subscribe`
- implementation of `_subscribe` is in position manager
- that implementation calls another function `setSubscribe`
- when you read `setSubscribe` it says `/// @ dev this does not actually set any storage`... so its still not set that its subscribed
- then back to position manager to find the actual line that then sets the return data into storage

hensha256: basically that second flow feels quite unnatural to follow currently, trying to think through how it could be better designed

src/PositionManager.sol
```diff
@@ -110,24 +110,20 @@ contract PositionManager is
...
-    using PositionConfigLibrary for PositionConfig;
     using StateLibrary for IPoolManager;
     using TransientStateLibrary for IPoolManager;
     using SafeCast for uint256;
     using SafeCast for int256;
     using CalldataDecoder for bytes;
     using SlippageCheckLibrary for BalanceDelta;
-    using PositionConfigIdLibrary for PositionConfigId;
+    using PoolKeyChecker for PoolKey;
+    using PositionInfoLibrary for PackedPositionInfo;
```
hensha256, line 120: nit: would rather the names were the same like
`PositionInfo` and `PositionInfoLibrary`
or `PackedPositionInfo` and `PackedPositionInfoLibrary`

src/PositionManager.sol
```diff
@@ -110,24 +110,20 @@ contract PositionManager is
...
     /// @dev The ID of the next token that will be minted. Skips 0
     uint256 public nextTokenId = 1;
 
-    mapping(uint256 tokenId => PositionConfigId configId) internal positionConfigs;
-
-    /// @notice an internal getter for PositionConfigId to be used by Notifier
-    function _positionConfigs(uint256 tokenId) internal view override returns (PositionConfigId storage) {
-        return positionConfigs[tokenId];
-    }
+    mapping(uint256 tokenId => PackedPositionInfo info) public positionInfo;
```
hensha256, line 125: now that its packed maybe worth having this internal and exposing a getter that returns the elements individually?

src/PositionManager.sol
```diff
@@ -259,37 +232,41 @@ contract PositionManager is
...
     function _increase(
         uint256 tokenId,
-        PositionConfig calldata config,
         uint256 liquidity,
         uint128 amount0Max,
         uint128 amount1Max,
         bytes calldata hookData
-    ) internal onlyIfApproved(msgSender(), tokenId) onlyValidConfig(tokenId, config) {
+    ) internal onlyIfApproved(msgSender(), tokenId) {
+        (PackedPositionInfo info, PoolKey memory poolKey) = getPoolPositionInfo(tokenId);
```
hensha256, line 240: maybe im wrong but i imagine it would be decently cheaper to just get these directly from the mappings inline instead of calling a public function

```
PositionInfo info = positionInfo[tokenId];
...
_modifyLiquidity(info, poolKeys[info.poolId()], liquidity.toInt256(), bytes32(tokenId), hookData);
```

src/libraries/PoolKeyChecker.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.24;
+
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+
+library PoolKeyChecker {
+    // On UniswapV4, the minimum tick spacing is 1.
+    // This means that if the tick spacing is 0, the pool key has not been set.
+    function isEmpty(PoolKey memory poolKey) internal pure returns (bool) {
```
hensha256, line 9: this library is just used for 1 single function call, and its 1 line of code.
I think I'd personally rather have no library and just do it inline in PosM like

```
        // Store the poolKey if it is not already stored.
        if (poolKeys[info.poolId()].tickSpacing == 0) {
            poolKeys[info.poolId()] = poolKey;
        }
```

src/PositionManager.sol
```diff
@@ -304,35 +281,39 @@ contract PositionManager is
...
-            _modifyLiquidity(config, liquidity.toInt256(), bytes32(tokenId), hookData);
+            _modifyLiquidity(info, poolKey, liquidity.toInt256(), bytes32(tokenId), hookData);
         // Slippage checks should be done on the principal liquidityDelta which is the liquidityDelta - feesAccrued
         (liquidityDelta - feesAccrued).validateMaxIn(amount0Max, amount1Max);
-        positionConfigs[tokenId].setConfigId(config.toId());
 
-        emit MintPosition(tokenId, config);
+        // Store the poolKey if it is not already stored.
+        if (poolKeys[info.poolId()].isEmpty()) {
+            poolKeys[info.poolId()] = poolKey;
```
hensha256, line 295: double calculating `info.poolId()`

src/PositionManager.sol
hensha256: I prefer the first one i think. Packing it is just an implementation detail - so `PositionInfo` feels better

src/PositionManager.sol
hensha256: i do see its neater though, not against keeping it
I'd maybe call it `getPoolAndPositionInfo`

src/PositionManager.sol
```diff
@@ -383,57 +364,78 @@ contract PositionManager is
         if (balance > 0) currency.transfer(to, balance);
     }
 
+    // implementation of abstract function DeltaResolver._pay
+    function _pay(Currency currency, address payer, uint256 amount) internal override {
+        if (payer == address(this)) {
+            // TODO: currency is guaranteed to not be eth so the native check in transfer is not optimal.
```
hensha256, line 370: remove the TODO... unless you actually plan to make this change by tuesday?

src/PositionManager.sol
```diff
@@ -304,35 +281,39 @@ contract PositionManager is
...
         if (liquidity > 0) {
             (BalanceDelta liquidityDelta, BalanceDelta feesAccrued) =
-                _modifyLiquidity(config, -(liquidity.toInt256()), bytes32(tokenId), hookData);
+                _modifyLiquidity(info, poolKey, -(liquidity.toInt256()), bytes32(tokenId), hookData);
             // Slippage checks should be done on the principal liquidityDelta which is the liquidityDelta - feesAccrued
             (liquidityDelta - feesAccrued).validateMinOut(amount0Min, amount1Min);
         }
 
-        delete positionConfigs[tokenId];
+        positionInfo[tokenId] = PositionInfoLibrary.EMPTY_POSITION_INFO;
```
hensha256, line 316: cant you just do `delete positionInfo[tokenId]`?
and then `EMPTY_POSITION_INFO` is unused

src/libraries/PositionInfoLibrary.sol
```diff
@@ -0,0 +1,103 @@
...
+ * A flag to know if the tokenId is subscribed to an address
+ * uint8 hasSubscriber;
+ *
+ * The tickUpper of the position
+ * int24 tickUpper;
+ *
+ * The tickLower of the position
+ * int24 tickLower;
+ *
+ * The truncated poolId
```
hensha256, line 25: more description ideally, which end is truncated?

src/base/Notifier.sol
snreynolds: we used to be able to set it in notifiers but bc the mapping is now a uint256 I cant pass a storage flag so it has to be set in manager.

if you have suggestions lmk!

src/PositionManager.sol
snreynolds: Yeah I just didnt wanna save a variable cause I was hitting stack too deep like every 5 mins last night

src/PositionManager.sol
snreynolds: that todo is on main?

src/PositionManager.sol
snreynolds: you can't do that cause its a uint256

src/PositionManager.sol
snreynolds: sgtm!

src/PositionManager.sol
snreynolds: hm I dont feel strongly about it. For offchain, a simple sdk helper can dissect it. and onchain someone can just import the PositionInfoLibrary?

But if you prefer I can add

src/PositionManager.sol
snreynolds: either way. I just did it cause it was repeated in 4 funcs

src/libraries/PoolKeyChecker.sol
snreynolds: I mostly liked it as a lib cause its not obvious that tickSpacing should be checked so i was just getting it out of the main codepath but i can change if you would prefer

src/PositionManager.sol
hensha256: yeah i know, i just think we should remove it given we're finalising periphery

src/base/Notifier.sol
```diff
@@ -30,26 +28,24 @@ abstract contract Notifier is INotifier {
...
         external
         payable
         onlyIfApproved(msg.sender, tokenId)
-        onlyValidConfig(tokenId, config)
     {
-        // will revert below if the user already has a subcriber
-        _positionConfigs(tokenId).setSubscribe();
+        _setSubscribe(tokenId);
```
hensha256, line 42: i think to officially follow checks-effects-interactions as our audits have brought up this should go after the check that
```
if (_subscriber != NO_SUBSCRIBER)
```

so it does that check, then sets subscribers in both ways, and then does the call

src/base/Notifier.sol
```diff
@@ -59,34 +55,24 @@ abstract contract Notifier is INotifier {
...
-    function unsubscribe(uint256 tokenId, PositionConfig calldata config, bytes calldata data)
-        external
-        payable
-        onlyIfApproved(msg.sender, tokenId)
-        onlyValidConfig(tokenId, config)
-    {
-        _positionConfigs(tokenId).setUnsubscribe();
+    function unsubscribe(uint256 tokenId, bytes calldata data) external payable onlyIfApproved(msg.sender, tokenId) {
+        _setUnsubscribe(tokenId);
         ISubscriber _subscriber = subscriber[tokenId];
```
hensha256, line 60: should this revert if they dont already have a subscriber?

src/libraries/PositionInfoLibrary.sol
```diff
@@ -0,0 +1,103 @@
...
+    function setSubscribe(PositionInfo info) internal pure returns (PositionInfo _info) {
+        assembly ("memory-safe") {
+            _info := or(info, SET_SUBSCRIBE)
+        }
+    }
+
+    /// @dev this does not actually set any storage
+    function setUnsubscribe(PositionInfo info) internal pure returns (PositionInfo _info) {
+        assembly ("memory-safe") {
+            _info := and(info, not(SET_SUBSCRIBE))
```
hensha256, line 79: doesnt this set it to `0xfffffffe`? wouldnt we rather just set it to 0?

src/interfaces/IPositionManager.sol
```diff
@@ -29,16 +27,12 @@ interface IPositionManager is INotifier {
...
     /// @return liquidity the position's liquidity, as a liquidityAmount
     /// @dev this value can be processed as an amount0 and amount1 by using the LiquidityAmounts library
-    function getPositionLiquidity(uint256 tokenId, PositionConfig calldata config)
-        external
-        view
-        returns (uint128 liquidity);
+    function getPositionLiquidity(uint256 tokenId) external view returns (uint128 liquidity);
+
+    /// @param tokenId the ERC721 tokenId
+    /// @return  a uint256 packed value holding information about the position including the range (tickLower, tickUpper)
```
hensha256, line 35:
```suggestion
    /// @return a uint256 packed value holding information about the position including the range (tickLower, tickUpper)
```

src/PositionManager.sol
hensha256: oh i see its literally from this file - how come youve moved it up? the diff confused me

hensha256: esp if we're getting this audited would be good to have a minimal diff

src/PositionManager.sol
snreynolds: sorry maybe its actually cause its a custom type.. im not sure but I tried this and it doesnt work

test/libraries/CalldataDecoder.t.sol
```diff
@@ -19,48 +19,39 @@ contract CalldataDecoderTest is Test {
...
             decoder.decodeBurnParams(params);
 
         assertEq(tokenId, _tokenId);
         assertEq(hookData, _hookData);
-        _assertEq(_config, config);
         assertEq(amount0Min, _amount0Min);
         assertEq(amount1Min, _amount1Min);
     }
 
+    // TODO: fix stack too deep here
```
hensha256, line 54: is it fixed now?

src/libraries/PositionConfig.sol
hensha256: woudl rather delete now that we arent using it

test/libraries/CalldataDecoder.t.sol
snreynolds: yes i just removed that comment! thanks

test/mocks/MockCalldataDecoder.sol
```diff
@@ -1,40 +1,38 @@
...
+    struct MintParams {
+        PoolKey poolKey;
+        int24 tickLower;
+        int24 tickUpper;
+        uint256 liquidity;
+        uint128 amount0Max;
+        uint128 amount1Max;
+        address owner;
+        bytes hookData;
+    }
```
saucepoint, line 22: this is struct-ified to avoid IR? if so, maybe worth adding a small comment why its struct-ified in tests and only for mint

src/PositionManager.sol
```diff
@@ -412,25 +395,40 @@ contract PositionManager is
...
 
+    /// @notice an internal helper used by Notifier
+    function _setSubscribe(uint256 tokenId) internal override {
+        positionInfo[tokenId] = positionInfo[tokenId].setSubscribe();
+    }
+
+    /// @notice an internal helper used by Notifier
+    function _setUnsubscribe(uint256 tokenId) internal override {
+        positionInfo[tokenId] = positionInfo[tokenId].setUnsubscribe();
+    }
```
saucepoint, line 406: mini nit: should we rename this to `_setSubscribed` and `_setUnsubscribed` just to be clearer. Feel like subscribe/subscriber is quite overloaded

src/libraries/PositionInfoLibrary.sol
```diff
@@ -0,0 +1,104 @@
...
+    uint256 internal constant SET_UNSUBSCRIBE = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF00;
+    uint256 internal constant SET_SUBSCRIBE = 0x01;
+    uint8 internal constant TICK_LOWER_OFFSET = 8;
+    uint8 internal constant TICK_UPPER_OFFSET = 32;
+
+    function poolId(PositionInfo info) internal pure returns (bytes25 _poolId) {
+        assembly ("memory-safe") {
+            _poolId := and(MASK_UPPER_200_BITS, info)
+        }
+    }
```
saucepoint, line 50: Lets add a natspec saying `info.poolId()` is NOT compatible with `PoolId` on core.

i'm tempted to suggest calling it something else like `truncatedPoolId()` but 🤮

saucepoint (approved):
very clean work per usual 😌

src/libraries/PositionInfoLibrary.sol
snreynolds: Yeah good call out, I'll add natspec 😛

test/mocks/MockCalldataDecoder.sol
snreynolds: yes will do

snreynolds: and yes to avoid ir lol

src/PositionManager.sol
```diff
@@ -412,25 +395,40 @@ contract PositionManager is
...
     /// @inheritdoc IPositionManager
-    function getPositionLiquidity(uint256 tokenId, PositionConfig calldata config)
-        public
-        view
-        returns (uint128 liquidity)
-    {
-        bytes32 positionId =
-            Position.calculatePositionKey(address(this), config.tickLower, config.tickUpper, bytes32(tokenId));
-        liquidity = poolManager.getPositionLiquidity(config.poolKey.toId(), positionId);
+    function getPoolAndPositionInfo(uint256 tokenId) public view returns (PositionInfo info, PoolKey memory poolKey) {
```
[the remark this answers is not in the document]
snreynolds: yep!

saucepoint:
reminder: consider SB-86, hasSubscriber should revert if the token does not exist

stance: probably ugly to enforce, and adds gas. BUT should be natspec'd in hasSubscriber function

src/PositionManager.sol
```diff
@@ -301,35 +276,41 @@ contract PositionManager is
...
         (liquidityDelta - feesAccrued).validateMaxIn(amount0Max, amount1Max);
-        positionConfigs[tokenId].setConfigId(config.toId());
 
-        emit MintPosition(tokenId, config);
+        bytes25 poolId = info.poolId();
+        // Store the poolKey if it is not already stored.
+        // On UniswapV4, the minimum tick spacing is 1, which means that if the tick spacing is 0, the pool key has not been set.
+        if (poolKeys[poolId].tickSpacing == 0) {
+            poolKeys[poolId] = poolKey;
+        }
```
saucepoint, line 293: let's do this write prior to `_modifyLiquidity` for CEI anyway (the hook still has access to tick range and poolKey regardless of CEI)

src/PositionManager.sol
```diff
@@ -301,35 +276,41 @@ contract PositionManager is
...
-                _modifyLiquidity(config, -(liquidity.toInt256()), bytes32(tokenId), hookData);
+                _modifyLiquidity(info, poolKey, -(liquidity.toInt256()), bytes32(tokenId), hookData);
             // Slippage checks should be done on the principal liquidityDelta which is the liquidityDelta - feesAccrued
             (liquidityDelta - feesAccrued).validateMinOut(amount0Min, amount1Min);
         }
 
-        delete positionConfigs[tokenId];
+        positionInfo[tokenId] = PositionInfoLibrary.EMPTY_POSITION_INFO;
         // Burn the token.
         _burn(tokenId);
```
saucepoint, line 315: thinking this logic should run prior to `_modifyLiquidity` for CEI

saucepoint (commented):
some CEI violations in _mint / _burn

https://github.com/spearbit-audits/review-uniswap-v4/issues/93

src/PositionManager.sol
saucepoint: solmate _burn might be save to leave it, but positionInfo may want to happen before?

src/libraries/PositionInfoLibrary.sol
```diff
@@ -0,0 +1,105 @@
...
+    function initialize(PoolKey memory _poolKey, int24 _tickLower, int24 _tickUpper)
+        internal
+        pure
+        returns (PositionInfo info)
+    {
+        bytes25 _poolId = bytes25(PoolId.unwrap(_poolKey.toId()));
+        assembly {
+            info :=
+                or(
+                    or(and(MASK_UPPER_200_BITS, _poolId), shl(TICK_UPPER_OFFSET, and(MASK_24_BITS, _tickUpper))),
```
0age, line 100: I think this mask is redundant as it's already bytes25

src/PositionManager.sol
snreynolds: ah yeah that makes sense

test/mocks/MockCalldataDecoder.sol
```diff
@@ -1,40 +1,39 @@
...
 import {CalldataDecoder} from "../../src/libraries/CalldataDecoder.sol";
 import {IV4Router} from "../../src/interfaces/IV4Router.sol";
 import {Currency} from "@uniswap/v4-core/src/types/Currency.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
 
 // we need to use a mock contract to make the calls happen in calldata not memory
 contract MockCalldataDecoder {
     using CalldataDecoder for bytes;
 
+    // This is used to avoid compiling with via-ir.
```
saucepoint, line 13: image

test/libraries/PositionInfoLibrary.t.sol
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    function test_fuzz_initialize_setSubscribed(PoolKey memory poolKey, int24 tickLower, int24 tickUpper) public pure {
+        PositionInfo info = PositionInfoLibrary.initialize(poolKey, tickLower, tickUpper);
+        assertEq(info.hasSubscriber(), false);
+        info = info.setSubscribe();
+        assertEq(info.hasSubscriber(), true);
+        assertEq(info.tickLower(), tickLower);
+        assertEq(info.tickUpper(), tickUpper);
+        assertEq(info.poolId(), bytes25(PoolId.unwrap(poolKey.toId())));
```
saucepoint, line 31: library itself looks good, but as an insane person we should assert that the tickLower/tickUpper/poolId is NOT changed by `.setSubscribe`

i.e. asserting that the info data is the same before/after `.setSubscribe`

src/PositionManager.sol
```diff
@@ -310,39 +285,46 @@ contract PositionManager is
...
+        bytes25 poolId = info.poolId();
+        if (poolKeys[poolId].tickSpacing == 0) {
+            poolKeys[poolId] = poolKey;
+        }
 
-        emit MintPosition(tokenId, config);
+        // fee delta can be ignored as this is a new position
+        (BalanceDelta liquidityDelta,) =
+            _modifyLiquidity(info, poolKey, liquidity.toInt256(), bytes32(tokenId), hookData);
+        (liquidityDelta).validateMaxIn(amount0Max, amount1Max);
```
hensha256, line 302:
```suggestion
        liquidityDelta.validateMaxIn(amount0Max, amount1Max);
```

src/PositionManager.sol
```diff
@@ -391,24 +373,26 @@ contract PositionManager is
...
+                tickUpper: info.tickUpper(),
                 liquidityDelta: liquidityChange,
                 salt: salt
             }),
             hookData
         );
 
-        if (positionConfigs[uint256(salt)].hasSubscriber()) {
-            _notifyModifyLiquidity(uint256(salt), config, liquidityChange, feesAccrued);
+        // TODO: Audit issue for burn, decide if we want to keep this and also unsubscribe.
```
hensha256, line 393: remove

hensha256 (dismissed):

hensha256 (approved):

hensha256 (approved):

PR #311, opened 2024-08-28
OZ L-11

dianakocsis:
## Related Issue
OZ L-11 Discrepancies Between Interfaces and Implementation Contracts
https://defender.openzeppelin.com/v2/#/audit/43d4e491-c5b4-42ba-b0fc-7f5adc46a6b6/issues/L-11

## Description of changes
fix discrepancies between interfaces and implementation contracts

### DIFF
--- a/src/base/EIP712_v4.sol
+++ b/src/base/EIP712_v4.sol
@@ -26,9 +26,8 @@ contract EIP712_v4 is IEIP712_v4 {
     }
 
     /// @inheritdoc IEIP712_v4
-    /// @notice Returns the domain separator for the current chain.
-    /// @dev Uses cached version if chainid is unchanged from construction.
     function DOMAIN_SEPARATOR() public view override returns (bytes32) {
+        // uses cached version if chainid is unchanged from construction
         return block.chainid == _CACHED_CHAIN_ID ? _CACHED_DOMAIN_SEPARATOR : _buildDomainSeparator();
     }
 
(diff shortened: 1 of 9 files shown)


src/base/EIP712_v4.sol
```diff
@@ -25,6 +25,7 @@ contract EIP712_v4 is IEIP712_v4 {
         _CACHED_DOMAIN_SEPARATOR = _buildDomainSeparator();
     }
 
+    /// @inheritdoc IEIP712_v4
     /// @notice Returns the domain separator for the current chain.
     /// @dev Uses cached version if chainid is unchanged from construction.
```
saucepoint, line 30: can remove now that we're inheriting from the interface

src/lens/Quoter.sol
```diff
@@ -55,7 +55,7 @@ contract Quoter is IQuoter, SafeCallback {
     constructor(IPoolManager _poolManager) SafeCallback(_poolManager) {}
 
     /// @inheritdoc IQuoter
-    function quoteExactInputSingle(QuoteExactSingleParams memory params)
+    function quoteExactInputSingle(QuoteExactSingleParams calldata params)
```
saucepoint, line 58: in the OZ issue, i think we wanted memory on both implementation and interface. I think there were incidents in IQuoter that were using calldata.

Because we are redoing Quoter entirely, i dont think we need to change it back

saucepoint (commented):
also: 
```
    @inheritdoc BaseActionsRouter
    function msgSender() public view override returns (address) {
        return _getLocker();
    }
```

saucepoint (approved):

PR #314, opened 2024-08-28
OZ-L07: Unsubscribe on Burn (+ remove arbitrary data)

saucepoint:
## Related Issue
OZ-L07: burning a position should unsubscribe a position

## Description of changes

When burning a position, if it has a subscriber, automatically unsubscribe.

# Big Change:

Removed `bytes memory data` from `unsubscribe()`
* Passing additional data to burn adds gas overhead for obscure cases
* Users can forcibly provide invalid `data` and still be able to unsubscribe. I.e. There was never really an enforcement that `data` was valid


src/PositionManager.sol
```diff
@@ -327,6 +327,8 @@ contract PositionManager is
             (liquidityDelta - feesAccrued).validateMinOut(amount0Min, amount1Min);
         }
 
+        if (positionConfigs[tokenId].hasSubscriber()) _unsubscribe(tokenId, config);
```
hensha256, line 330: this somewhat inefficiently must be doing an SSTORE of
positionConfig.isSubscribed = false

and then because it burns it then just deletes the position config anyway?

hensha256: but probably much neater code to keep it like this

src/base/Notifier.sol
```diff
@@ -59,19 +59,23 @@ abstract contract Notifier is INotifier {
...
+
+    function _unsubscribe(uint256 tokenId, PositionConfig calldata config) internal {
         _positionConfigs(tokenId).setUnsubscribe();
         ISubscriber _subscriber = subscriber[tokenId];
 
         delete subscriber[tokenId];
 
         uint256 subscriberGasLimit = block.gaslimit.calculatePortion(BLOCK_LIMIT_BPS);
-        try _subscriber.notifyUnsubscribe{gas: subscriberGasLimit}(tokenId, config, data) {} catch {}
+        try _subscriber.notifyUnsubscribe{gas: subscriberGasLimit}(tokenId, config) {} catch {}
```
hensha256, line 78: messaged on slack but i think im pro keeping this

hensha256 (approved):

src/PositionManager.sol
kmbarry1, line 330: Is there some reason this can't be done after the storage writes below to follow CEI? E.g. state the subscriber might need to read?

saucepoint: on main, both modifyLiquidity + unsubscribe should adhere to CEI now

https://github.com/Uniswap/v4-periphery/blob/main/src/PositionManager.sol#L314-L327

PR #318, opened 2024-08-30
OZ-L10, SB-L53, SB-L54, SB-M78; Check code length on Notifications, Unsubscribe Minimum Gas Limit

saucepoint:
## Related Issue

OZ-L10: calls to addresses without bytecode will still hard-revert, even with try/catch

For chains that still support `selfdestruct`, you can subscribe to a valid contract but then lose access to unsubscribing if the contract self-destructs

---

SB L53: Notifications call contracts without code

SB L54: Subscribing multiple times to contracts without code

SB M78: Unsubscribe should have a Minimum Gas Limit

## Description of changes

* if subscriber has no code then `unsubscribe` will **not** revert
* Added a custom error for unsubscribing when already unsubscribed
* if subscriber has no code then notifying `subscribe, transfer, and modifyLiquidity` will **revert**
* Unsubscribe requires a minimum gas limit otherwise revert. before users can use pick a gas limit where notifyUnsubscribe reverts (and is try/catched) with OutOfGas but the overall transaction still succeeds


hensha256 (dismissed):

src/base/Notifier.sol
```diff
@@ -67,11 +67,20 @@ abstract contract Notifier is INotifier {
...
-        try _subscriber.notifyUnsubscribe{gas: subscriberGasLimit}(tokenId, config, data) {} catch {}
+        if (address(_subscriber).code.length > 0) {
+            uint256 subscriberGasLimit = block.gaslimit.calculatePortion(BLOCK_LIMIT_BPS);
+
+            // require that the remaining gas is sufficient to notify the subscriber
+            // otherwise, users can select a gas limit where .notifyUnsubscribe hits OutOfGas yet the transaction/unsubscription
+            // can still succeed
+            // to account for EIP-150, condition could be 64 * gasleft() / 63 <= subscriberGasLimit
+            if ((64 * gasleft() / 63) < subscriberGasLimit) GasLimitTooLow.selector.revertWith();
+            try _subscriber.notifyUnsubscribe{gas: subscriberGasLimit}(tokenId, config, data) {} catch {}
```
saucepoint, line 82: need extra eyes on this (and the natspec)

feels weirdly unintuitive that users may need to specify a very high gas limit on `.unsubscribe()` calls

@snreynolds @hensha256

hensha256: yeah it means that all users will have to send in `gasLimit > subscriberGasLimit`, even if their particular subsciber only needs like 10k in gas. Which sucks really... but I can't see any other way around it?
I think it also means that it will be really hard to run estimateGas because of this require statement.

src/base/Notifier.sol
```diff
@@ -67,11 +67,20 @@ abstract contract Notifier is INotifier {
...
-        uint256 subscriberGasLimit = block.gaslimit.calculatePortion(BLOCK_LIMIT_BPS);
-        try _subscriber.notifyUnsubscribe{gas: subscriberGasLimit}(tokenId, config, data) {} catch {}
+        if (address(_subscriber).code.length > 0) {
+            uint256 subscriberGasLimit = block.gaslimit.calculatePortion(BLOCK_LIMIT_BPS);
+
+            // require that the remaining gas is sufficient to notify the subscriber
+            // otherwise, users can select a gas limit where .notifyUnsubscribe hits OutOfGas yet the transaction/unsubscription
+            // can still succeed
+            // to account for EIP-150, condition could be 64 * gasleft() / 63 <= subscriberGasLimit
+            if ((64 * gasleft() / 63) < subscriberGasLimit) GasLimitTooLow.selector.revertWith();
```
hensha256, line 81: in core we dont bother with the *64/63. Theres just a comment about it
https://github.com/Uniswap/v4-core/blob/4dc48bbd89f4faf1e9493cd1563908d79aa10b0d/src/ProtocolFees.sol#L76

thoughts?

src/interfaces/INotifier.sol
```diff
@@ -33,6 +39,7 @@ interface INotifier {
     /// @param tokenId the ERC721 tokenId
     /// @param config the corresponding PositionConfig for the tokenId
     /// @param data caller-provided data that's forwarded to the subscriber contract
+    /// @dev Callers must specify a high gas limit (remaining gas should be higher than 300,000 gas) such that the subscriber can be notified
```
hensha256, line 42:
```suggestion
    /// @dev Callers must specify a high gas limit (remaining gas should be higher than subscriberGasLimit) such that the subscriber can be notified
```

hensha256: it wont be 300k on all chains, plus block gas limit can change so we dont know what it will be in future

src/base/Notifier.sol
snreynolds: agree

src/interfaces/ISubscriber.sol
```diff
@@ -10,6 +10,9 @@ interface ISubscriber {
     /// @param config details about the position
     /// @param data additional data passed in by the caller
     function notifySubscribe(uint256 tokenId, PositionConfig memory config, bytes memory data) external;
+    /// @notice Called when a position unsubscribes from the subscriber
+    /// @dev This call's gas is capped at `unsubscribeGasLimit` (set at deployment)
+    /// @dev Because of EIP-150, solidity may only allocate 63/64 of gasleft()
```
hensha256, line 15:
```suggestion
    /// @dev Because of EIP-150, solidity may only allocate 63/64 of unsubscribeGasLimit
```

src/base/Notifier.sol
```diff
@@ -65,13 +66,17 @@ abstract contract Notifier is INotifier {
     function _unsubscribe(uint256 tokenId, PositionConfig calldata config) internal {
         _positionConfigs(tokenId).setUnsubscribe();
         ISubscriber _subscriber = subscriber[tokenId];
+        if (_subscriber == NO_SUBSCRIBER) NotSubscribed.selector.revertWith();
```
hensha256, line 69: i see that this check makes sense, surprised it wasnt here before
doesnt one of these issues say to add it? i cant find it

src/base/Notifier.sol
```diff
@@ -65,13 +66,17 @@ abstract contract Notifier is INotifier {
...
         delete subscriber[tokenId];
 
-        // A gas limit and a try-catch block are used to protect users from a malicious subscriber.
-        // Users should always be able to unsubscribe, not matter how the subscriber behaves.
-        uint256 subscriberGasLimit = block.gaslimit.calculatePortion(BLOCK_LIMIT_BPS);
-        try _subscriber.notifyUnsubscribe{gas: subscriberGasLimit}(tokenId, config) {} catch {}
+        if (address(_subscriber).code.length > 0) {
+            // require that the remaining gas is sufficient to notify the subscriber
+            // otherwise, users can select a gas limit where .notifyUnsubscribe hits OutOfGas yet the transaction/unsubscription
+            // can still succeed
```
hensha256, line 76:
```suggestion
            // otherwise, users can select a gas limit where .notifyUnsubscribe hits OutOfGas yet the 
            // transaction/unsubscription can still succeed
```

hensha256: lol such a nit but that line is longgg

src/base/Notifier.sol
```diff
@@ -76,14 +75,12 @@ abstract contract Notifier is INotifier {
...
-            uint256 subscriberGasLimit = block.gaslimit.calculatePortion(BLOCK_LIMIT_BPS);
-
             // require that the remaining gas is sufficient to notify the subscriber
             // otherwise, users can select a gas limit where .notifyUnsubscribe hits OutOfGas yet the transaction/unsubscription
             // can still succeed
-            // to account for EIP-150, condition could be 64 * gasleft() / 63 <= subscriberGasLimit
-            if ((64 * gasleft() / 63) < subscriberGasLimit) GasLimitTooLow.selector.revertWith();
-            try _subscriber.notifyUnsubscribe{gas: subscriberGasLimit}(tokenId, config) {} catch {}
+            // to account for EIP-150, condition could be 64 * gasleft() / 63 <= unsubscribeGasLimit
+            if ((64 * gasleft() / 63) < unsubscribeGasLimit) GasLimitTooLow.selector.revertWith();
```
snreynolds: I think we do not need to calculate the (64 * gasLeft / 63) check. 

Really this line is just being added to enforce some minimum amount that a user MUST send. 

That minimum amount doesn't exactly need to be 63/64ths of the unsubscribeGasLimit. It can just be at least the unsubscribeGasLimit. And we can add a comment that in 1 case (when gasLeft == gasLimit) the call will only forward 63/64ths of the gasLimit... which is fine. Basically we can just keep this in mind when parameterizing the `unsubscribeGasLimit`.

We really want to add this line because we do not want the user to ever send a minimal amount of gas such that the unsubscribe block reverts but the rest of the function succeeds

src/base/Notifier.sol
```diff
@@ -4,26 +4,27 @@ pragma solidity ^0.8.0;
...
-    using BipsLibrary for uint256;
     using CustomRevert for bytes4;
     using PositionConfigIdLibrary for PositionConfigId;
 
     ISubscriber private constant NO_SUBSCRIBER = ISubscriber(address(0));
 
-    // a percentage of the block.gaslimit denoted in BPS, used as the gas limit for subscriber calls
-    // 100 bps is 1%, at 30M gas, the limit is 300K
-    uint256 private constant BLOCK_LIMIT_BPS = 100;
+    /// @inheritdoc INotifier
```
snreynolds, line 18: nit: i think there is an extra line added that can be removed.

src/base/Notifier.sol
snreynolds: I think it may make more sense to check _postionConfigs().hasSubscriber? i guess see if its cheaper in the case where they do have a subscriber?

src/interfaces/ISubscriber.sol
snreynolds: i actually think its clearer to leave it as 

`/// @dev Because of EIP-150, solidity may only allocate 63/64 of gasleft()`

because this only happens when gasLeft == unsubscriberGasLimit

snreynolds (commented):
I see that we changed it to document EIP150. I just left some comments in response to Alice's about the natspec!

Looks good!

src/base/Notifier.sol
saucepoint: this natspec? fwiw i added this to INotifier

```solidity
    /// @notice Returns and determines the maximum allowable gas-used for notifying unsubscribe
    /// @return uint256 the maximum gas limit when notifying a subscriber's `notifyUnsubscribe` function
    function unsubscribeGasLimit() external view returns (uint256);
```

src/base/Notifier.sol
saucepoint: yep looked like .hasSubscriber() was slightly cheaper! refactored it

src/interfaces/ISubscriber.sol
snreynolds: yeah agree that is the lower bound !

snreynolds (dismissed):
nice

snreynolds (approved):

hensha256 (approved):

PR #320, opened 2024-08-30
SB-L73: payable initializePool

saucepoint:
## Related Issue
SB-L73: missing payable

**NOTE:** the spearbit issue states we are missing `payable` on the ERC721 methods. However, we cannot override Solmate's functions and change from `nonpayable` to `payable`

## Description of changes

* Added `payable` to `initializePool` 
* unit test `multicall(initializePool, mint)`

### DIFF
--- a/src/base/PoolInitializer.sol
+++ b/src/base/PoolInitializer.sol
@@ -8,6 +8,7 @@ import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
 abstract contract PoolInitializer is ImmutableState {
     function initializePool(PoolKey calldata key, uint160 sqrtPriceX96, bytes calldata hookData)
         external
+        payable
         returns (int24)
     {
         return poolManager.initialize(key, sqrtPriceX96, hookData);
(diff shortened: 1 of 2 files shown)


hensha256:
>NOTE: the spearbit issue states we are missing payable on the ERC721 methods. However, we cannot override Solmate's functions and change from nonpayable to payable

interesting. Do you think thats a problem?

hensha256 (approved):

saucepoint:
Nonpayable functions from Solmate:

* approve
* setApprovalForAll
* transferFrom
* safeTransferFrom

---

the following use-cases, via multicall, would not be supported:

* `modify + transfer`: user wants to change the position before changing ownership
* `approve + modify or modify + approve`: user wants to batch two independent actions, for gas savings
* `transfer + permit + modify`: user wants to transfer ownership but still wants to retain permission

these cases feel obscure, and case 2 can use permit

snreynolds:
> Nonpayable functions from Solmate:
> 
> * approve
> * setApprovalForAll
> * transferFrom
> * safeTransferFrom
> 
> the following use-cases, via multicall, would not be supported:
> 
> * `modify + transfer`: user wants to change the position before changing ownership
> * `approve + modify or modify + approve`: user wants to batch two independent actions, for gas savings
> * `transfer + permit + modify`: user wants to transfer ownership but still wants to retain permissions
> 
> these cases feel obscure, and case 2 can use permit

I don't think we should fork to add payable on these functions. Use cases are obscure, only affect eth pools, and all of these batched actions won't be directly supported in the interface 🤷‍♀️

snreynolds (approved):

PR #321, opened 2024-08-30
Various spearbit fixes

hensha256:
Issue numbers in commit messages


.forge-snapshots/V4Router_Bytecode.snap
```diff
@@ -1 +1 @@
-8389
\ No newline at end of file
+8394
```
dianakocsis, line 1: this increases bytecode and doesn't change gas at all 🤔

dianakocsis:
Looks good, just a lot left with 62 which we can do in separate PR. I can take that one

saucepoint (dismissed):

saucepoint (approved):

PR #325, opened 2024-08-31
sb-i74

dianakocsis:
## Related Issue
Spearbit I-74 Memory Cleaning
https://github.com/spearbit-audits/review-uniswap-v4/issues/74

## Description of changes
clean the memory we used


src/base/EIP712_v4.sol
```diff
@@ -46,6 +46,11 @@ contract EIP712_v4 is IEIP712_v4 {
             mstore(add(fmp, 0x02), domainSeparator)
```
dianakocsis, line 46: @hensha256 when is it left packed vs right packed?

hensha256: bytes are left-aligned, numbers and addresses are right-aligned. So the line above this
```
      mstore(fmp, hex"1901")
```

stores 0x1901 in `[fmp..fmp+2)`
whereas if you wrote

```
      mstore(fmp, 6401)     // 6401 == 0x1901
```

it would right align it into `[fmp+30..fmp+32)`

hensha256: In fact even just `mstore(fmp, 0x1901)` would right align it because 0x1901 gets compiled as a number. You have to write `hex"1901"` for it to compile it as bytes and left-align

hensha256 (approved):

PR #332, opened 2024-09-02
consistent pragma versions

hensha256:
- 0.8.26 for V4Router and PositionManager which we will be deploying
- ^0.8.24 for files that directly need to use transient storage
- ^0.8.0 for all other files

This setup allows us to give integrators flexbility


dianakocsis (approved):

hensha256:
good catch ty

dianakocsis (approved):

PR #336, opened 2024-09-04
Remove Wrap__ notation from permit2 revert reason

saucepoint:
## Related Issue

We believe the `Wrap__` notation should only be reserved for "unknown"/"variable" external calls (subscribers, hooks, tokens, etc). So we are removing it from permit2 error handling, to be consistent with external calls to PoolManager

## Description of changes

Removed `Wrap__` notation from permit2 error handling


snreynolds (dismissed):

snreynolds:
Looks good just fix linter!

snreynolds (approved):

PR #339, opened 2024-09-05
Spearbit 99

dianakocsis:
## Related Issue
Spearbit 99 _swapExactOutput... does not compare the params.amountOut to ActionConstants.OPEN_DELTA
https://github.com/spearbit-audits/review-uniswap-v4/issues/99

## Description of changes
compare params.amountOut to ActionConstants.OPEN_DELTA


snreynolds (commented):
this looks good. 

it pains me that we don't use the mapTakeAmount and mapSettleAmount functions that we define in DeltaResolver.sol but I think its cause they added quite a bit more gas with extra casting. I wonder if we could overload those functions for uint128s now that we want to continue this logic in other places?

snreynolds (commented):
ehhh ok I think you can't overload those functions. Let's send it. 

& I'm commenting why we don't use those functions for future reference which is because they add extra gas in the router with the extra casting

snreynolds (approved):

PR #342, opened 2024-09-05
Improve calldata decoder

dianakocsis:
## Related Issue
mask when decoding

## Description of changes


snreynolds:
Think we can close this given it gets truncated on return & is unused within the assembly block

PR #349, opened 2024-09-09
Quoter update

hensha256:
- huge simplification of the return parameters - now we just return the gasEstimate, and the quoted amount in unspecified
- removed the cache, its unneeded as it can just be checked within the _swap function with the parameters available
- refactored various functions into libraries for readability
- updated how the revert is caught to be more accurate with custom errors


src/base/BaseV4Quoter.sol
```diff
@@ -0,0 +1,62 @@
...
+
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
+import {BalanceDelta} from "@uniswap/v4-core/src/types/BalanceDelta.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+import {QuoterRevert} from "../libraries/QuoterRevert.sol";
+import {SqrtPriceLimitHelper} from "../libraries/SqrtPriceLimitHelper.sol";
+import {SafeCallback} from "../base/SafeCallback.sol";
+import {PoolId, PoolIdLibrary} from "@uniswap/v4-core/src/types/PoolId.sol";
+
+abstract contract BaseV4Quoter is SafeCallback {
```
hensha256, line 11: This base contract is used so that we can inherit it into siyu's MixedQuoter (that quotes V2, V3, and V4)

src/interfaces/IQuoter.sol
```diff
@@ -74,26 +56,20 @@ interface IQuoter {
     /// exactAmount The desired output amount
     /// sqrtPriceLimitX96 The price limit of the pool that cannot be exceeded by the swap
     /// hookData arbitrary hookData to pass into the associated hooks
-    /// @return deltaAmounts Delta amounts resulted from the swap
-    /// @return sqrtPriceX96After The sqrt price of the pool after the swap
-    /// @return initializedTicksLoaded The number of initialized ticks that the swap loaded
+    /// @return amountIn The input quote for the exactOut swap
+    /// @return gasEstimate Estimated gas units used for the swap
     function quoteExactOutputSingle(QuoteExactSingleParams memory params)
```
snreynolds, line 61: does it make sense to add batched functions so you can get many quotes in one call?

src/libraries/QuoterRevert.sol
```diff
@@ -0,0 +1,55 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import {ParseBytes} from "@uniswap/v4-core/src/libraries/ParseBytes.sol";
+
+library QuoterRevert {
```
snreynolds, line 6: I find this library a little difficult to reason about because all these functions seem to be doing some of the same things but at various lifecycles of a quote (various parsing, and various reverting) but its unclear when to use what.

some suggestions??

- does `validateRevertReason` need to be its own function? Seems like its just used in this file. Can you collapse it?
- for all of the reverting code paths can we decide if we want to use the revert logic in assembly or not? `revertQuote` and `bubbleReason` are not consistent with the reverting info.
- can you add documentation for whats the difference between those two functions? its unclear.. it seems like `revertQuote` could just use `bubbleReason`?
- do we need `parseAmountUnspecified as its own function as well? it seems like its just used in parseReturnData? Maybe we can change parseReturnData -> parseAmountUnspecified to be more clear since it doesnt parse any other kind of data?

src/base/BaseV4Quoter.sol
```diff
@@ -0,0 +1,62 @@
...
+    modifier selfOnly() {
+        if (msg.sender != address(this)) revert NotSelf();
+        _;
+    }
+
+    function _unlockCallback(bytes calldata data) internal override returns (bytes memory) {
+        // Call this contract with the data in question. Each quote path
+        (bool success, bytes memory returnData) = address(this).call(data);
+        if (success) return returnData;
+        if (returnData.length == 0) revert LockFailure();
```
snreynolds, line 33: not sure I understand why this is necessarily a lock failure?

test/Quoter.t.sol
```diff
@@ -480,33 +362,19 @@ contract QuoterTest is Test, Deployers, GasSnapshot {
...
-            uint160[] memory sqrtPriceX96AfterList,
-            uint32[] memory initializedTicksLoadedList
-        ) = quoter.quoteExactOutput(params);
+        (uint256 amountIn,) = quoter.quoteExactOutput(params);
 
-        assertEq(deltaAmounts[0], 6283);
-        assertEq(sqrtPriceX96AfterList[0], 79708304437530892332449657932);
-        assertEq(initializedTicksLoadedList.length, 1);
-        assertEq(initializedTicksLoadedList[0], 2);
+        assertEq(amountIn, 6283);
```
snreynolds, line 367: assert gas is > 0? I know it's hard to get accurate on the gas but we know it should at least not be 0

src/interfaces/IQuoter.sol
hensha256: they use an external multicall contract for multicalling quoters onchain 
on mainnet its [here](https://etherscan.io/address/0x1F98415757620B543A52E61c46B32eB19261F984#code)

test/Quoter.t.sol
hensha256: I've done asserts `50k < gas < 400k`

src/interfaces/IQuoter.sol
snreynolds: ah yeah cool

src/base/BaseV4Quoter.sol
```diff
@@ -0,0 +1,62 @@
...
+    /// external call to catch and parse revert reasons
+    modifier selfOnly() {
+        if (msg.sender != address(this)) revert NotSelf();
+        _;
+    }
+
+    function _unlockCallback(bytes calldata data) internal override returns (bytes memory) {
+        (bool success, bytes memory returnData) = address(this).call(data);
+        // Every quote path gathers a quote, and then reverts either with QuoteSwap(quoteAmount) or alternative error
+        if (success) revert UnexpectedCallSuccess();
```
snreynolds, line 32: Just starting to look through this but my first thought here.. is I'm wondering if its even possible that this returns true? Like could you write a test that hits this revert?

src/libraries/QuoterRevert.sol
```diff
@@ -28,28 +30,20 @@ library QuoterRevert {
...
-        reason.validateRevertReason();
-        return reason.parseAmountUnspecified();
-    }
 
-    /// @notice extracts the amountUnspecified from an encoded QuoteSwap(amountUnspecified)
-    function parseAmountUnspecified(bytes memory reason) internal pure returns (uint256 amountUnspecified) {
         // reason -> reason+0x1f is the length of the reason string
         // reason+0x20 -> reason+0x23 is the selector of QuoteSwap
-        // reason+0x24 -> reason+0x43 is the amountUnspecified
+        // reason+0x24 -> reason+0x43 is the quoteAmount
```
snreynolds, line 44: I guess this is more of a comment to all of this natspec but this is kinda confusing.

0x20 -> 0x23 is a 3 byte difference?
0x24 -> 0x43 is a 31 byte difference? 

Is something not inclusive?

src/interfaces/IQuoter.sol
```diff
@@ -74,26 +56,20 @@ interface IQuoter {
...
+    /// @return amountIn The input quote for the exactOut swap
+    /// @return gasEstimate Estimated gas units used for the swap
     function quoteExactOutput(QuoteExactParams memory params)
         external
-        returns (
-            int128[] memory deltaAmounts,
-            uint160[] memory sqrtPriceX96AfterList,
-            uint32[] memory initializedTicksLoadedList
-        );
+        returns (uint256 amountIn, uint256 gasEstimate);
```
snreynolds, line 74: Ok I think we should actually return the BalanceDelta, will follow up async w you

src/libraries/QuoterRevert.sol
```diff
@@ -0,0 +1,49 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+import {ParseBytes} from "@uniswap/v4-core/src/libraries/ParseBytes.sol";
+
+library QuoterRevert {
```
snreynolds, line 6: Sweet this looks great!

snreynolds:
Looks like gas snapshots just need to be updated

src/libraries/QuoterRevert.sol
hensha256: it is inclusive thats why... 0x20, 0x21, 0x22, 0x23 is 4 bytes inclusively

src/base/BaseV4Quoter.sol
hensha256: yeah I dont think so - but its such a tiny check for a lens contract to do that feels worth catching?

marktoda (approved):

PR #350, opened 2024-09-11
add updated deploys

snreynolds:
## Related Issue

## Description of changes


broadcast/01_PoolManager.s.sol/11155111/run-latest.json
```diff
@@ -1,20 +1,18 @@
 {
   "transactions": [
     {
-      "hash": "0xae8b5df5aeaac2faf35a24f27312c4defd3eefbca9243f504e68621e5fae551c",
+      "hash": "0x4f7feefa3bf81d750076ebb545a632b986548e14b634bc20ae9efc531830299e",
       "transactionType": "CREATE",
       "contractName": "PoolManager",
-      "contractAddress": "0xf242ce588b030d0895c51c0730f2368680f80644",
+      "contractAddress": "0xe8e23e97fa135823143d6b9cba9c699040d51f70",
```
hensha256, line 7: i feel like ive said this every time... but imo we should never be pushing `run-latest` to github as its a file that gets updated regularly. It would be better to put `run-timestamp` to the repo so we can keep track of all deployments through time

snreynolds: See this PR https://github.com/Uniswap/v4-periphery/pull/293

snreynolds: Also you can just look at the github history if you rlly wanna know...

script/DeployPosm.s.sol
```diff
@@ -18,7 +18,7 @@ contract DeployPosmTest is Script {
     {
         vm.startBroadcast();
 
-        posm = new PositionManager{salt: hex"03"}(
+        posm = new PositionManager{salt: hex"01"}(
```
hensha256, line 21: why not just use create1 to stop needing to change the salt every time?

hensha256 (approved):

script/DeployPosm.s.sol
snreynolds: We actually usually don't need to change the salts each time bc the bytecode is so different from deploy to deploy. On posm I started using salt when I was testing to see how different ascii look on etherscan 😄

snreynolds: I dont wanna change that though bc we've deployed the current deploys with create2 so nice to have a record in this PR what salt it was deployed with

broadcast/01_PoolManager.s.sol/11155111/run-latest.json
neelansh15: Hey all. Would really appreciate having timestamped `run-latest`. I'm developing Uniswap hooks during a hackathon and the deployments in general seem all over the place sometimes haha. Just an input from my side, cheers!

snreynolds: @neelansh15 Thats great feedback! We will soon have a better and more consistent record for deploys

PR #365, opened 2024-10-15
Remove sqrtpricelimit from v4 router

hensha256:


src/V4Router.sol
```diff
@@ -170,22 +158,15 @@ abstract contract V4Router is IV4Router, BaseActionsRouter, DeltaResolver {
...
         unchecked {
             BalanceDelta delta = poolManager.swap(
                 poolKey,
                 IPoolManager.SwapParams(
-                    zeroForOne,
-                    amountSpecified,
-                    sqrtPriceLimitX96 == 0
-                        ? (zeroForOne ? TickMath.MIN_SQRT_PRICE + 1 : TickMath.MAX_SQRT_PRICE - 1)
-                        : sqrtPriceLimitX96
+                    zeroForOne, amountSpecified, zeroForOne ? TickMath.MIN_SQRT_PRICE + 1 : TickMath.MAX_SQRT_PRICE - 1
```
snreynolds, line 169: Could we leave a comment somewhere explaining motivation for this? Its an edge case for sure, but maybe like "for protection of end users, this contract automatically sets a maximum value for the sqrtPriceLimit, and slippage checks are relied on to enforce proper amount out checks" or something..

snreynolds (commented):
Clean.

Would just like to document in this PR though that IF a pool's price is initialized to an undesired price and there is no liquidity, then setting the `sqrtPriceLimitX96` can be used to push the pool's price to a desired value. With this change, someone would need to do that outside of this router.

snreynolds (approved):

PR #373, opened 2024-10-17
Pashov L-04 display Native instead of the zero address

dianakocsis:
## Related Issue
Pashov L-04 Using zero address for native token can be confusing for end users

## Description of changes
display Native instead of the zero address


src/libraries/SafeERC20Metadata.sol
```diff
@@ -2,41 +2,36 @@
...
 import {AddressStringUtil} from "./AddressStringUtil.sol";
 
-/// @title SafeCurrencyMetadata
+/// @title SafeERC20Metadata
 /// @notice can produce symbols and decimals from inconsistent or absent ERC20 implementations
 /// @dev Reference: https://github.com/Uniswap/solidity-lib/blob/master/contracts/libraries/SafeERC20Namer.sol
-library SafeCurrencyMetadata {
-    using CurrencyLibrary for Currency;
-
+library SafeERC20Metadata {
```
hensha256, line 10: why did you rename this back? it was currency because it also handles native right? why is it ERC20 now?

src/libraries/Descriptor.sol
```diff
@@ -45,15 +44,15 @@ library Descriptor {
...
             addressToString(params.poolManager)
         );
         string memory descriptionPartTwo = generateDescriptionPartTwo(
             params.tokenId.toString(),
-            escapeQuotes(params.baseCurrencySymbol),
-            addressToString(Currency.unwrap(params.quoteCurrency)),
-            addressToString(Currency.unwrap(params.baseCurrency)),
+            escapeQuotes(params.baseAddressSymbol),
+            params.quoteAddress == address(0) ? "Native" : addressToString(params.quoteAddress),
+            params.baseAddress == address(0) ? "Native" : addressToString(params.baseAddress),
```
hensha256, line 55: Is there Native tests?

src/libraries/SafeERC20Metadata.sol
dianakocsis: mm i renamed it back since we're not unwrapping the currency in the file anymore, its just working with address

hensha256: mmm but its not only working with ERC20s either...? its still working with native e.g.
```
        if (addr == address(0)) {
            return nativeLabel;
        }
```

i'd personally still rather the name "currency"? its still working with both native and ERC20s even if its not using the Currency type

hensha256: `AddressRatioSortOrder` is fine.. could you `address` if you prefer. But ERC20 is still wrong imo

test/PositionDescriptor.t.sol
```diff
@@ -122,7 +123,54 @@ contract PositionDescriptorTest is Test, PosmTestSetup, GasSnapshot {
...
+        string memory json = string(decoded);
+
+        // decode json
+        bytes memory data = vm.parseJson(json);
+        Token memory token = abi.decode(data, (Token));
+
+        assertEq(token.name, "Uniswap - 0.3% - TEST/ETH - 1.0060<>1.0121");
+        assertEq(
+            token.description,
+            unicode"This NFT represents a liquidity position in a Uniswap v4 TEST-ETH pool. The owner of this NFT can modify or redeem the position.\n\nPool Manager Address: 0x5615deb798bb3e4dfa0139dfa1b3d433cc23b72f\nTEST Address: 0x5991a2df15a8f6a256d3ec51e99254cd3fb576a9\nETH Address: Native\nHook Address: 0x0000000000000000000000000000000000000000\nFee Tier: 0.3%\nToken ID: 1\n\n⚠️ DISCLAIMER: Due diligence is imperative when assessing this NFT. Make sure addresses match the expected addresses, as symbols may be imitated."
```
hensha256, line 173: This should say "No Hook" right? Maybe you didnt merge main

dianakocsis: yeah i didnt merge main!

hensha256 (approved):

PR #374, opened 2024-10-17
Pashov L-05 add more special characters to escape

dianakocsis:
## Related Issue
Pashov L-05 https://github.com/PashovAuditGroup/Uniswap_October_MERGED/issues/9
Special characters not escaped can produce invalid JSON

## Description of changes
check form feeds, new lines, carriage returns, or tabs to escape


src/libraries/Descriptor.sol
```diff
@@ -503,6 +503,10 @@ library Descriptor {
         }
     }
 
+    function isSpecialCharacter(bytes1 b) private pure returns (bool) {
+        return b == '"' || b == "\u000c" || b == "\n" || b == "\r" || b == "\t";
+    }
```
djb15, line 508: These new strings are now multiple bytes long so I don't think this byte by byte parsing will match

src/libraries/Descriptor.sol
```diff
@@ -503,6 +503,10 @@ library Descriptor {
         }
     }
 
+    function isSpecialCharacter(bytes1 b) private pure returns (bool) {
+        return b == '"' || b == "\u000c" || b == "\n" || b == "\r" || b == "\t";
```
hensha256, line 507: am i mad? surely `\u000c` is longer than a single byte?

djb15: I had exactly the same thought! Turns out it get's properly escaped into 1 bytes:

```
pragma solidity ^0.8.0;

contract RandomTest {
    function stringToBytes32() public pure returns (uint256 length) {
        string memory source = "\u000c";
        bytes memory tempEmptyStringTest = bytes(source);
        return tempEmptyStringTest.length;
    }
}
```

If you test this out in Remix the return value is 1 :)

hensha256: wild, thanks for the poc :)

hensha256 (dismissed):

hensha256 (approved):

PR #379, opened 2024-10-18
check pool manager lock on transfer, subscribe, and unsubscribe

snreynolds:
## Related Issue

## Description of changes


src/base/Notifier.sol
```diff
@@ -29,6 +29,10 @@ abstract contract Notifier is INotifier {
     /// @param tokenId the tokenId of the position
     modifier onlyIfApproved(address caller, uint256 tokenId) virtual;
 
+    /// @notice Enforces that the PoolManager is locked.
+    /// @dev Reverts if the caller tries to transfer, subscribe, or unsubscribe the position while the PoolManager is unlocked.
```
hensha256, line 33: nit but i feel like this comment shouldnt say which functions the modifier is on? The modifier's definition is separate to the fact that it is included on certain functions

hensha256: I'd rather it was added as a comment to the natspec for subscribe, unsubscribe, and transfer, to say we check that

hensha256: Same on PositionManager.sol
Also the function is `transferFrom` not `transfer`

hensha256 (approved):

Issue #380, opened 2024-10-20
[Bug]: Lint error "explicitly mark visibility of state" in Actions.sol


three9s:
### Describe the bug

In Solidity, the visibly keyword needs to be specified.

### Expected Behavior

Specify `internal`, `private`, or `public`. It defaults to `internal` if not specified, so that is correct.

PR #381, opened 2024-10-20
fix: explicitly mark visibility of state for actions constants

three9s:
## Related Issue
#380

## Description of changes

In Solidity, the visibly keyword needs to be specified. It defaults to `internal` if not specified, so that is correct to make this a no-op change.


snreynolds:
Im happy to have this merged if you would fix the merge conflicts

three9s:
@snreynolds I fixed the conflicts, it is good to merge.

snreynolds (approved):

PR #387, opened 2024-10-30
feat: update v4 core

marktoda:
this commit updates v4 core to the latest main and updates custom revert handling to support the latest bubble up

## Related Issue

## Description of changes


remappings.txt
```diff
@@ -1,6 +1,6 @@
 @uniswap/v4-core/=lib/v4-core/
-forge-gas-snapshot/=lib/v4-core/lib/forge-gas-snapshot/src/
```
marktoda, line 2: installing directly here since it was removed from core. Will update to native snaps in separate PR

src/base/Notifier.sol
```diff
@@ -9,6 +9,7 @@ import {PositionInfo} from "../libraries/PositionInfoLibrary.sol";
 
 /// @notice Notifier is used to opt in to sending updates to external contracts about position modifications or transfers
 abstract contract Notifier is INotifier {
+    using CustomRevert for address;
     using CustomRevert for bytes4;
```
hensha256, line 13: I'd personally just do 
```
    using CustomRevert for *;
```

instead of 2 different lines

gretzke: or only keep `using CustomRevert for bytes4` and call `bubbleUpAndRevertWith` as a normal function

hensha256 (dismissed):

hensha256 (approved):

PR #395, opened 2024-11-21
Unsubscribe on transfer

hensha256:
Remove notifyTransfer, and instead unsubscribe on transfer


test/position-managers/PositionManager.notifier.t.sol
```diff
@@ -554,75 +566,6 @@ contract PositionManagerNotifierTest is Test, PosmTestSetup, GasSnapshot {
         lpm.modifyLiquidities(calls, _deadline);
     }
 
-    function test_notifyTransfer_withTransferFrom_wraps_revert() public {
```
hensha256, line 557: unsubscribe cannot revert, so these tests are no longer relevant

snreynolds (approved):
lgtm

PR #397, opened 2024-11-22
try-catch-initialize

snreynolds:
## Related Issue

## Description of changes


hensha256 (dismissed):

src/base/PoolInitializer.sol
```diff
@@ -12,7 +12,8 @@ abstract contract PoolInitializer is ImmutableState {
     /// @notice Initialize a Uniswap v4 Pool
     /// @param key the PoolKey of the pool to initialize
     /// @param sqrtPriceX96 the initial sqrtPriceX96 of the pool
+    /// @return tick The current tick of the pool
     function initializePool(PoolKey calldata key, uint160 sqrtPriceX96) external payable returns (int24) {
-        return poolManager.initialize(key, sqrtPriceX96);
+        try poolManager.initialize(key, sqrtPriceX96) returns (int24) {} catch {}
```
hensha256: i think if you want to return the tick (if we care) then you have to do like
```
        try poolManager.initialize(key, sqrtPriceX96) returns (int24 tick) {
            return tick;
        } catch {
            return 0;
        }
```

i think it currently never returns something

hensha256: or you could remove the return parameter

hensha256: or you could name the return parameter of `initializePool` to `tick` and then do
```
        try poolManager.initialize(key, sqrtPriceX96) returns (int24 _tick) {
            tick = _tick;
        } catch {}
```

hensha256: happy with any of those 3 solutions :)

hensha256: hm actually you dont want to return `0` on failure because thats a valid tick too 😆 i'm pushing a suggestion

hensha256 (dismissed):

hensha256 (approved):

hensha256 (approved):

PR #399, opened 2024-11-22
use notifyBurn

snreynolds:
## Related Issue

## Description of changes


src/base/Notifier.sol
```diff
@@ -88,6 +88,26 @@ abstract contract Notifier is INotifier {
...
     }
 
+    /// @dev note this function also deletes the subscriber address from the mapping
+    function _notifyBurn(uint256 tokenId, address owner, PositionInfo info, uint256 liquidity, BalanceDelta feesAccrued)
+        internal
+    {
+        ISubscriber _subscriber = subscriber[tokenId];
+
+        // remove the subscriber
+        delete subscriber[tokenId];
```
snreynolds, line 98: i don't love removing this here. also it feels like something we should do before any modifyLiquidity happens at all, but then notifyBurn would have to be called before modifyLiquidity which is weird..

saucepoint: if you're implying we're still violating CEI -- it might be okay since previously we were calling `_unsubscribe` (with a state change) after modifyLiquidity (interaction)

essentially were writing a special cased unsubscribe which receives additional data

saucepoint (dismissed):
im happy with this -- you are right that it feels weird to have two different places call `delete subscriber[tokenId];`

we could use an internal function but thats only marginally better

src/base/Notifier.sol
hensha256: mmm yeah i think a couple of things here
1. if its going to be in here i think the function should be called `_notifyBurnAndDeleteSubscriber` or something
2. the CEI breaking ... I cant think of any problems. But worth flagging for the audit

PR #404, opened 2024-11-28
remove stale transfer error

snreynolds:
## Related Issue

## Description of changes


.forge-snapshots/positionDescriptor bytecode size.snap
```diff
@@ -1 +1 @@
-24177
\ No newline at end of file
+24108
```
hensha256, line 1: this gas snapshot is all over the place 😓

hensha256 (approved):

.forge-snapshots/positionDescriptor bytecode size.snap
snreynolds: Yeah.. I've been noticing that. Unsure why

PR #409, opened 2024-12-02
actions clarification natspec

snreynolds:
## Related Issue

## Description of changes

### DIFF
--- a/src/libraries/Actions.sol
+++ b/src/libraries/Actions.sol
@@ -3,6 +3,7 @@ pragma solidity ^0.8.0;
 
 /// @notice Library to define different pool actions.
 /// @dev These are suggested common commands, however additional commands should be defined as required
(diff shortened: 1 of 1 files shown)


src/libraries/Actions.sol
```diff
@@ -3,6 +3,7 @@ pragma solidity ^0.8.0;
 
 /// @notice Library to define different pool actions.
 /// @dev These are suggested common commands, however additional commands should be defined as required
+/// Some of these actions are not actually supposed in the Router contracts or Position Manager contracts, but are left as they may be helpful commands for other peripheral contracts.
```
hensha256, line 6:
```suggestion
/// Some of these actions are not supported in the Router contracts or Position Manager contracts, but are left as they may be helpful commands for other peripheral contracts.
```

hensha256 (approved):

Issue #66, opened 2023-10-14
[Bug]: TWAMM SinglePool LiquidityNet update formula is incorrect


xben12:
### Describe the bug
(the whole code description [here](https://medium.com/@xben12/defi-decode-an-inside-out-explanation-of-uniswap-v4-hook-with-twamm-as-example-e9e962d95d7d)) 

In function _advanceTimestampForSinglePoolSell, the liquidity is updated as:
                params.pool.liquidity = params.zeroForOne
                    ? params.pool.liquidity - uint128(liquidityNetAtTick)
                    : params.pool.liquidity + uint128(-liquidityNetAtTick);

The above is only correct for cases where zeroForOne == true && liquidityNetAtTick >0. It is incorrect for other cases. 

For example,  
Wrong case 1: assume 
1) swap zeroForOne = false, & price go up/right-side, 
2) cross from a lower to higher liquidity zone, (liquidityNetAtTick is positive), 
the correct computation is: liquidity = liquidity + liquidityNetAtTick. 

Wrong case 2:
zeroForOne == false && liquidityNetAtTick < 0. 

in short, the correct formula also depends on liquidityNetAtTick>0 or not. 

### Expected Behavior

![liquidity_net_update](https://github.com/Uniswap/v4-periphery/assets/68003841/83911d59-7c98-4cca-8e9c-ac3bd24157c5)

The correct way has for cases, shown in the picture attached. 

we can bring down the 4 into 2 cases using 
1) get liquidityNet Absolute value
2) check whether zeroForOne == (liquidityNetAtTick>0) and update as below: 

                uint128 liqudityNetAbs =
                    liquidityNetAtTick > 0 ? uint128(liquidityNetAtTick) : uint128(-liquidityNetAtTick);
                params.pool.liquidity = params.zeroForOne == (liquidityNetAtTick > 0)
                    ? params.pool.liquidity - liqudityNetAbs
                    : params.pool.liquidity + liqudityNetAbs;

### To Reproduce
Have Added 3 test functions. the below --match-test will reproduce. and the new code solves it. 

forge test --match-test testTWAMM_OrderFill_SingleSide -vv          
[⠒] Compiling...
No files changed, compilation skipped

Running 3 tests for test/TWAMM.t.sol:TWAMMTest
[PASS] testTWAMM_OrderFill_SingleSide_OneForZero_Crosstick() (gas: 567406)
[...]

PR #410, opened 2024-12-05
TWAMM: fix single sell liquidity calc

ewilz:
## Related Issue
fixes https://github.com/Uniswap/v4-periphery/issues/66
## Description of changes


ewilz:
FullRange gas updates are from before any TWAMM upgrades and the build doesn't catch gas, so these are old gas metrics that needed to be updated.

PR #416, opened 2024-12-12
feat: compile PositionManager without `IR`

abdulla-cb:
## Related Issue

## Description of changes
- Update the `PositionManager` to avoid stack-too-deep issues

The changes to the number of optimizer runs is necessary to keep the PositionManager below the contract size limit.


saucepoint:
replaced by #421

PR #417, opened 2024-12-12
add code hash snap

snreynolds:
## Related Issue

## Description of changes


test/position-managers/PositionManager.t.sol
```diff
@@ -61,6 +61,13 @@ contract PositionManagerTest is Test, PosmTestSetup, LiquidityFuzzers {
         approvePosmFor(alice);
     }
 
+    function test_posm_initcodeHash() public {
+        vm.snapshotValue(
+            "position manager initcode hash (without constructor params, as uint256)",
+            uint256(keccak256(abi.encodePacked(vm.getCode("PositionManager.sol:PositionManager"))))
```
hensha256, line 67: can you add one for v4router too please?

saucepoint: i think this function should be in `PositionManager.gas.t.sol`

its more consistent with how we're doing it in core, and will add the snapshot to `PosMGasTest.json`

saucepoint (approved):

PR #428, opened 2025-01-11
Update BaseHook.sol

sechra:
## Related Issue

## Description of changes


hensha256:
Each slot in the EVM is 32 bytes, for that reason 32 must be added. If you truly believe it must be 34, please open a pull request explaining why the current solution is wrong, and providing a test that fails when it is not changed.

PR #429, opened 2025-01-13
Update DeltaResolver.sol

sechra:
## Related Issue

## Description of changes


saucepoint:
hey were unlikely to accept this change, as we believe the bytecode to be frozen - however i'm curious to hear your thoughts on why you think we should make this change

snreynolds:
this code is frozen so we will be closing this!

PR #434, opened 2025-01-23
HookMiner

saucepoint:
A minimal library for mining hook addresses in `forge test` or `forge script`


src/libraries/HookMiner.sol
```diff
@@ -0,0 +1,54 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.21;
+
+/// @title HookMiner
+/// @notice a minimal library for mining hook addresses
+library HookMiner {
+    // mask to slice out the bottom 14 bit of the address
+    uint160 constant FLAG_MASK = 0x3FFF; // 0000 ... 0000 0011 1111 1111 1111
```
marktoda, line 8: can just use https://github.com/Uniswap/v4-core/blob/e50237c43811bd9b526eff40f26772152a42daba/src/libraries/Hooks.sol#L26

src/libraries/HookMiner.sol
```diff
@@ -0,0 +1,54 @@
...
+
+    // Maximum number of iterations to find a salt, avoid infinite loops or MemoryOOG
```
marktoda, line 10: add how this is derived?

saucepoint: added a note, but its an arbitrary number. i can make it less "v4 coded" if 160_444 appears to be a magic number

snreynolds (commented):
I actually found the Deploy script that uses hook miner to be quite useful as well, can we PR that in as well?

test/libraries/HookMiner.t.sol
```diff
@@ -0,0 +1,59 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.24;
+
+import {Test} from "forge-std/Test.sol";
+import {Hooks} from "@uniswap/v4-core/src/libraries/Hooks.sol";
+import {HookMiner} from "../../src/libraries/HookMiner.sol";
+
+contract Blank {
```
snreynolds, line 8: actually its probably useful to test that the BaseHook validation passes with this Hook miner logic (where the constructor checks the address/and permissions) of the contract

snreynolds (approved):

PR #438, opened 2025-02-03
Move BaseHook to new utils directory

gretzke:
## Related Issue

## Description of changes

### DIFF
--- a/src/base/ImmutableState.sol
+++ b/src/base/ImmutableState.sol
@@ -10,6 +10,15 @@ contract ImmutableState is IImmutableState {
     /// @inheritdoc IImmutableState
     IPoolManager public immutable poolManager;
 
+    /// @notice Thrown when calling unlockCallback where the caller is not PoolManager
+    error NotPoolManager();
+
+    /// @notice Only allow calls from the PoolManager contract
+    modifier onlyPoolManager() {
+        if (msg.sender != address(poolManager)) revert NotPoolManager();
+        _;
+    }
+
     constructor(IPoolManager _poolManager) {
         poolManager = _poolManager;
     }
--- a/src/base/ImmutableState.sol
+++ b/src/base/ImmutableState.sol
@@ -10,7 +10,7 @@ contract ImmutableState is IImmutableState {
     /// @inheritdoc IImmutableState
     IPoolManager public immutable poolManager;
 
-    /// @notice Thrown when calling unlockCallback where the caller is not PoolManager
+    /// @notice Thrown when the caller is not PoolManager
     error NotPoolManager();
 
     /// @notice Only allow calls from the PoolManager contract
--- a/src/utils/BaseHook.sol
+++ b/src/utils/BaseHook.sol
@@ -16,18 +16,6 @@ abstract contract BaseHook is IHooks, ImmutableState {
     error InvalidPool();
     error HookNotImplemented();
 
-    /// @dev Only this address may call this function
-    modifier selfOnly() {
-        if (msg.sender != address(this)) revert NotSelf();
-        _;
-    }
-
-    /// @dev Only pools with hooks set to this contract may call this function
-    modifier onlyValidPools(IHooks hooks) {
-        if (hooks != this) revert InvalidPool();
-        _;
-    }
-
     constructor(IPoolManager _manager) ImmutableState(_manager) {
         validateHookAddress(this);
     }
--- a/src/utils/BaseHook.sol
+++ b/src/utils/BaseHook.sol
@@ -7,15 +7,16 @@ import {IHooks} from "@uniswap/v4-core/src/interfaces/IHooks.sol";
 import {BalanceDelta} from "@uniswap/v4-core/src/types/BalanceDelta.sol";
 import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
 import {BeforeSwapDelta} from "@uniswap/v4-core/src/types/BeforeSwapDelta.sol";
+import {ImmutableState} from "../base/ImmutableState.sol";
 
 /// @title Base Hook
 /// @notice abstract contract for hook implementations
-abstract contract BaseHook is IHooks {
+abstract contract BaseHook is IHooks, ImmutableState {
     error NotSelf();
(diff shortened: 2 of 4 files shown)


saucepoint (dismissed):
in favor of the change,

BUT in this tidy-up PR, it makes sense to change `BaseHook` to inherit `ImmutableState` instead

https://github.com/Uniswap/v4-periphery/pull/436#discussion_r1938529121

gretzke:
Only `SafeCallback` relied on the pool manager variable / modifier, we can remove `ImmutableState` entirely

saucepoint:
@gretzke we should decorate the hook functions with a `onlyPoolManager` modifier, so we should inherit ImmutableState i think

gretzke:
If someone overrides the hook function they also override the function modifier, because the functions are reverting anyways right now we could not have it there. If we want to ensure the modifier is not overwritten accidentally, we should split the functions into two:

```diff
-function beforeInitialize(address, PoolKey calldata, uint160) external virtual returns (bytes4) {
-    revert HookNotImplemented();
-}

+function beforeInitialize(address sender, PoolKey calldata key, uint160 sqrtPriceX96) external returns (bytes4) {
+    return _beforeInitialize(sender, key, sqrtPriceX96);
+}
+
+function _beforeInitialize(address sender, PoolKey calldata key, uint160 sqrtPriceX96) internal virtual returns (bytes4)
+    revert HookNotImplemented();
+}
```
This ensures that the external function cannot be overwritten and is always protected

gretzke:
do we still need the `selfOnly` and `onlyValidPools` modifiers?

hensha256:
I'm pro moving the modifier to ImmutableState... 👀 @saucepoint interested in your reason that we cant? just because of code freeze?

And yeah I think I'm pro adding the modifier and then adding internal functions to baseHook.sol. We follow that same pattern in SafeCallback too.

saucepoint:
> just because of code freeze?

yeah, if were not careful -- we might end up deploying a version of posm with different bytecode than whats currently in the wild

gretzke:
bytecode for positionmanager did not change after my implementation, feel free to double check

src/base/ImmutableState.sol
```diff
@@ -10,6 +10,15 @@ contract ImmutableState is IImmutableState {
...
     IPoolManager public immutable poolManager;
 
+    /// @notice Thrown when calling unlockCallback where the caller is not PoolManager
+    error NotPoolManager();
+
+    /// @notice Only allow calls from the PoolManager contract
+    modifier onlyPoolManager() {
+        if (msg.sender != address(poolManager)) revert NotPoolManager();
+        _;
+    }
```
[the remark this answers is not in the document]
gretzke: This is following the order of layout recommendations
https://docs.soliditylang.org/en/latest/style-guide.html#order-of-layout
https://docs.soliditylang.org/en/latest/style-guide.html#order-of-functions

constructor should be placed on the top of all functions, and functions come after errors and modifiers

src/utils/BaseHook.sol
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /// @dev Only this address may call this function
+    modifier selfOnly() {
+        if (msg.sender != address(this)) revert NotSelf();
+        _;
+    }
+
+    /// @dev Only pools with hooks set to this contract may call this function
+    modifier onlyValidPools(IHooks hooks) {
```
dianakocsis, line 30: when is this used?

src/utils/BaseHook.sol
```diff
@@ -0,0 +1,230 @@
...
+    error NotSelf();
+    error InvalidPool();
+    error HookNotImplemented();
+
+    constructor(IPoolManager _manager) ImmutableState(_manager) {
+        validateHookAddress(this);
+    }
+
+    /// @dev Only this address may call this function
+    modifier selfOnly() {
```
dianakocsis, line 24: or this? are these needed?

src/utils/BaseHook.sol
```diff
@@ -0,0 +1,230 @@
...
+import {ImmutableState} from "../base/ImmutableState.sol";
+
+/// @title Base Hook
+/// @notice abstract contract for hook implementations
+abstract contract BaseHook is IHooks, ImmutableState {
+    error NotSelf();
+    error InvalidPool();
+    error HookNotImplemented();
+
+    constructor(IPoolManager _manager) ImmutableState(_manager) {
```
dianakocsis, line 19: we can move the constructor below the modifier section like ImmutableState

src/utils/BaseHook.sol
dianakocsis: could these just be implemented by the hooks that extend from this contract themselves?

src/base/ImmutableState.sol
```diff
@@ -10,6 +10,15 @@ contract ImmutableState is IImmutableState {
     /// @inheritdoc IImmutableState
     IPoolManager public immutable poolManager;
 
+    /// @notice Thrown when calling unlockCallback where the caller is not PoolManager
```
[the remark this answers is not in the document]
gretzke: pushed a change here, in the future, could you suggest a change instead? e.g.,
```suggestion
    /// @notice Thrown when the caller is not PoolManager
```

src/utils/BaseHook.sol
gretzke: removed unused modifiers

src/base/ImmutableState.sol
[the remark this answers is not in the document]
dianakocsis: ah forgot again! my bad!

arora-anmol:
@hensha256 @dianakocsis @gretzke with this change, might need to update the V4 hook docs here -- https://docs.uniswap.org/contracts/v4/guides/hooks/your-first-hook as it uses `base/hooks/BaseHook` for import. 

Also if you guys could tag the commit that works with `v4-core@v4.0.0`, that could help keep track of dependencies

Hugoo:
Note, the `README.md` was not updated to reflect the directory change, here is a fix: https://github.com/Uniswap/v4-periphery/pull/450

PR #441, opened 2025-02-06
fix: npm deploy

mr-uniswap:
Removes yarn from the npm deploy

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -42,7 +42,6 @@ jobs:
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -32,7 +32,6 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@1a4442cacd436585916779262731d5b162bc6ec7
         with:
-          cache: yarn
```
hensha256, line 35: dont you need to get rid of  `yarn install --immutable --immutable-cache` too?

hensha256 (approved):

PR #443, opened 2025-02-10
add npm ignore

hensha256:
Identical to `.gitignore` minus `foundry-out`


.npmignore
```diff
@@ -0,0 +1,4 @@
+cache/
+.vscode/
+broadcast/*/*/dry-run/*.json
+broadcast/*/*/run-[0-9]*.json
```
dianakocsis, line 4: can you add a newline here?

dianakocsis: also do u need to add foundry-out/?

dianakocsis (approved):

PR #451, opened 2025-02-19
feat: add Wrapped Staked Eth hook

marktoda:
## Related Issue
Which issue does this pull request resolve?

## Description of changes

PR #452, opened 2025-02-25
feat: add ERC4626 Tokenized Vault hook

scorpion9979:
## Related Issue

## Description of changes

Adds a hook for wrapping/unwrapping assets to/from an ERC4626 vault, inspired by https://github.com/Uniswap/v4-periphery/pull/451.

### DIFF
--- a/src/hooks/ERC4626Hook.sol
+++ b/src/hooks/ERC4626Hook.sol
@@ -1,0 +1,65 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.26;
+
+import {ERC20} from "solmate/src/tokens/ERC20.sol";
+import {ERC4626} from "solmate/src/mixins/ERC4626.sol";
+import {BaseTokenWrapperHook} from "../base/hooks/BaseTokenWrapperHook.sol";
+import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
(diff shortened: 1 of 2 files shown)


scorpion9979:
Ideally ERC4626Hook should be defined as an abstract contract, but I wasn't sure what's the desired convention for testing an abstract contract in this repo.

partylikeits1983 (commented):
Overall this PR looks interesting and combining a hook with ERC4626 opens many interesting use cases. However, could you explain the exact use case for when this hook would be used?

marktoda:
> Ideally ERC4626Hook should be defined as an abstract contract, but I wasn't sure what's the desired convention for testing an abstract contract in this repo.

Why should it be abstract? Typical convention for testing abstract contracts is to make simple test implementations of virtuals and test those!

scorpion9979:
> > Ideally ERC4626Hook should be defined as an abstract contract, but I wasn't sure what's the desired convention for testing an abstract contract in this repo.
> 
> Why should it be abstract? Typical convention for testing abstract contracts is to make simple test implementations of virtuals and test those!

@marktoda I was torn whether it should be or not from a nitpick standpoint, as ERC4626 is a generic standard rather than a specific implementation like wstETH. That being said, it doesn't contain any virtual functions and any specific contract implementations extending it would only utilize the constructor instantiation, so perhaps it's fine as is.

partylikeits1983 (approved):
LGTM

PR #458, opened 2025-04-02
Add `msgSender()` to V4Quoter

saucepoint:
## Related Issue

There's a growing subset of hooks which read the "wallet" of the swapper (or position creator). These hooks normally use the `msgSender()` view function exposed by Universal Router and PosM, and can facilitate transactions without failure

However, `msgSender()` was not available on V4Quoter, causing reverts/failures when using the quoter against a hook'd pool reading `msgSender()`

## Description of changes

Add `msgSender() external view returns (address)` to V4Quoter.sol via transient storage

Add `IMsgSender` and added inheritance to IV4Quoter, BaseActionsRouter

### DIFF
--- a/src/base/BaseActionsRouter.sol
+++ b/src/base/BaseActionsRouter.sol
@@ -5,10 +5,11 @@ import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
 import {SafeCallback} from "./SafeCallback.sol";
 import {CalldataDecoder} from "../libraries/CalldataDecoder.sol";
 import {ActionConstants} from "../libraries/ActionConstants.sol";
+import {IMsgSender} from "../interfaces/IMsgSender.sol";
 
 /// @notice Abstract contract for performing a combination of actions on Uniswap v4.
 /// @dev Suggested uint256 action values are defined in Actions.sol, however any definition can be used
-abstract contract BaseActionsRouter is SafeCallback {
+abstract contract BaseActionsRouter is IMsgSender, SafeCallback {
     using CalldataDecoder for bytes;
 
     /// @notice emitted when different numbers of parameters and actions are provided
--- a/src/interfaces/IMsgSender.sol
+++ b/src/interfaces/IMsgSender.sol
@@ -1,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/// @title IMsgSender
+/// @notice Interface for contracts that expose the original msg.sender
+interface IMsgSender {
+    function msgSender() external view returns (address);
+}
--- a/src/interfaces/IMsgSender.sol
+++ b/src/interfaces/IMsgSender.sol
@@ -7,7 +7,7 @@ interface IMsgSender {
     /// @notice Returns the address of the original caller (msg.sender)
     /// @dev Uniswap v4 periphery contracts implement a callback pattern which lose
     /// the original msg.sender caller context. This view function provides a way for
-    /// integrating contracts (hooks) to access the original caller address.
+    /// integrating contracts (e.g. hooks) to access the original caller address.
     /// @return The address of the original caller
     function msgSender() external view returns (address);
 }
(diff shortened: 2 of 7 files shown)


src/base/BaseActionsRouter.sol
```diff
@@ -5,10 +5,11 @@ import {IPoolManager} from "@uniswap/v4-core/src/interfaces/IPoolManager.sol";
 import {SafeCallback} from "./SafeCallback.sol";
 import {CalldataDecoder} from "../libraries/CalldataDecoder.sol";
 import {ActionConstants} from "../libraries/ActionConstants.sol";
+import {IMsgSender} from "../interfaces/IMsgSender.sol";
 
 /// @notice Abstract contract for performing a combination of actions on Uniswap v4.
 /// @dev Suggested uint256 action values are defined in Actions.sol, however any definition can be used
-abstract contract BaseActionsRouter is SafeCallback {
+abstract contract BaseActionsRouter is IMsgSender, SafeCallback {
```
hensha256, line 12: I think the msgSender function should be removed from BaseActionsRouter file then, if its being inherited from an interface?

src/lens/V4Quoter.sol
```diff
@@ -138,4 +143,15 @@ contract V4Quoter is IV4Quoter, BaseV4Quoter {
         uint256 amountIn = params.zeroForOne ? uint128(-swapDelta.amount0()) : uint128(-swapDelta.amount1());
         amountIn.revertQuote();
     }
+
+    function msgSender() external view returns (address) {
```
hensha256, line 147: i'm wondering if this should be implemented in MsgSender.sol and it could be an abstract contract? or do you think some people will implement the function in a different way?

src/lens/V4Quoter.sol
```diff
@@ -138,4 +143,15 @@ contract V4Quoter is IV4Quoter, BaseV4Quoter {
...
+        // despite using the Locker library, V4Quoter does not have a reentrancy lock
+        return Locker.get();
+    }
+
+    modifier setMsgSender() {
```
hensha256, line 152: can you move this definition above the functions?

src/interfaces/IMsgSender.sol
```diff
@@ -0,0 +1,8 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/// @title IMsgSender
+/// @notice Interface for contracts that expose the original msg.sender
+interface IMsgSender {
+    function msgSender() external view returns (address);
```
hensha256, line 7: natspec 🙏

src/lens/V4Quoter.sol
```diff
@@ -138,4 +143,15 @@ contract V4Quoter is IV4Quoter, BaseV4Quoter {
         uint256 amountIn = params.zeroForOne ? uint128(-swapDelta.amount0()) : uint128(-swapDelta.amount1());
         amountIn.revertQuote();
     }
+
```
hensha256, line 146: `@inheritdoc`

src/base/BaseActionsRouter.sol
saucepoint: the virtual function itself is used by BaseActionsRouter, so I dont think we can remove it

```solidity
    function msgSender() public view virtual returns (address);

    /// @notice Calculates the address for a action
    function _mapRecipient(address recipient) internal view returns (address) {
        if (recipient == ActionConstants.MSG_SENDER) {
            return msgSender(); // <======
        } else if (recipient == ActionConstants.ADDRESS_THIS) {
            return address(this);
        } else {
            return recipient;
        }
    }
```

I think either BaseActionsRouter inherits `IMsgSender` or `IBaseActionsRouter is IMsgSender` -- the latter is weird because BAR's only external function IS `msgSender()` haha

src/lens/V4Quoter.sol
saucepoint: i think its okay as-is, posm/UR implement `msgSender()` through _ReentrancyLock_ while quoter implements it directly via _Locker_

so theres a possibility someone implements it different, i.e. with contract storage for some reason

snreynolds (approved):
lgtm! can you add a test though?

PR #460, opened 2025-04-18
[audit] Staked ETH Rounding - Disable exact output

marktoda:
Due to an unavoidable property of the share mechanics used internally within the stETH implementation, it is not currently possible for v4 to support exact outputs in either direction when interacting with stETH. Rather than violate the spirit of the exactOut request by either overshooting the conversion target or ignoring the unexpected delta, this implementation opts to disallow exactOut by reverting in the `beforeSwap` hook. This resolves H-01.
h/t @ccashwell 

Issue: _getWrapInputRequired function calculates the amount of underlying tokens required to be wrapped in order to achieve the exact desired output amount. Since conversions for wrapping round down (in favor of the protocol), the inverse function, which determines how much underlying must be wrapped to achieve the exact output, must round up.
In the wstEthHook[ contract](https://github.com/Uniswap/v4-periphery/blob/9628c36b4f5083d19606e63224e4041fe748edae/src/hooks/WstETHHook.sol), the getStETHByWstETH function rounds in the wrong direction. The input amount is calculated by getStETHByWstETH, which in turn calls getPooledEthByShares, a function that [rounds down](https://github.com/lidofinance/core/blob/9a5a20a2044bcf904603a08ac4eee367b7f2dc2e/contracts/0.4.24/StETH.sol#L309-L313). As a result, after wrapping, if one more asset is required to achieve the exact output, the pool manager attempts to swap through the pool. If the pool lacks sufficient liquidity, the swap attempt will revert. This issue also arises when unwrapping wstETH to achieve an exact output, due to the rounding in the _getUnwrapInputRequired function.
In order to account for the rounding direction, when wrapping, consider dividing the desired output by the conversion rate used for wrapping and adding 1 if there is a remainder. In addition, consider adding comments above the _getWrapInputRequired and _getUnwrapInputRequired functions to clarify the correct rounding direction when wrapping and unwrapping assets.


src/base/hooks/BaseTokenWrapperHook.sol
```diff
@@ -154,17 +158,29 @@ abstract contract BaseTokenWrapperHook is BaseHook, DeltaResolver {
 
     /// @notice Deposits underlying tokens to receive wrapper tokens
     /// @param underlyingAmount The amount of underlying tokens to deposit
+    /// @return actualUnderlyingAmount the actual number of underlying tokens used, i.e. to account for rebasing rounding errors
```
hensha256, line 161:
```suggestion
    /// @return actualUnderlyingAmount The actual number of underlying tokens wrapped, i.e. to account for rebasing rounding errors
```

src/base/hooks/BaseTokenWrapperHook.sol
```diff
@@ -154,17 +158,29 @@ abstract contract BaseTokenWrapperHook is BaseHook, DeltaResolver {
...
     /// @return wrappedAmount The amount of wrapper tokens received
-    /// @dev Implementing contracts should handle the wrapping operation
-    ///      The base contract will handle settling tokens with the pool manager
-    function _deposit(uint256 underlyingAmount) internal virtual returns (uint256 wrappedAmount);
+    /// @dev Implementing contracts should handle:
+    //    - taking tokens from PoolManager
+    //    - performing the wrapping operation
+    //    - settling tokens on PoolManager
```
hensha256, line 166:
```suggestion
    //    - taking underlying currency from PoolManager
    //    - performing the wrapping operation
    //    - settling wrapped currency on PoolManager
```

src/base/hooks/BaseTokenWrapperHook.sol
```diff
@@ -154,17 +158,29 @@ abstract contract BaseTokenWrapperHook is BaseHook, DeltaResolver {
...
+    function _deposit(uint256 underlyingAmount)
+        internal
+        virtual
+        returns (uint256 actualUnderlyingAmount, uint256 wrappedAmount);
 
     /// @notice Withdraws wrapper tokens to receive underlying tokens
     /// @param wrappedAmount The amount of wrapper tokens to withdraw
+    /// @return actualWrappedAmount the actual number of wrapped tokens used, i.e. to account for rebasing rounding errors
```
hensha256, line 174:
```suggestion
    /// @return actualWrappedAmount the actual number of wrapped tokens used, i.e. to account for rebasing rounding errors
```
```suggestion
    /// @return actualWrappedAmount The actual number of wrapped tokens unwrapped, i.e. to account for rebasing rounding errors
```

src/base/hooks/BaseTokenWrapperHook.sol
```diff
@@ -154,17 +158,29 @@ abstract contract BaseTokenWrapperHook is BaseHook, DeltaResolver {
...
     /// @return underlyingAmount The amount of underlying tokens received
-    /// @dev Implementing contracts should handle the unwrapping operation
-    ///      The base contract will handle settling tokens with the pool manager
-    function _withdraw(uint256 wrappedAmount) internal virtual returns (uint256 underlyingAmount);
+    /// @dev Implementing contracts should handle:
+    //    - taking tokens from PoolManager
+    //    - performing the unwrapping operation
+    //    - settling tokens on PoolManager
```
hensha256, line 179:
```suggestion
    //    - taking wrapped currency from PoolManager
    //    - performing the unwrapping operation
    //    - settling underlying currency on PoolManager
```

src/base/hooks/BaseTokenWrapperHook.sol
```diff
@@ -154,17 +158,29 @@ abstract contract BaseTokenWrapperHook is BaseHook, DeltaResolver {
...
-    ///      The base contract will handle settling tokens with the pool manager
-    function _deposit(uint256 underlyingAmount) internal virtual returns (uint256 wrappedAmount);
+    /// @dev Implementing contracts should handle:
+    //    - taking tokens from PoolManager
+    //    - performing the wrapping operation
+    //    - settling tokens on PoolManager
+    function _deposit(uint256 underlyingAmount)
+        internal
+        virtual
+        returns (uint256 actualUnderlyingAmount, uint256 wrappedAmount);
```
hensha256, line 170: when `actualUnderlyingAmount` is different to `underlyingAmount` (or the analogous for withdraw), wouldnt this mean that some dust delta might be left open in that currency?
i.e. someone settles 1 ETH, the hook tries to wrap that 1 ETH but only wraps 0.999999 ETH, and then a delta of 0.000001 ETH is left open in ETH?

marktoda: in this case the extra 1 wei is entirely lost due to rounding. if the wei is still held (ie checked via balanceof) then the correct value here is 1 eth

hensha256 (approved):

Issue #473, opened 2025-05-14
Add Swap to Router that enables providing SwapParams directly


johnny-emp:
### Component

General design optimization (improving efficiency, cleanliness, or developer experience)

### Describe the suggested feature and problem it solves.

I have use cases where i'd like to use the universal router to call a swap, but I typically would provide the sqrtPriceLimitX96.  This was exposed in the UniswapV3 router, but it's documented as no longer being exposed to protect exactOut swaps.

https://github.com/Uniswap/v4-periphery/blob/main/src/V4Router.sol#L160

PR #474, opened 2025-05-14
add raw swap to enable providing sqrt price limit to router

johnny-emp:
## Related Issue
#473

## Description of changes
I add an additional action to the router, where you can provide the SwapParams directly.  I did not do the unchecked serialization, but I wanted to see if there was a strong reason for not including this in the V4Router


eversoluna:
Great suggestion! Why this PR isn't reviewed yet?

PR #479, opened 2025-06-25
Permissioned router tests, permissioned posm tests

ericneil-sanc:
This PR adds a 100% coverage test suite for the permissioned pools functionality, specifically tests for the permissioned router and position manager. The implementation includes comprehensive test coverage with proper setup, mock contracts, and helper utilities.

### **New Test Files Added:**
- **`test/hooks/permissionedPools/PermissionedPositionManager.t.sol`** 
  - Test suite for permissioned position manager functionality
  - Tests for minting, burning, slippage protection, and permission checks
  - Tests all combinations of permissioned/unpermissioned
  - Fuzz testing for various liquidity scenarios

- **`test/hooks/permissionedPools/PermissionedV4Router.t.sol`** 
  - Router test suite covering exact input/output swaps
  - Multi-hop swap testing (1-hop, 2-hop, 3-hop)
  - Native token integration tests
  - Gas optimization tests

- **`test/hooks/permissionedPools/shared/PermissionedDeployers.sol`** 
  - Deployment utilities for permissioned components
  - Currency and token setup helpers

- **`test/hooks/permissionedPools/shared/PermissionedPosmTestSetup.sol`** 
  - Position manager test setup and utilities
  - Integration helpers

- **`test/hooks/permissionedPools/shared/PermissionedRoutingTestHelpers.sol`** 
  - Router test setup and helper functions
  - Pool creation and liquidity seeding utilities

- **`test/hooks/permissionedPools/mocks/MockPermissionedV4Router.sol`** 
  - Mock router for testing purposes

### **Infrastructure:**
- **`test/shared/Deploy.sol`** 
  - Added permissioned position manager deployment functionality

- **`script/FindHookAddressCreate3.s.sol`** 
  - CREATE3-based hook address finder script
  
  ### **Contracts**
- **`src/hooks/permissionedPools/PermissionedHooks.sol`**
  - Had to seperate the hooks from the router to ensure beforeSwap hook actually gets called


test/hooks/permissionedPools/PermissionedV4Router.t.sol
```diff
@@ -0,0 +1,862 @@
...
+import {Actions} from "../../../src/libraries/Actions.sol";
+import {ActionConstants} from "../../../src/libraries/ActionConstants.sol";
+import {IAllowlistChecker} from "../../../src/hooks/permissionedPools/interfaces/IAllowlistChecker.sol";
+import {WrappedPermissionedToken} from "../../../src/hooks/permissionedPools/WrappedPermissionedToken.sol";
+import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import {IHooks} from "@uniswap/v4-core/src/interfaces/IHooks.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+import {MockPermissionedToken} from "./PermissionedPoolsBase.sol";
+
+contract PermissionedV4RouterTest is PermissionedRoutingTestHelpers {
```
gretzke, line 17: can you add a test where the permissioned token is in the middle of a 3 hop swap? so non permissioned => permissioned => non permissioned, this would mean the user wouldn't have to transfer a permissioned token but the allowlist should still be verified on the middle swap

test/hooks/permissionedPools/PermissionedPositionManager.t.sol
```diff
@@ -0,0 +1,522 @@
...
+import {ReentrancyLock} from "../../../src/base/ReentrancyLock.sol";
+import {ModifyLiquidityParams} from "@uniswap/v4-core/src/types/PoolOperation.sol";
+
+import {MockAllowlistChecker, MockPermissionedToken} from "./PermissionedPoolsBase.sol";
+import {WrappedPermissionedToken, IERC20} from "../../../src/hooks/permissionedPools/WrappedPermissionedToken.sol";
+import {WrappedPermissionedTokenFactory} from "../../../src/hooks/permissionedPools/WrappedPermissionedTokenFactory.sol";
+import {Hooks} from "@uniswap/v4-core/src/libraries/Hooks.sol";
+import {SortTokens} from "@uniswap/v4-core/test/utils/SortTokens.sol";
+
+contract PermissionedPositionManagerTest is Test, PermissionedPosmTestSetup, LiquidityFuzzers {
```
gretzke, line 46: Can you add a test that liquidity can only be added to pools with this hook by the permissioned position manager and no other position managers? Even if they are approved as wrappers by the wrapped permissioned token

gretzke, line 46: There should be tests covering the 4 branches in the overridden _pay function:
1. non permissioned token - caller is payer
2. non permissioned token - posm is payer
3. permissioned token - caller is payer
4. permissioned token - posm is payer
same for the v4 router

test/hooks/permissionedPools/PermissionedV4Router.t.sol
gretzke, line 17: There should also be a test with the success case of a swap with a permissioned token

src/hooks/permissionedPools/PermissionedHooks.sol
```diff
@@ -0,0 +1,145 @@
...
+contract PermissionedHooks is IHooks, ReentrancyLock {
+    IWrappedPermissionedTokenFactory public immutable WRAPPED_TOKEN_FACTORY;
+    address public immutable PERMISSIONED_POSITION_MANAGER;
+
+    address public PERMISSIONED_ROUTER;
+
+    error Unauthorized();
+    error HookNotImplemented();
+
+    /// @dev as this contract and the swap router rely on each others addresses in the constructor, both contracts need
```
marktoda, line 28: is there a cleaner workaround for this? Perhaps one could deploy the other?

marktoda: requirement to use create3 is a bit of a smell to me

marktoda:
big diff; are we intermixing commits with previous PRs?

src/hooks/permissionedPools/WrappedPermissionedTokenFactory.sol
```diff
@@ -35,7 +35,7 @@ contract WrappedPermissionedTokenFactory is IWrappedPermissionedTokenFactory {
         IERC20 permissionedToken = IERC20(permissionedTokenOf[wrappedToken]);
         if (address(permissionedToken) == address(0)) revert WrappedTokenNotFound(wrappedToken);
         if (verifiedPermissionedTokenOf[wrappedToken] != address(0)) revert WrappedTokenAlreadyVerified(wrappedToken);
-        if (permissionedToken.balanceOf(wrappedToken) != 0) revert WrappedTokenNotVerified(wrappedToken);
+        if (permissionedToken.balanceOf(wrappedToken) == 0) revert WrappedTokenNotVerified(wrappedToken);
```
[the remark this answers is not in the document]
ericneil-sanc: It is to prove that the person who made the contract actually has some ownership of the permissioned token

ericneil-sanc:
> The tests themselves look great, but this is a big diff; are we intermixing commits with previous PRs?
@marktoda  This PR is only supposed to be adding the test suite. Any other changes with the non-test Permissioned Pools contracts are fixes or necessary modifications with the logic, that were highlighted by the tests

src/hooks/permissionedPools/PermissionedHooks.sol
ericneil-sanc: Actually IIRC one of the reasons we went with circular-dependant CREATE3 instead of auto-deploying was we were extremely close to hitting contract size limits on PermissionedPositionManager as is

src/base/DeltaResolver.sol
```diff
@@ -76,7 +76,7 @@ abstract contract DeltaResolver is ImmutableState {
     }
 
     /// @notice Calculates the amount for a settle action
-    function _mapSettleAmount(uint256 amount, Currency currency) internal view returns (uint256) {
+    function _mapSettleAmount(uint256 amount, Currency currency) internal view virtual returns (uint256) {
```
snreynolds, line 79: We should use tags to make sure that we are keeping track of what DeltaResolver contracts are used for which protocols

Ie with this change we will have now a few different versions of DeltaResolver

src/hooks/permissionedPools/PermissionedHooks.sol
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /// @dev Returns the hook permissions configuration for this contract
+    function getHookPermissions() public pure returns (Hooks.Permissions memory permissions) {
+        permissions.beforeSwap = true;
+        permissions.beforeAddLiquidity = true;
+    }
+
+    /// @inheritdoc IHooks
+    function beforeSwap(address sender, PoolKey calldata key, SwapParams calldata, bytes calldata)
```
snreynolds, line 37: Can you add a comment why you do not need to check that msg.sender == poolManager in beforeAddLiquidity and beforeSwap?

This should also include the reasoning why checking that the router and the routers saved sender is safe!

snreynolds: Its just important to note that because we have a direct trustful relationship with the router, we do not need the extra check that the call is coming from the v4 pool manager. Of course, a fake pool manager could maliciously set the sender as the router address, bypassing this check, but it couldn't actually do anything on the real pool. 

just worth noting bc in a lot of hook contracts we recommend checking that the call comes from the pool manager but here that check is not needed

src/hooks/permissionedPools/PermissionedHooks.sol
```diff
@@ -0,0 +1,135 @@
...
+import {ActionConstants} from "../../libraries/ActionConstants.sol";
+import {ReentrancyLock} from "../../base/ReentrancyLock.sol";
+import {IAllowanceTransfer} from "permit2/src/interfaces/IAllowanceTransfer.sol";
+import {Hooks, IHooks} from "@uniswap/v4-core/src/libraries/Hooks.sol";
+import {PoolKey} from "@uniswap/v4-core/src/types/PoolKey.sol";
+import {ModifyLiquidityParams, SwapParams} from "@uniswap/v4-core/src/types/PoolOperation.sol";
+import {BeforeSwapDelta, BeforeSwapDeltaLibrary} from "@uniswap/v4-core/src/types/BeforeSwapDelta.sol";
+import {BalanceDelta} from "@uniswap/v4-core/src/types/BalanceDelta.sol";
+
+contract PermissionedHooks is IHooks, ReentrancyLock {
```
snreynolds, line 19: We could probably inherit a version of BaseHooks so that we don't have to define all of the unused callbacks that revert at the top level, if we wanted to.. Just to clean up this file

src/hooks/permissionedPools/PermissionedPositionManager.sol
```diff
@@ -61,7 +61,7 @@ contract PermissionedPositionManager is PositionManager {
         bytes calldata hookData
     ) internal override {
         // allowlist is verified in the hook call
-        if (address(poolKey.hooks) != PERMISSIONED_SWAP_ROUTER) revert InvalidHook();
+        if (poolKey.hooks != PERMISSIONED_HOOKS) revert InvalidHook();
```
snreynolds, line 64: hm so there is only 1 hook that is valid to operate with this position manager?

src/base/DeltaResolver.sol
gretzke: It realistically should not affect the bytecode for any deployed contracts

src/hooks/permissionedPools/PermissionedHooks.sol
gretzke: Good catch, we can use BaseHooks here. The reason why we implemented this in the first place was when we were using the swap router / posm as the hook we would get an error that the constructor was initialized twice, because both contracts rely on `ImmutableState`

ericneil-sanc: The only thing with that is on BaseHook, only the internal functions are virtual, we would have to either edit BaseHook to make the external functions virtual, OR make some copy of BaseHook that has those functions as virtual

src/hooks/permissionedPools/PermissionedHooks.sol
ericneil-sanc: Sure!

gretzke (approved):

PR #481, opened 2025-06-28
feat: Add comprehensive reentrancy protection to Notifier system

SteOnChain:
- Implement dedicated reentrancy guard for external subscription calls
- Add notifierNonReentrant modifier to subscribe() and unsubscribe() functions
- Enhance _safeCall() with self-call prevention and status monitoring
- Use state-before-call pattern with automatic rollback on failure
- Add defensive try-catch in unsubscribe to prevent blocking by malicious subscribers
- Maintain compatibility with existing ReentrancyLock system

Fixes potential reentrancy vulnerabilities in subscriber notifications during position modifications, subscriptions, and burns.

## Related Issue

## Description of changes

### DIFF
--- a/src/PositionDescriptor.sol
+++ b/src/PositionDescriptor.sol
@@ -18,6 +18,9 @@ import {SafeCurrencyMetadata} from "./libraries/SafeCurrencyMetadata.sol";
 contract PositionDescriptor is IPositionDescriptor {
     using StateLibrary for IPoolManager;
 
+    /// @notice Thrown when a zero address is provided for a critical parameter
+    error ZeroAddress(string parameter);
+
     // mainnet addresses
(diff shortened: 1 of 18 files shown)


BhariGowda:
## Critical Bug: transferFrom uses msgSender() instead of msg.sender

transferFrom is gated by onlyIfPoolManagerLocked, meaning it only executes 
when NOT inside a modifyLiquidities unlock callback. In this context, 
msgSender() returns address(0) because Locker.get() reads transient storage 
that is only set during an active unlock.

Result: every transferFrom call reverts with NotAuthorized. All ERC-721 
position transfers are permanently broken if this PR merges.

Fix: replace address caller = msgSender() with address caller = msg.sender 
in PositionManager.transferFrom.

PR #495, opened 2025-08-20
feat(WIP): Uniswap V2 Adapter Hook

marktoda:
This commit is a proof of concept for a hook that adapts an existing liquidity source such as Uniswap V2. This allows routers to interact with various liquidity systems using the Uniswap V4 interface and route across Uniswap V4 and others without leaving the lock.


src/hooks/UniswapV2AdapterHook.sol
```diff
@@ -0,0 +1,128 @@
...
+            uint256 reserveOut,
+            uint256 amountSpecified
+        ) = _parseSwap(poolKey, params);
+        amountOut = UniswapV2Library.getAmountOut(amountSpecified, reserveIn, reserveOut);
+        (uint256 amount0Out, uint256 amount1Out) = params.zeroForOne ? (uint256(0), amountOut) : (amountOut, uint256(0));
+
+        // Sync output token balance before V2 swap
+        poolManager.sync(tokenOut);
+        // Transfer input tokens from V4 pool manager directly to V2 pair
+        _take(tokenIn, pair, amountSpecified);
```
snreynolds, line 56: This means that tokenIn must exist in pool manager which might not be the case _always_

unless we force that the router send in the user's tokens to v4 before the swap is called

marktoda: Yeah this is the case with the token wrapper hooks as well (WETH Hook, wsteth hook). I think Universal Router sends user's tokens in first now so typically these work!

PR #523, opened 2026-04-13
Validating Permission-Flags in PermissionedV4Router and PermissionedPositionManager

SocksNFlops:
## Related Issue
- Fixed [ECO-191](https://linear.app/uniswap/issue/ECO-191/sc-h-01-user-permission-checks-can-be-bypassed-when-permissioned-pools)

## Description of changes
- Adds an additional check to validate that SWAP_ALLOWED and LIQUIDITY_ALLOWED flags can't be used for the other's operations during user-funded settlement paths 
- Added unit tests to validate checks


test/hooks/permissionedPools/shared/PermissionedDeployers.sol
```diff
@@ -147,6 +149,18 @@ contract PermissionedDeployers is Test {
         deployedHooksAddr = calculatedAddr;
     }
 
+    function deployInsecureHooks(address manager_) internal returns (address deployedHooksAddr) {
+        uint160 flags = (1 << 11) | (1 << 7);
```
ericneil-sanc, line 153: Flags should now include BEFORE_SWAP_RETURNS_DELTA_FLAG (1 << 3). Here and everywhere else we deploy the hooks

ericneil-sanc (approved):

PR #534, opened 2026-04-17
fix(permissioned-pools): use solmate SafeTransferLib in PermissionsAdapter._unwrap (ECO-198)

codyborn:
## Summary
- Replaces the bare `PERMISSIONED_TOKEN.transfer` call in `PermissionsAdapter._unwrap` with solmate's `SafeTransferLib.safeTransfer` so burn + payout stay atomic. Matches the pattern already used elsewhere in the repo (e.g. `WstETHHook`).
- Handles two edge cases that the prior code silently ignored:
  - ERC-20 tokens that return `false` on failed transfers instead of reverting.
  - Tokens with pause/freeze logic where `transfer` silently returns `false`.
- With `safeTransfer`, any such failure now reverts the whole transaction, preventing a scenario where adapter supply is burned but the underlying payout never lands.
- Because the contract already inherits from OpenZeppelin's `ERC20`, solmate's `ERC20` is imported as `SafeERC20` to avoid the name collision in the `using SafeTransferLib for SafeERC20` directive.

Fixes ECO-198.

## Test plan
- [x] `testRevert_UnwrapWhenUnderlyingReturnsFalse` — mock token that returns `false` on `transfer` causes unwrap to revert; adapter supply and PM balance are unchanged.
- [x] `testRevert_UnwrapWhenUnderlyingPausedSilentFail` — mock pausable token silently returning `false` when paused causes unwrap to revert; adapter supply and recipient balance are unchanged.
- [x] `test_UnwrapSucceedsAfterUnpause` — well-behaved ERC-20 still unwraps through the new `safeTransfer` path.
- [x] Existing `test_UnwrapOnPoolManagerTransfer` fuzz test continues to pass.

### DIFF
--- a/src/hooks/permissionedPools/PermissionsAdapter.sol
+++ b/src/hooks/permissionedPools/PermissionsAdapter.sol
@@ -3,15 +3,15 @@ pragma solidity 0.8.26;
 
 import {IHooks} from "@uniswap/v4-core/src/interfaces/IHooks.sol";
 import {Ownable2Step, Ownable} from "@openzeppelin/contracts/access/Ownable2Step.sol";
-import {ERC20, IERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
-import {ERC20 as SolmateERC20} from "solmate/src/tokens/ERC20.sol";
+import {ERC20 as OZERC20, IERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import {ERC20} from "solmate/src/tokens/ERC20.sol";
 import {SafeTransferLib} from "solmate/src/utils/SafeTransferLib.sol";
 import {IPermissionsAdapter} from "./interfaces/IPermissionsAdapter.sol";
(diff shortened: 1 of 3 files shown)


src/hooks/permissionedPools/PermissionsAdapter.sol
```diff
@@ -108,7 +109,7 @@ contract PermissionsAdapter is ERC20, Ownable2Step, IPermissionsAdapter {
 
     function _unwrap(address account, uint256 amount) internal {
         _burn(account, amount);
-        PERMISSIONED_TOKEN.transfer(account, amount);
+        SafeERC20.safeTransfer(PERMISSIONED_TOKEN, account, amount);
```
SocksNFlops: `using SafeERC20 for IERC20`?

SocksNFlops (approved):
There's a failing test. Please fix.

src/hooks/permissionedPools/PermissionsAdapter.sol
codyborn: Updated to use solmate's SafeTransferLib for consistency

PR #537, opened 2026-04-20
fix(permissioned-posm): enforce hook allowlist on liquidity increases

SocksNFlops:
## Related Issue
[ECO-218](https://linear.app/uniswap/issue/ECO-218/sc-l-07-hook-allowlist-revocation-is-not-enforced-on-liquidity)

## Description of changes
- Checks `allowedHooks` on liquidity increases to prevent users from increasing their liquidity on hooks that have already been removed by the admin. Does not check on decreases to allow them to exit.


test/hooks/permissionedPools/PermissionedPositionManager.t.sol
```diff
@@ -1512,4 +1512,179 @@ contract PermissionedPositionManagerTest is Test, PermissionedPosmTestSetup, Liq
...
+    }
+
+    // Helpers for toggling the hook allowlist in tests. Uses a raw selector call to mirror the
+    // pattern established by `setAllowedHooks` (line 227).
+    function _setHookAllowedForKey(PoolKey memory key, bool allowed) internal {
+        _maybeSetHookAllowed(key.currency0, key.hooks, allowed);
+        _maybeSetHookAllowed(key.currency1, key.hooks, allowed);
+    }
+
+    function _maybeSetHookAllowed(Currency currency, IHooks hooks_, bool allowed) internal {
```
ericneil-sanc: nit; I don't like the term maybe in function names, outside of randomness, since that implies non-determinism

ericneil-sanc (approved):

PR #538, opened 2026-04-24
fix(permissions-adapter): tolerate tokens without standard ERC20 meta…

SocksNFlops:
## Related Issue
Fixes [ECO-223](https://linear.app/uniswap/issue/ECO-223/sc-n-03-permissionsadapter-constructor-reverts-for-tokens-without)

## Description of changes
- Adds error-handling for situations when underlying secToken doesn't implement `name()`, `symbol()`, nor `decimals()` metadata functions
- Not using try/catch because it's insufficient for abi-decoding failures (expected types don't match)

### DIFF
--- a/src/hooks/permissionedPools/PermissionsAdapter.sol
+++ b/src/hooks/permissionedPools/PermissionsAdapter.sol
@@ -4,6 +4,7 @@ pragma solidity 0.8.26;
 import {IHooks} from "@uniswap/v4-core/src/interfaces/IHooks.sol";
 import {Ownable2Step, Ownable} from "@openzeppelin/contracts/access/Ownable2Step.sol";
 import {ERC20, IERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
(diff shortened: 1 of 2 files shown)


src/hooks/permissionedPools/PermissionsAdapter.sol
```diff
@@ -115,16 +116,37 @@ contract PermissionsAdapter is ERC20, Ownable2Step, IPermissionsAdapter {
...
+    function _readString(address token, bytes4 selector, string memory fallback_) private view returns (string memory) {
+        (bool ok, bytes memory data) = token.staticcall(abi.encodeWithSelector(selector));
+        if (!ok || data.length < 64) return fallback_;
+        uint256 offset;
+        uint256 length;
+        assembly ("memory-safe") {
+            offset := mload(add(data, 0x20))
+            length := mload(add(data, 0x40))
+        }
+        if (offset != 0x20 || length == 0 || length > data.length - 64) return fallback_;
```
ericneil-sanc, line 130: May want to ensure that length isn't so long that we run out of gas or something?

ericneil-sanc (approved):

PR #542, opened 2026-05-04
feat(permissioned-position-manager): Adds `unwindPosition` and `withdrawClaim` functionality

SocksNFlops:
## Related Issue
- [ECO-173](https://linear.app/uniswap/issue/ECO-221/sc-l-10-unilateral-admin-transfer-authority-in)

## Description of changes
- Reverting https://github.com/Uniswap/v4-periphery/pull/532
- Removing all transferability of PermissionedPositionManager positions
- Introducing `unwindPosition` + `withdrawClaim` functions

## Tests
Unwinding Tests:
| # | Test | Spec | What it verifies |                                                                                                                                                                                                         
  |---|---|---|---|                      
  | 1 | `test_unwindPosition_two_pas_same_admin_routes_both_to_lp(admin, defaultRecipient)` | Same admin owns both PAs, happy path | Helper called with `(admin, admin, defaultRecipient)` so both PAs end up owned by one address. LP compliant on both underlyings (default setUp), so cascade lands at step 1 for both → asserted via "alice ≈ whole on currency0 and currency2." Also checks `PositionUnwound(tokenId, alice, admin, defaultRecipient)` emission. |                         
  | 2a | `test_unwindPosition_pa_and_regular_routes_both_to_lp_when_compliant(admin1, admin2, defaultRecipient)` | PA + regular, LP compliant | key0 = `[pa0, regular]`. LP on currency0 allowlist (default). Cascade hits step 1 for both legs. Asserts alice is whole on currency0 + key0.currency1. |                                                                                                                                                                                        
  | 2b | `test_unwindPosition_pa_and_regular_routes_pa_to_default_when_lp_blocked(admin1, admin2, defaultRecipient)` | PA + regular, LP blocked, goes to defaultRecipient | LP delisted on currency0; defaultRecipient force-allowlisted. Captures `deposited0` at mint. After unwind: defaultRecipient delta on currency0 ≈ `deposited0` (step 2). Regular currency goes to alice (step 1, no compliance gate). |                                                                       
  | 2c | `test_unwindPosition_pa_and_regular_credits_6909_when_lp_and_default_blocked(admin1, admin2, defaultRecipient)` | PA + regular, both blocked, 6909 to defaultRecipient | LP and defaultRecipient both delisted on currency0. Asserts `manager.balanceOf(defaultRecipient, currency0.toId()) ≈ deposited0` (step 3). Regular still goes to alice. |                                                                                                                                  
  | 3a | `test_unwindPosition_two_pas_different_admins_routes_both_to_lp(admin1, admin2, defaultRecipient)` | Two PAs, different admins, both compliant | Helper splits ownership (pa0→admin1, pa2→admin2). admin1 calls. Both legs hit step 1. Asserts alice whole on both. |                                                                                                                                                                                                                 
  | 3b | `test_unwindPosition_two_pas_different_admins_routes_pa1_to_default_when_lp_blocked(admin1, admin2, defaultRecipient)` | admin1 calls; pa0 happy, pa2 → defaultRecipient | LP delisted on currency2; defaultRecipient force-allowlisted on currency2. pa0 leg: step 1 (alice). pa2 leg: step 2 (defaultRecipient). |                                                                                                                                                                   
  | 3c | `test_unwindPosition_two_pas_different_admins_credits_6909_when_lp_and_default_blocked_on_pa1(admin1, admin2, defaultRecipient)` | admin1 calls; pa0 happy, pa2 → 6909 | LP and defaultRecipient both delisted on currency2. pa0 leg: step 1. pa2 leg: step 3 (6909 mint to defaultRecipient). |

### DIFF
--- a/src/hooks/permissionedPools/PermissionedPositionManager.sol
+++ b/src/hooks/permissionedPools/PermissionedPositionManager.sol
@@ -243,7 +243,7 @@ contract PermissionedPositionManager is PositionManager {
     function _handleAction(uint256 action, bytes calldata params) internal override {
         if (action == _ACTION_TAKE_WITH_FALLBACK) {
             (Currency currency, address lp, address defaultRecipient) = abi.decode(params, (Currency, address, address));
-            _routeCredit(currency, lp, defaultRecipient);
+            _takeWithFallback(currency, lp, defaultRecipient);
             return;
         }
         if (action == Actions.BURN_6909) {
@@ -256,7 +256,7 @@ contract PermissionedPositionManager is PositionManager {
 
     /// @dev Cascade: try take to LP → take to defaultRecipient → mint 6909 claim to defaultRecipient.
     ///      Final mint never reverts.
-    function _routeCredit(Currency currency, address lp, address defaultRecipient) internal {
+    function _takeWithFallback(Currency currency, address lp, address defaultRecipient) internal {
         uint256 amount = _getFullCredit(currency);
         if (amount == 0) return;
 
--- a/src/hooks/permissionedPools/PermissionedPositionManager.sol
+++ b/src/hooks/permissionedPools/PermissionedPositionManager.sol
@@ -105,6 +105,9 @@ contract PermissionedPositionManager is PositionManager {
     ///      currencies, `to` must clear the underlying token's issuer compliance on unwrap. `to` follows the
     ///      standard `Actions.TAKE` recipient sentinels: `address(1)` remaps to the caller, `address(2)` to this
     ///      contract.
+    /// @param currency The currency whose 6909 claim is being burned
+    /// @param amount The amount of claim to burn (and underlying to deliver)
+    /// @param to The recipient of the underlying currency
     function withdrawClaim(Currency currency, uint256 amount, address to) external isNotLocked {
         bytes memory actions = abi.encodePacked(uint8(Actions.BURN_6909), uint8(Actions.TAKE));
         bytes[] memory params = new bytes[](2);
--- a/src/hooks/permissionedPools/PermissionedPositionManager.sol
+++ b/src/hooks/permissionedPools/PermissionedPositionManager.sol
@@ -15,13 +15,25 @@ import {IPermissionsAdapterFactory} from "./interfaces/IPermissionsAdapterFactor
 import {IHooks} from "@uniswap/v4-core/src/interfaces/IHooks.sol";
 import {PermissionFlags} from "./libraries/PermissionFlags.sol";
 import {ActionConstants} from "../../libraries/ActionConstants.sol";
+import {Actions} from "../../libraries/Actions.sol";
+import {CalldataDecoder} from "../../libraries/CalldataDecoder.sol";
 
 contract PermissionedPositionManager is PositionManager {
+    using CalldataDecoder for bytes;
+
+    /// @dev Takes a currency's positive delta with a fallback cascade: LP → defaultRecipient → 6909 mint to
+    ///      defaultRecipient. Picked above the standard Actions.sol range to avoid collision.
+    uint256 private constant _ACTION_TAKE_WITH_FALLBACK = 0x20;
+
     IPermissionsAdapterFactory public immutable PERMISSIONS_ADAPTER_FACTORY;
 
     mapping(Currency currency => mapping(IHooks hooks => bool)) public isAllowedHooks;
 
     event AllowedHooksUpdated(Currency currency, IHooks hooks, bool allowed);
+    event PositionUnwound(
+        uint256 indexed tokenId, address indexed lp, address indexed admin, address defaultRecipient
+    );
+    event ClaimWithdrawn(Currency indexed currency, address indexed from, address indexed to, uint256 amount);
 
     error InvalidHook();
     error TransferDisabled();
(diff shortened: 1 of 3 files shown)


src/hooks/permissionedPools/PermissionedPositionManager.sol
```diff
@@ -15,16 +15,26 @@ import {IPermissionsAdapterFactory} from "./interfaces/IPermissionsAdapterFactor
...
 import {ActionConstants} from "../../libraries/ActionConstants.sol";
+import {Actions} from "../../libraries/Actions.sol";
+import {CalldataDecoder} from "../../libraries/CalldataDecoder.sol";
 
 contract PermissionedPositionManager is PositionManager {
+    using CalldataDecoder for bytes;
+
+    /// @dev Takes a currency's positive delta with a fallback cascade: LP → defaultRecipient → 6909 mint to
+    ///      defaultRecipient. Picked above the standard Actions.sol range to avoid collision.
+    uint256 private constant _ACTION_TAKE_WITH_FALLBACK = 0x20;
```
codyborn: Nit: it seems like our other contracts don't prefix constants with `_`
e.g. https://github.com/Uniswap/v4-core/blob/46c6834698c48bc4a463a86d8420f4eb1d7f3b75/src/PoolManager.sol#L89

codyborn: Should we call it `ACTION_UNWIND_WITH_FALLBACK`?

codyborn: Should we add to `src/libraries/Actions.sol` to prevent overlap in the future?

SocksNFlops: > Should we add to src/libraries/Actions.sol to prevent overlap in the future?
I was hesitant to add it to `src/libraries/Actions.sol` because it's specific to PermissionedPositionManager.

SocksNFlops: > Should we call it `ACTION_UNWIND_WITH_FALLBACK`?

I called it `TAKE` not `UNWIND`, because there are 3 steps to the unwind:
1. Burn the Position
2. Send out assetA
3. Send out assetB

This action is what gets called for Step 2 & 3. It's not really "unwinding" anymore as much as it's "taking" an asset out of PoolManager.

ericneil-sanc: The only reason I don't like TAKE is because take already has meaning in V4 and is used all the time (see take/settle pattern)

src/hooks/permissionedPools/PermissionedPositionManager.sol
```diff
@@ -59,28 +69,62 @@ contract PermissionedPositionManager is PositionManager {
...
+        // Pre-approve so BURN_POSITION inside the unlock passes its onlyIfApproved check.
+        // ERC-721 _burn clears getApproved as part of its teardown, so the approval is self-cleaning.
+        getApproved[tokenId] = msg.sender;
+
+        bytes memory actions = abi.encodePacked(
+            uint8(Actions.BURN_POSITION), uint8(_ACTION_TAKE_WITH_FALLBACK), uint8(_ACTION_TAKE_WITH_FALLBACK)
+        );
+        bytes[] memory params = new bytes[](3);
+        params[0] = abi.encode(tokenId, uint128(0), uint128(0), bytes(""));
+        params[1] = abi.encode(poolKey.currency0, lp, defaultRecipient);
```
codyborn: I thought your proposal was to have a new role defined in the Adapter that specified who the defaultRecipient was rather than having it be passed in by one of the admins. 
The benefits:
1. More granular separation of admin and unwind recipient (defaults to the same)
2. One admin can't claim the tokens from the other admin or non-permissioned tokens.

src/hooks/permissionedPools/PermissionedPositionManager.sol
```diff
@@ -59,28 +69,62 @@ contract PermissionedPositionManager is PositionManager {
...
+        params[2] = abi.encode(poolKey.currency1, lp, defaultRecipient);
+        poolManager.unlock(abi.encode(actions, params));
+
+        emit PositionUnwound(tokenId, lp, msg.sender, defaultRecipient);
```
codyborn: It'd be worth also specifying in the event what was the ultimate recipient of each token

SocksNFlops: Migrated PositionUnwound to `CurrencyUnwound` which is emitted in each leg that is distributed

src/hooks/permissionedPools/PermissionedPositionManager.sol
SocksNFlops: Fixed. Added the additional case of "Regular ERC20 but transfer to LP fails." This defaults to a 6909 claim to the original LP.

src/hooks/permissionedPools/PermissionedPositionManager.sol
SocksNFlops: Updated to Actions.UNWIND_WITH_FALLBACK

src/hooks/permissionedPools/PermissionedPositionManager.sol
```diff
@@ -59,28 +74,60 @@ contract PermissionedPositionManager is PositionManager {
...
+        poolManager.unlock(abi.encode(actions, params));
+
+        emit ClaimWithdrawn(currency, msg.sender, to, amount);
+    }
+
     /// @inheritdoc PositionManager
-    /// @dev Only allow admins of permissioned tokens to transfer positions that contain their tokens,
-    /// and only to recipients that are allowlisted for each permissioned currency in the pool.
+    /// @dev Positions of permissioned tokens are not transferable.
     function transferFrom(address from, address to, uint256 id) public override onlyIfPoolManagerLocked {
```
codyborn, line 123: should we allow the position to be transferable by the LP (current owner) if the recipient is allowed for both tokens?

src/libraries/Actions.sol
```diff
@@ -46,4 +46,8 @@ library Actions {
     // note this is not supported in the position manager or router
```
codyborn, line 46: it is now!

src/hooks/permissionedPools/PermissionedPositionManager.sol
SocksNFlops: I think there are a number of reasons to lean against this:
1. It's possible for the allowList and underlying token's transfer checks to diverge. Normally, you have to be able to hold the underlying AND be on the allowList to make a position. Transferrable positions could take advantage of this gap.
2. If a user's LIQUIDITY_ALLOWED flag is removed, they can transfer it to a friend, which bypasses a lot of the intended checks.
3. It's equivalent to burning, transferring, and reminting, but goes through less issuer checks, so it simplifies existing functionality at the cost of bypassing issuer restrictions (i.e. they might have a daily minting limit).

src/libraries/Actions.sol
SocksNFlops: Only in the PermissionedPositionManager, not the regular one. Plus, it's probably better that users just talk directly to the PoolManager for this. It's only needed in this circumstance to unwrap PAs.

src/hooks/permissionedPools/PermissionedPositionManager.sol
ericneil-sanc: That is fair, although some other thoughts on this: I worry that disallowing #1 would preclude some of the benefits and unique aspects/abilities they would otherwise get from moving to a DeFi token model in the first place, like liquidity games and stuff. And I think one of the assumptions we have to operate on is that allowList = token transfer checks. 
For #2 that is true but I am not sure if that means we should make the sacrifice on non-transferrable LP positions

codyborn (approved):

src/hooks/permissionedPools/PermissionedPositionManager.sol
```diff
@@ -59,28 +73,60 @@ contract PermissionedPositionManager is PositionManager {
...
+        params[2] = abi.encode(poolKey.currency1, lp, tokenId);
+        poolManager.unlock(abi.encode(actions, params));
+    }
+
+    /// @notice Burn an ERC-6909 claim on the PoolManager and transfer the underlying currency to `to`.
+    /// @dev Caller must hold the claim or have called PoolManager.setOperator(permPosm, true). For permissioned
+    ///      currencies, `to` must clear the underlying token's issuer compliance on unwrap. `to` follows the
+    ///      standard `Actions.TAKE` recipient sentinels: `address(1)` remaps to the caller, `address(2)` to this
+    ///      contract.
+    function withdrawClaim(Currency currency, uint256 amount, address to) external isNotLocked {
```
ericneil-sanc, line 111: nit; externals should have proper NATSPEC (meaning it should include params)

ericneil-sanc (approved):

PR #543, opened 2026-05-06
fix(Permissioned-position-manager): unsubscribing on unwind

SocksNFlops:
## Related Issue
[ECO-371](https://linear.app/uniswap/issue/ECO-371/fixing-subscribe-griefing-in-unwindposition)

## Description of changes
- Auto-unsubscribes during `unwindPosition()` flows to prevent reverting subscriptions from blocking unwinds.

### DIFF
--- a/src/libraries/Actions.sol
+++ b/src/libraries/Actions.sol
@@ -50,4 +50,6 @@ library Actions {
     // permissioned-pools specific actions
     // routes a currency's positive delta with a fallback cascade: LP → defaultRecipient → 6909 mint to defaultRecipient
     uint256 internal constant UNWIND_WITH_FALLBACK = 0x19;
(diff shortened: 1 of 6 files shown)


src/libraries/Actions.sol
```diff
@@ -50,4 +50,6 @@ library Actions {
     // permissioned-pools specific actions
     // routes a currency's positive delta with a fallback cascade: LP → defaultRecipient → 6909 mint to defaultRecipient
     uint256 internal constant UNWIND_WITH_FALLBACK = 0x19;
+    // detaches a position's subscriber via the gas-capped + revert-swallowed unsubscribe path
+    uint256 internal constant UNSUBSCRIBE = 0x1a;
```
codyborn: should we add an enum placeholder for SUBSCRIBE so that they're sequential?

codyborn (approved):

PR #545, opened 2026-05-08
Cantina audit/pr 526 snapshot

SocksNFlops:
Reference Snapshot for Cantina Audit - **Do Not Merge**


BhariGowda:
Security Review permissionedPools (two issues)
Reviewed against commit e40c0a3. Both issues are in the new src/hooks/permissionedPools/ subtree.

[H-01] PermissionedPositionManager._increase() not overridden — isAllowedHooks check bypassed for existing positions
_mint() is correctly overridden to enforce the hook allow-list. _increase() is not overridden, so INCREASE_LIQUIDITY falls through to base PositionManager._increase() → _modifyLiquidity() → poolManager.modifyLiquidity() without consulting isAllowedHooks.
Impact: When an admin calls setAllowedHook(currency, H, false) to revoke a compromised hook, existing position holders can still call INCREASE_LIQUIDITY  depositing tokens into a pool backed by the revoked hook, with the hook's afterModifyLiquidity callback executing normally. The emergency control is defeated.
Fix: Override _increase() with the same _checkAllowedHooks + _checkRecipientAllowed guards as _mint(). Also evaluate _decrease() and _burn()  a compromised hook runs on any modifyLiquidity call.

[H-02] createPermissionsAdapter is permissionless; verifyPermissionsAdapter passes on 1 wei
createPermissionsAdapter accepts arbitrary permissionedToken, initialOwner, and allowListChecker from any caller. verifyPermissionsAdapter promotes an adapter to verified status based solely on balanceOf(adapter) > 0 satisfied by a 1 wei transfer.
Once verified, _getOwner() returns the adapter's owner() as the authority for setAllowedHook, giving the attacker hook governance over any pool using their adapter as currency.
Attack (cost: 1 wei):

factory.createPermissionsAdapter(USDC, attacker, maliciousChecker)
USDC.transfer(evilAdapter, 1)
factory.verifyPermissionsAdapter(evilAdapter) passes
posManager.setAllowedHook(evilAdapter, maliciousHook, true) passes

Fix: Gate createPermissionsAdapter to the token contract's owner. Replace the balance check in verifyPermissionsAdapter with a signature from the token's owner.

Happy to provide a Foundry PoC if useful for the Cantina engagement.

PR #562, opened 2026-06-12
hardening: track and refund msg.value for position manager settle

mrheyday:
## Summary
- track msg.value in transient storage during modifyLiquidities
- settle native with exact msg.value consumption
- refund unused ETH at end of execution
- reject msg.value on non-native settle paths

## Files
- src/PositionManager.sol
- src/base/DeltaResolver.sol

## Note
Based on existing hardening work from support-native-routing branch; replays cleanly and remains minimal.

### DIFF
--- a/src/PositionManager.sol
+++ b/src/PositionManager.sol
@@ -109,6 +109,9 @@ contract PositionManager is
     Permit2Forwarder,
     NativeWrapper
 {
+    using CurrencyLibrary for Currency;
+    using PoolIdLibrary for PoolKey;
+    using PositionConfigLibrary for PositionConfig;
     using StateLibrary for IPoolManager;
     using TransientStateLibrary for IPoolManager;
     using SafeCast for uint256;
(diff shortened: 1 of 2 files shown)


kimjune01:
The `UnexpectedValue` check makes mixed native/ERC20 `SETTLE_PAIR` order-dependent and defeats the overpayment refund path. [`_settlePair` settles `currency0` before `currency1`](https://github.com/Uniswap/v4-periphery/blob/f6d9987f3f24a1b11e2cff910cca64191b478869/src/PositionManager.sol#L471-L475). Native currency sorts first, so if the caller supplied the usual rounding buffer—an existing helper sends [`amount0 + 1`](https://github.com/Uniswap/v4-periphery/blob/f6d9987f3f24a1b11e2cff910cca64191b478869/test/shared/LiquidityOperations.sol#L44-L50)—the native branch [leaves that remainder in `MSG_VALUE_SLOT`](https://github.com/Uniswap/v4-periphery/blob/f6d9987f3f24a1b11e2cff910cca64191b478869/src/base/DeltaResolver.sol#L49-L56). The following ERC20 leg then [reverts on any nonzero tracked value](https://github.com/Uniswap/v4-periphery/blob/f6d9987f3f24a1b11e2cff910cca64191b478869/src/base/DeltaResolver.sol#L57-L62), so the modifier never reaches `_refundETH`.

A distinguishing test would mint/increase a native–ERC20 position through `SETTLE_PAIR` with `requiredNative + 1 wei`, then assert the ERC20 leg settles and 1 wei is refunded. I think the non-native guard needs to distinguish value intended for that leg from residual value awaiting the end-of-call refund, rather than rejecting any nonzero execution-level balance.

PR #563, opened 2026-06-15
feat: add margin trading periphery

ccashwell:
# Margin Trading Periphery

Adds a periphery that opens leveraged spot positions in a single transaction by composing a swap with a borrow/supply against an external lending protocol. **Four venues are integrated behind one router: Morpho Blue, Aave v3, Aave v4, and Compound v3**; the caller selects the venue per call by passing the matching adapter. The whole leverage loop runs inside one `PoolManager` unlock using v4 flash accounting, so it nets to zero with no intermediate capital and no router-held residual.

Full protocol and integration guide: [`docs/margin-trading.md`](docs/margin-trading.md).

## What a position is

Borrow the debt token, swap it into the collateral token (exact-output), supply the collateral (equity plus the bought amount), and draw the debt back to settle the swap. The result is a position **long the collateral and short the debt**, at a caller-chosen leverage bounded by the market's max LTV. Direction is set entirely by the `(collateral, debt)` pairing; there is no separate flag. Live markets today: long ETH (WETH/USDC on Morpho or Aave), short ETH (USDC/WETH on Morpho, Aave v3, or Aave v4), and long UNI (UNI/USDC on Compound v3).

## Routed swaps

Position swaps route through the **Universal Router** (the `ROUTE_SWAP` action), so a route can source liquidity across v2, v3, and v4 rather than a single v4 pool. The UR self-settles its own swap, so the router wraps the call in a flash-take envelope: it flash-takes the input from the PoolManager, funds the UR through a Permit2 allowance scoped to exactly the input cap (and zeroed after the swap, so no spendable allowance outlives the call), runs the caller-built route, and settles only this call's unspent take. The Universal Router is a **per-call parameter** (in `IncreaseParams`/`DecreaseParams`, or the first `ROUTE_SWAP` field in a plan), so callers can adopt newer UR deployments without a router redeploy.

The router does not trust the route's internals. Fills are all-or-nothing: the curated flows set the `ASSERT_ACCOUNT_BALANCE` threshold to the account's pre-unlock balance plus the amount the route must deliver, so the absolute check enforces the swap **delta** and a pre-existing or donated balance cannot mask a short fill (`IncompleteFill`).

## Architecture

| Contract | Role |
|----------|------|
| `MarginRouter` | Entry point and orchestrator. Builds each flow as a v4 action plan run inside one unlock. Inherits `V4Router`, `ReentrancyLock`, `Permit2Forwarder`, `Multicall_v4`, `NativeWrapper`, and the account factory. Is the trusted manager of every account. |
| `MarginAccount` | Per-user position container, a Solady clone-with-immutable-args. It is the lending counterparty (`onBehalf == account`), so it acts as itself with no delegated authorization. Owner and manager are baked into bytecode (soulbound: no initializer, no transfer). Borrowed and withdrawn funds are delivered to the account and forwarded to the validated receiver (measure-and-forward), so venues that lack a receiver argument are handled uniformly. |
| `MarginAccountFactory` | Deterministic, idempotent, permissionless account deployment (the owner is salt-bound, so a third-party deploy cannot change who owns an account). An abstract mixin inherited by the router, so the router is the manager of every account it deploys. |
| `MorphoLendingAdapter` | Singleton encoder over a governed `(collateral, debt) -> MarketParams` routing table (Morpho Blue). |
| `AaveLendingAdapter` | Singleton encoder over the Aave v3 Pool. Governed `(collateral, debt)` allowlist; the Pool (a stable proxy) is stored immutably, while the protocol data provider is re-resolved from the addresses provider on each use because Aave can repoint it. |
| `AaveV4LendingAdapter` | Singleton encoder over a single Aave v4 **Spoke** (hub-and-spoke). Governed `(collateral, debt) -> (collateralReserveId, debtReserveId)` routes, validated onchain at registration; reads premium-inclusive debt. |
| `CompoundV3LendingAdapter` | Singleton encoder over a single Compound v3 **Comet**. Single-base: the Comet's base token is the only borrowable debt; borrow/repay map onto base `withdraw`/`supply`. The base price feed is read live (it is governance-mutable behind the Comet proxy). |
| `ILendingAdapter` / `IMarginAccount` / `IMarginRouter` | Public surfaces; new lending venues are added behind new adapters with no router or account changes. |
| Value types | `Market` (the pair), `Ltv` (WAD ratio), `MarketRegistry`, `MarketAllowlist`, `Owner`, `PositionData`. |

Adapters are encoders: each returns the `(target, value, callData)` the account executes and holds no funds. The encode surface is `encodeSupplyCollateral`, `encodeEnableCollateral` (run by the account after every supply, for venues needing an explicit collateral enable; empty calldata is the skip signal), `encodeWithdrawCollateral`, `encodeBorrow`, and `encodeRepay`.

Entry points: `increasePosition` (opens or adds leverage; payable, equity via Permit2 or native ETH), `decreasePosition` (partial delever or full close via the `type(uint256).max` sentinel), `addCollateral`, and `execute`, the advanced entrypoint that runs an arbitrary caller-composed plan of the same actions (see the guide's section 4.1 for the composition contract). Plus `createAccount`/`accountOf` and the governance surface.

## Key design decisions

- **Soulbound, self-custodial accounts.** Owner and manager are immutable; only the manager (router) or owner can move funds, and only to the manager or owner. An owner-only `execute` escape hatch on the account lets the owner always act directly on the lending protocol, so funds can never be trapped by router or adapter configuration.
- **Venue-agnostic via adapters.** The router and account speak only `ILendingAdapter`. Four venues ship today; the same audited router/account served each new venue unchanged.
- **Uniform venue semantics at the edges.** Collateral enabling is an explicit post-supply hook on every venue (Aave v3/v4 encode their enable calls; Morpho/Compound skip), which also immunizes supplies against Aave's first-supply auto-enable gap. A debt-free repay encodes an empty no-op on every venue, so a generic repay-then-withdraw exit plan always runs.
- **Allowlist gates exposure-increasing flows only.** Supply and borrow require an allowlisted adapter; withdraw, repay, and sweep never do, so a position is always exitable even after its adapter is disallowed.
- **Slippage and fill protection.** Mandatory absolute input caps (`maxDebtIn`/`maxCollateralIn`) are the binding protection; `minHopPriceX36` is an optional per-hop bound priced on the swap's realized output; fills are all-or-nothing via the delta-threshold fill check. A partial decrease requires live debt and gates on the **measured** repay carried out of the unlock, so a mid-transaction external repay cannot turn it into a silent no-op (`NoDebtToRepay`).
- **Events are indexer-first and never block funds.** Every position mutation on every router path emits a `PositionUpdated` resulting-state snapshot; the curated flows add richer delta events. All event-only reads are best-effort: an oracle revert or a mid-transaction state change can skip or zero an event field but can never roll back a completed mutation.
- **Margin opcodes occupy a distinct `0x30` range** (`0x1c`-`0x2f` stays reserved for future core v4 actions), and undefined opcodes revert `UnsupportedAction` regardless of plan state.
- **ERC-20-only markets** (no fee-on-transfer / rebasing); every flow nets to zero under that constraint.
- **Two-step, zero-address-guarded ownership** for router governance and adapter ownership, with lifecycle events at construction, proposal, and acceptance.

## Indexer

`indexer/` ships a Ponder app that reconstructs positions, actions, and PnL from the router's event surface, with a venue truth layer (Morpho and Aave v3 flows and liquidations) for activity the router cannot see (escape-hatch operations, liquidations). Pool attribution parses each margin transaction's receipt for v4 `Swap` logs, so it is correct for any Universal Router the route names; liquidation terminality is decided from the venue's live debt, not stored state. GraphQL and SQL-over-HTTP included.

## Security review and audit

An internal 8-agent adversarial review of the router/account core preceded the external engagement; its findings (opaque zero-amount reverts, balance-based residuals, exit gating, ownership hardening, realized-output price guards, all-or-nothing fills) were fixed in this branch.

**OpenZeppelin audited the full margin suite (Jul 28 - Aug 12, 2026): 33 findings, no High or Critical.** 3 Medium (two reclassified to Low by the auditors after review), 14 Low, 16 Notes. Every finding is either fixed in this branch with regression tests or acknowledged with documented reasoning; the report is committed at [`audits/OpenZeppelin_audit_margin_trading.pdf`](audits/OpenZeppelin_audit_margin_trading.pdf). Highlights of the fix set:

| Area | Hardening |
|---|---|
| Routed-swap fill check (M-03) | Threshold enforces the swap delta, so pre-existing balances cannot mask a short fill |
| Aave collateral enabling (M-02) | Explicit `encodeEnableCollateral` post-supply hook on all venues; immune to dust-aToken griefing |
| Aave data provider (M-01) | Re-resolved on each use instead of cached (Aave can repoint it) |
| Morpho boundary repays (L-01) | Exact-debt and debt-free repays no longer underflow; partial decreases gate on the measured repay |
| Event reads (L-05) | Best-effort with saturated deltas; a reverting read or hostile-route state change never rolls back a mutation |
| ROUTE_SWAP allowances (L-08) | The scoped Permit2 allowance to the caller-named UR is zeroed after the swap |
| Comet base price feed (L-14) | Read live per valuation instead of cached |
| Observability (N-12 and others) | Owner-lifecycle events, `PositionUpdated` on every path, documented per-venue event field semantics |

A follow-up Codex review of the indexer surfaced three findings (swap-sender filtering, liquidation terminality, requested-vs-measured repay deltas), all fixed and verified against live mainnet data.

## Testing

Unit and fuzz suites for all four adapters, the account, the factory, the value types, and the router; margin integration tests against a real local PoolManager (Permit2 and native-ETH equity, short fills, event-read failure modes, mid-transaction interference); a stateful invariant suite for router/account accounting; and mainnet-fork lifecycle tests per venue (Morpho e2e and multi-borrower, Aave v3, Aave v4, Compound v3) plus hedging (cross-venue, same-venue, shared-account via `subId`), `execute` plans, and routed swaps. Gas snapshots are committed in isolate mode.

## Deployment

The suite is **live on mainnet** (2026-08-12 redeploy): `MarginRouter` at `0x000000000075e82F7B7DdC5DD1B4984b560eF5D4` (mined vanity, CREATE2), with the four adapters allowlisted and the canonical markets registered. The full address registry, with provenance, is in the guide's section 10 and `indexer/addresses.ts`. Governance is currently the deploy key and should be handed to the timelock or a multisig via the two-step transfer before broader rollout.

## New dependencies

Adds the `morpho-blue` and `solady` submodules. Aave v3, Aave v4, and Compound v3 are integrated via minimal vendored interfaces under `src/interfaces/external/` (no new submodules); the Universal Router dependency was already vendored.


src/MarginRouter.sol
```diff
@@ -0,0 +1,521 @@
...
+    }
+
+    /// @notice The authenticated caller for the current lock. Overrides `BaseActionsRouter.msgSender`
+    ///         to return the address stored by `ReentrancyLock._getLocker`, which is set to
+    ///         `msg.sender` at the start of each `isNotLocked` call. The active account is derived
+    ///         from this value, so correctness here is load-bearing for the entire position system.
+    function msgSender() public view override returns (address) {
+        return _getLocker();
+    }
+
```
zhongeric, line 415: Is there something we can borrow from v4 core here? noting the nested locking funtionality

src/MarginRouter.sol
```diff
@@ -0,0 +1,521 @@
...
+        isNotLocked
+        checkDeadline(params.deadline)
+        returns (address account)
+    {
+        account = _open(params);
+        emit PositionIncreased(
+            msgSender(), account, params.market.collateral, params.market.debt, params.collateralToBuy
+        );
+    }
+
```
zhongeric, line 153: Is there a specific reason you want to expose both open and increasePosition? Seem to be doing the same thing under the hood (unlke mint/increase from v4)

src/MarginRouter.sol
```diff
@@ -0,0 +1,521 @@
...
+    /// @notice Shared implementation for `openPosition` and `increasePosition`. Deploys the account
+    ///         if needed, pulls optional equity, then builds and runs the flash-style unlock:
+    ///         swap debt to collateral (exact-output), supply the collateral, borrow the debt owed,
+    ///         and settle the swap.
+    /// @param params The open/increase parameters; see `OpenParams`.
+    /// @return account The caller's MarginAccount address.
+    function _open(OpenParams calldata params) private returns (address account) {
+        // a zero buy would feed a zero amount into the exact-output swap, which the PoolManager rejects
+        if (params.collateralToBuy == 0) revert SlippageBoundRequired();
+        if (params.maxDebtIn == 0) revert SlippageBoundRequired();
```
zhongeric: nit could be combined?

src/MarginRouter.sol
```diff
@@ -0,0 +1,521 @@
...
+    }
+
+    /// @notice Stores the active account for the current unlock in transient storage (EIP-1153).
+    ///         It is always derived from the authenticated caller stored by `ReentrancyLock`, never
+    ///         from calldata, so margin handlers operate only on the caller's own account.
+    /// @param account The MarginAccount address to store, or `address(0)` to clear after unlock.
+    function _setActiveAccount(address account) private {
+        bytes32 slot = ACTIVE_ACCOUNT_SLOT;
+        assembly ("memory-safe") {
+            tstore(slot, account)
```
zhongeric, line 798: nit these would be cleaner imo as library functions instead of private

src/MarginRouter.sol
```diff
@@ -0,0 +1,521 @@
...
+            _wrap(msg.value);
+            Currency.wrap(address(WETH9)).transfer(account, msg.value);
+        } else if (params.equity > 0) {
+            permit2.transferFrom(
+                msgSender(), account, params.equity.toUint160(), Currency.unwrap(params.market.collateral)
+            );
+        }
+
+        // single choke point: validate the pool matches the market and derive the swap direction.
+        // opening sells the debt to buy the collateral.
```
zhongeric: rewrite this comment

src/MarginRouter.sol
```diff
@@ -0,0 +1,521 @@
...
+        // assert the exact-output swap fully filled: a v4 swap can partial-fill on a thin pool (the
+        // price hits the global limit before the full output is bought). Without this the open would
+        // take only the realized amount and silently open a smaller position. Asserting the router
+        // holds the full collateralToBuy credit makes the open all-or-nothing with a clear error.
+        actionParams[1] = abi.encode(params.market.collateral, uint256(params.collateralToBuy));
+        // take the bought collateral to the account (OPEN_DELTA == the full fill the assert just proved)
+        actionParams[2] = abi.encode(params.market.collateral, account, ActionConstants.OPEN_DELTA);
+        // supply the account's full collateral balance (equity + bought)
+        actionParams[3] = abi.encode(params.adapter, params.market, uint256(ActionConstants.OPEN_DELTA));
+        // borrow the debt owed for the swap, sent to the router for settling
```
zhongeric: cleaner as a modifier given you have to clear the slot?

src/MarginRouter.sol
```diff
@@ -0,0 +1,521 @@
...
+        if (action < MarginActions.ACCOUNT_SUPPLY_COLLATERAL) {
+            super._handleAction(action, params);
+            return;
+        }
+
+        address account = _activeAccount();
+        if (action == MarginActions.ACCOUNT_SUPPLY_COLLATERAL) {
+            (ILendingAdapter adapter, Market memory market, uint256 amount) = params.decodeAdapterMarketAmount();
+            if (amount == 0) amount = IERC20(Currency.unwrap(market.collateral)).balanceOf(account);
+            IMarginAccount(account).supplyCollateral(adapter, market, amount);
```
zhongeric: what is OPEN_DELTA? is that 0? if so use that constant in the equality check here

src/MarginRouter.sol
```diff
@@ -0,0 +1,521 @@
...
+            (ILendingAdapter adapter, Market memory market, uint256 amount, address to) =
+                params.decodeAdapterMarketAmountReceiver();
+            if (amount == 0) amount = _getFullDebt(market.debt);
+            IMarginAccount(account).borrow(adapter, market, amount, to);
+        } else if (action == MarginActions.ACCOUNT_REPAY) {
+            (ILendingAdapter adapter, Market memory market, uint256 amount) = params.decodeAdapterMarketAmount();
+            IMarginAccount(account).repay(adapter, market, amount);
+        } else if (action == MarginActions.ACCOUNT_SWEEP) {
+            (Currency currency, uint256 amount, address to) = params.decodeSweep();
+            IMarginAccount(account).sweep(currency, amount, to);
```
zhongeric: Not sure what this is used for

src/interfaces/ILendingAdapter.sol
```diff
@@ -0,0 +1,115 @@
...
+    /// @param market The (collateral, debt) pair identifying the target lending market.
+    /// @param amount The amount of collateral to supply, in the collateral token's native decimals.
+    /// @return target The call target (always `lendingProtocol()`).
+    /// @return value The call value. Always 0 for Morpho (non-payable); the account treats non-zero
+    ///         as a bug.
+    /// @return callData The calldata the account executes against `target`.
+    function encodeSupplyCollateral(address account, Market calldata market, uint256 amount)
+        external
+        view
+        returns (address target, uint256 value, bytes memory callData);
```
zhongeric, line 46: Consider tradeoffs of using a Call struct here. Especially given you don't support nonzero value

src/MarginRouter.sol
ccashwell: fair point, they really are the same thing just with different names. earlier on i was keeping track of positions and increase only worked on existing positions but for most lending protocols there is no difference so they ended up converging. can remove.

src/MarginRouter.sol
ccashwell: essentially funds are hanging out in the account after some operations and whether that's where it should live or it should be swept to the caller or some other address is a user preference that is best expressed explicitly imo

src/types/Owner.sol
```diff
@@ -0,0 +1,90 @@
...
+///         Composed by contracts that need governance gating; for example, the lending adapter gates
+///         `setMarket` with `store.owner.onlyOwner(msg.sender)`. Handoffs are two-step: the current
+///         owner `propose`s a successor, who must then `acceptOwnership`. A zero-address successor is
+///         rejected, so the role can never be transferred to an unrecoverable address.
+/// @param _inner The current owner address. Read via `read()`; mutated via `write()` and
+///        `acceptOwnership()`.
+/// @param _pending The address proposed as the next owner, or `address(0)` when none is pending.
+///        Read via `pendingOwner()`; set via `propose()`; cleared on `acceptOwnership()`.
+struct Owner {
+    address _inner;
```
hensha256: nit but why not `_current`? Feels easier to read `Owner._current`

src/MarginRouter.sol
hensha256: yeah OPEN_DELTA is 0

src/MarginRouter.sol
```diff
@@ -0,0 +1,521 @@
...
+    function _handleAction(uint256 action, bytes calldata params) internal override {
+        if (action < MarginActions.ACCOUNT_SUPPLY_COLLATERAL) {
+            super._handleAction(action, params);
+            return;
+        }
+
+        address account = _activeAccount();
+        if (action == MarginActions.ACCOUNT_SUPPLY_COLLATERAL) {
+            (ILendingAdapter adapter, Market memory market, uint256 amount) = params.decodeAdapterMarketAmount();
+            if (amount == 0) amount = IERC20(Currency.unwrap(market.collateral)).balanceOf(account);
```
hensha256: theres a currency library so could just write
`market.collateral.balanceOf(account)` without the unwrapping and casting

src/MarginRouter.sol
zhongeric: yeah and seeing as you only conditionally deploy a new margin account seems fine to just condense these

src/MarginAccount.sol
```diff
@@ -0,0 +1,194 @@
...
+    ///         no value, then performs a regular call (never a delegatecall). Reverts on failure via
+    ///         `Address.functionCall`.
+    /// @dev The target and value checks are defense in depth: the manager only routes allowlisted
+    ///      adapters, and the owner deliberately chooses the adapter for their own funds.
+    /// @param adapter The adapter whose `lendingProtocol()` is the permitted target.
+    /// @param target The encoded call target; must equal `adapter.lendingProtocol()`.
+    /// @param value The encoded call value; must be zero.
+    /// @param callData The calldata to forward.
+    /// @return The raw bytes returned by the lending protocol call.
+    function _execCall(ILendingAdapter adapter, address target, uint256 value, bytes memory callData)
```
zhongeric: You could just omit `target` and call adapter.lendingProtocol()

src/AaveLendingAdapter.sol
```diff
@@ -0,0 +1,287 @@
...
+    /// @notice The Aave v3 protocol data provider, resolved from the addresses provider at
+    ///         construction. Used to read reserve token addresses and reserve configuration.
+    IPoolDataProvider public immutable dataProvider;
+
+    /// @notice Internal storage for the market allowlist and the owner guard.
+    /// @param allowed The governed allowlist mapping `(collateral, debt)` to whether the pair is
+    ///        routable. Managed via `setMarket`.
+    /// @param owner The current adapter owner, gating `setMarket` and `transferOwnership`.
+    struct AdapterStore {
+        mapping(Currency collateral => mapping(Currency debt => bool)) allowed;
```
zhongeric: keccak256(collateral, debt) => bool?

src/AaveLendingAdapter.sol
```diff
@@ -0,0 +1,287 @@
...
+        Owner owner;
```
zhongeric: Could be separated unless there's a specific reason for the AdapterStore struct

src/AaveLendingAdapter.sol
```diff
@@ -0,0 +1,287 @@
...
+    ///         until then, and the zero address is rejected so the role cannot be bricked. Owner-gated.
+    /// @param newOwner The address proposed to become the new owner.
+    function transferOwnership(address newOwner) external {
+        store.owner.onlyOwner(msg.sender);
+        store.owner.propose(newOwner);
+    }
+
+    /// @notice Reverts `MarketNotSupported` unless the `(collateral, debt)` pair is allowlisted.
+    /// @param market The market pair to check.
+    function _require(Market calldata market) internal view {
```
zhongeric: Could be renamed to be more descriptive. Feels weird to have an internal func called require

src/types/Market.sol
```diff
@@ -0,0 +1,77 @@
...
+function hasCurrencies(Market memory self, Currency a, Currency b) pure returns (bool) {
+    return (a == self.collateral && b == self.debt) || (a == self.debt && b == self.collateral);
+}
+
+/// @notice The single choke point that reconciles a v4 pool with this market and derives swap
+///         direction. Every margin flow builds its swap params here, so a pool/market currency
+///         mismatch is structurally unable to reach the lending legs.
+/// @dev This is a runtime check (Solidity cannot make set-equality a compile-time guarantee); its
+///      value is locality and unavoidability: every swap passes through this function and it cannot
+///      be bypassed.
```
zhongeric:
```suggestion
```

src/types/Market.sol
```diff
@@ -0,0 +1,77 @@
...
+/// @return params Validated v4 `SwapParams`; `zeroForOne` is derived from `input` and pool
+///         ordering.
+function toSwapParams(
+    Market memory self,
+    Currency input,
+    int256 amountSpecified,
+    uint160 sqrtPriceLimitX96,
+    PoolKey memory key
+) pure returns (SwapParams memory params) {
+    // the pool must trade exactly this market's two currencies (order-independent)
```
zhongeric:
```suggestion
    // the pool must trade exactly this market's two currencies
```

src/MorphoLendingAdapter.sol
```diff
@@ -0,0 +1,229 @@
...
+    ///      with `onBehalf = account` and no callback data. The `value` field is always 0 because
+    ///      Morpho Blue is non-payable.
+    function encodeSupplyCollateral(address account, Market calldata market, uint256 amount)
+        external
+        view
+        returns (address, uint256, bytes memory)
+    {
+        MarketParams memory marketParams = store.markets.resolve(market);
+        bytes memory data;
+        return (address(morpho), 0, abi.encodeCall(IMorphoBase.supplyCollateral, (marketParams, amount, account, data)));
```
zhongeric: can't empty bytes be defined inline

src/MorphoLendingAdapter.sol
```diff
@@ -0,0 +1,229 @@
...
+    /// @dev Computes the current LTV as `debt * WAD / collateralValue`, where `collateralValue` is
+    ///      the oracle's price of collateral quoted in the loan token (1e36-scaled). Returns
+    ///      `type(uint256).max` (as an `Ltv`) when there is debt but zero collateral value (fully
+    ///      undercollateralized). Returns 0 when there is no debt.
+    function currentLtvWad(address account, Market calldata market) external view returns (Ltv) {
+        MarketParams memory marketParams = store.markets.resolve(market);
+        uint256 collateral = uint256(morpho.position(marketParams.id(), account).collateral);
+        uint256 debt = morpho.expectedBorrowAssets(marketParams, account);
+        // collateral value expressed in loan-token units
+        uint256 collateralValue = collateral * IOracle(marketParams.oracle).price() / ORACLE_PRICE_SCALE;
```
zhongeric: haven't looked at morpho in detail but there is risk of overflow here, `.price()` says it returns values scaled by 1e36

src/MorphoLendingAdapter.sol
```diff
@@ -0,0 +1,229 @@
...
+
+    /// @notice Emitted when a market is registered or replaced in the routing table. Includes
+    ///         oracle, IRM, and LLTV so offchain monitoring can vet the routed market configuration.
+    /// @param collateral The collateral token address of the registered market.
+    /// @param debt The debt (loan) token address of the registered market.
+    /// @param id The Morpho Blue market id derived from the `MarketParams`.
+    /// @param oracle The price oracle address for this Morpho market.
+    /// @param irm The interest rate model address for this Morpho market.
+    /// @param lltv The liquidation LTV for this Morpho market (WAD, 1e18 == 100%).
+    event MarketSet(address indexed collateral, address indexed debt, Id id, address oracle, address irm, uint256 lltv);
```
zhongeric: index `id`? also worth putting that first?

src/MorphoLendingAdapter.sol
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /// @notice Begins a two-step ownership handoff by proposing a successor. The successor takes
+    ///         effect only once it calls `acceptOwnership`; the current owner retains its powers
+    ///         until then, and the zero address is rejected so the role cannot be bricked. Owner-gated.
+    /// @param newOwner The address proposed to become the new owner.
+    function transferOwnership(address newOwner) external {
+        store.owner.onlyOwner(msg.sender);
+        store.owner.propose(newOwner);
+    }
```
zhongeric: this func is repeated in each lending adapter, worth just putting in a base shared contract

src/types/Owner.sol
```diff
@@ -0,0 +1,90 @@
...
+
+/// @notice Proposes a successor for the two-step handoff. The successor takes effect only once it
+///         calls `acceptOwnership`; the current owner remains authorized until then. Reverts
+///         `ZeroOwner` if `newOwner` is the zero address. The caller must gate access (e.g. by
+///         calling `onlyOwner` first); this function performs no authorization itself.
+/// @param self The `Owner` storage to update.
+/// @param newOwner The address proposed as the next owner.
+/// @return The same storage reference, for chaining.
+function propose(Owner storage self, address newOwner) returns (Owner storage) {
+    if (newOwner == address(0)) revert ZeroOwner();
```
hensha256, line 71: imo being able to pass in `address(0)` is a valuable feature ot be able to reverse the proposal of an owner? if you pass in `address(0)` by accident its not like its a problem

hensha256: the top of the contract says
```
A zero-address successor is
///         rejected, so the role can never be transferred to an unrecoverable address.
```

but i dont understand how proposing the 0 address would ever actually transfer to an unrecoverable address? the whole point in 2 step proposals is that they have to accept

src/types/Owner.sol
```diff
@@ -0,0 +1,90 @@
...
+    return self;
+}
+
+/// @notice Completes the two-step handoff. Reverts `NotPendingOwner` unless a successor is pending
+///         and `caller` is that successor. On success the pending owner becomes the current owner and
+///         the pending slot is cleared.
+/// @param self The `Owner` storage to update.
+/// @param caller The address accepting ownership; typically `msg.sender`.
+function acceptOwnership(Owner storage self, address caller) {
+    if (self._pending == address(0) || caller != self._pending) revert NotPendingOwner(caller);
```
hensha256:
```suggestion
    if (caller != self._pending) revert NotPendingOwner(caller);
```

src/MarginAccount.sol
```diff
@@ -0,0 +1,194 @@
...
+    ///      caller (the account), which forwards the measured delta here. Either way the recipient stays
+    ///      under the account's control. `withdrawn` is measured as the account's collateral-token
+    ///      balance increase across the call, so it is zero when the protocol delivered directly to `to`.
+    function withdrawCollateral(ILendingAdapter adapter, Market calldata market, uint256 amount, address to)
+        external
+        returns (uint256 withdrawn)
+    {
+        (address ownerAddr, address managerAddr) = _authCaller();
+        _requireReceiver(to, ownerAddr, managerAddr);
+        IERC20 collateral = IERC20(Currency.unwrap(market.collateral));
```
hensha256: i dont think you need to cast to a token, all these functions are available on the currency type

or if everything in this protocol _has_ to be a token (not native), then it feels better to use the ERC20 type by default not currency?

src/MarginRouter.sol
```diff
@@ -0,0 +1,521 @@
...
+/// @author Uniswap Labs
+/// @notice The entry point for the margin-trading suite. Composes the v4 action machinery and
+///         builds each leveraged position as a single flash-style swap inside one PoolManager
+///         unlock: borrow the debt, swap it into collateral, supply the collateral, then draw the
+///         debt back to settle. Each operation runs against the caller's own MarginAccount, derived
+///         from the authenticated caller (never from a caller-supplied address). The router is the
+///         manager of every account it deploys, so it can drive their lending primitives.
+///
+///         Supported markets are restricted to the governance allowlist of lending adapters, which
+///         curate standard ERC-20 markets only (Morpho Blue does not support fee-on-transfer or
```
mgretzke: "Morpho Blue does not support fee-on-transfer..." 
Morpho specific statements should probably land in the Morpho adapter, while in the MarinRouter, it should stay more generalized

src/MarginRouter.sol
```diff
@@ -0,0 +1,521 @@
...
+    /// @dev Reverts `DeadlinePassed` if `block.timestamp` has passed `deadline`.
+    modifier checkDeadline(uint256 deadline) {
+        if (block.timestamp > deadline) revert DeadlinePassed(deadline);
+        _;
+    }
+
+    /// @notice Deploys the margin router.
+    /// @param poolManager_ The v4 PoolManager singleton the router unlocks for every position flow.
+    /// @param permit2_ The Permit2 contract used to pull caller equity and settle swaps.
+    /// @param weth9_ The canonical WETH9 contract used to wrap native ETH equity.
```
mgretzke: nit: saying "wrap native token equity" instead of naming ETH directly makes this naming more in line with non mainnet chains

indexer/ponder.config.ts
```diff
@@ -0,0 +1,95 @@
...
+    },
+    /** Margin swaps only: the router is always the swap sender for position flows. */
+    PoolManagerSwaps: {
+      abi: poolManagerSwapAbi,
+      chain: "mainnet",
+      address: mainnet.poolManager,
+      startBlock: mainnet.startBlock,
+      filter: {
+        event: "Swap",
+        args: { sender: mainnet.marginRouter },
```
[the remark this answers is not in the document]
ccashwell: Fixed in 724b6ca: the sender-filtered stream is gone; the router handlers fetch the margin transaction's receipt and record the PoolManager Swap logs it contains, so swaps are attributed on every execution path regardless of which Universal Router the route named. Verified against live mainnet data: the previously-null openPoolId now resolves.

indexer/src/lendingFlows.ts
```diff
@@ -0,0 +1,229 @@
...
+    timestamp: liq.timestamp,
+    applied: true,
+  });
+
+  const live = await findActivePosition(context, liq.account, liq.collateral, liq.debt);
+  if (!live) return;
+
+  const debtCleared = liq.repaidDebt + liq.badDebt;
+  const remainingDebt = clamp0(live.debtPrincipal - debtCleared);
+  const fullyLiquidated = remainingDebt === 0n;
```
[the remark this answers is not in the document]
ccashwell: Fixed in 7ccf5a2: recordLiquidation now reads the venue's live remaining debt at the liquidation block (Morpho borrow shares via the virtual-offset share math, Aave v3 variable-debt balance via the data provider) and derives both the refreshed principal and the terminal decision from it, falling back to the stored-principal arithmetic only when the venue is unreadable.

indexer/src/router.ts
```diff
@@ -0,0 +1,425 @@
...
+    id: eventId(event.transaction.hash, event.log.logIndex),
+    positionId: row.id,
+    type: isClose ? "CLOSE" : "DECREASE",
+    txHash: event.transaction.hash,
+    logIndex: event.log.logIndex,
+    blockNumber: event.block.number,
+    timestamp: event.block.timestamp,
+    collateralDelta: -collateralWithdrawn,
+    debtDelta: -debtRepaid,
+    equityDelta: isClose ? -row.equity : 0n,
```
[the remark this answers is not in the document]
ccashwell: Fixed in 4a6a9af: drainFlows aggregates the measured repay from the same-tx venue flows and the decrease handler prefers it for the action's debtDelta and priceX18. Aave v4 and Compound stage no flow layer yet and fall back to the requested amount; the position row itself already stored the event's measured resulting totals.

src/libraries/MarginActions.sol
```diff
@@ -0,0 +1,84 @@
...
+
+    /// @notice Asserts that the position's current LTV does not exceed a caller-supplied bound,
+    ///         reverting `PositionUnhealthy` if it does. Decoded with
+    ///         `MarginCalldataDecoder.decodeHealthCheck`.
+    uint256 internal constant ASSERT_HEALTH = 0x35;
+
+    /// @notice Asserts that the router holds at least a minimum credit in a currency, i.e. an
+    ///         exact-output swap delivered the full requested amount, reverting `IncompleteFill` on a
+    ///         partial fill. Decoded with `MarginCalldataDecoder.decodeFillCheck`.
+    uint256 internal constant ASSERT_FILL = 0x36;
```
zhongeric, line 47: Do we still need `ASSERT_FILL`?

Nothing in `src/` emits this opcode. The curated plans (`increasePosition` / `decreasePosition`) use `ASSERT_ACCOUNT_BALANCE` after `ROUTE_SWAP`, because the bought output is delivered to the account, not as router credit. The only in-tree consumers are `execute` tests that follow a native `SWAP_EXACT_OUT` with this check.

That path already has an all-or-nothing guarantee on the swap itself (`V4ExactOutputUnfilled` in `V4Router`). `ASSERT_FILL` then re-reads cumulative `_getFullCredit`, which is a slightly different predicate and is easy to confuse with `ASSERT_ACCOUNT_BALANCE`.

If `execute` is the intended consumer, worth calling that out — or folding a min-credit check into `TAKE` / exact-out so we do not grow an `ASSERT_*` family that the product flows never use. `ASSERT_HEALTH` and `ASSERT_ACCOUNT_BALANCE` earn their slots because of where they have to sit in the plan; this one is less clear.

src/AaveLendingAdapter.sol
```diff
@@ -0,0 +1,340 @@
...
+///           router-enforced invariant: each Aave position must use its own `(owner, subId)` account.
+///           Co-locating two of this Pool's markets under one `subId` blends the reads and can make a
+///           close/decrease revert or withdraw collateral still backing another debt. Use a distinct
+///           `subId` per Aave position (Morpho markets are isolated and not subject to this).
+///         - Routing is curated: every `encode*` and read reverts `MarketNotSupported` for a pair the
+///           owner has not allowlisted, never returning a silent default market.
+/// @custom:security-contact security@uniswap.org
+contract AaveLendingAdapter is ILendingAdapter, OwnableAdapter, PositionAmountResolver {
+    // WAD scale for loan-to-value ratios.
+    uint256 private constant WAD = 1e18;
```
zhongeric: WAD and BPS are duplicated everywhere

src/libraries/MarginActions.sol
ccashwell: Yeah this is mainly for directly composed `execute` plans to have per-step control over the outcome of their swap paths which could touch any combination of v2/v3/v4 pools. This was added before V4 router guaranteed exactOut fills and I felt that leaving it still gives some benefit to plan builders even with router level guarantees in place.

src/interfaces/ILendingAdapter.sol
```diff
@@ -0,0 +1,165 @@
...
+    ///         `callData`.
+    /// @dev The account executes `callData` against `target` only when `callData` is non-empty;
+    ///      an empty `callData` is the skip signal. Moves no tokens (no approval needed). Idempotent
+    ///      where implemented, so it is safe to run on every supply, including a top-up.
+    /// @param account The MarginAccount whose reserve is enabled as collateral.
+    /// @param market The (collateral, debt) pair identifying the target lending market.
+    /// @return target The call target (`lendingProtocol()`) when `callData` is non-empty.
+    /// @return value The call value. Always 0 for the in-tree non-payable lending protocols.
+    /// @return callData The calldata to execute, or empty bytes to skip.
+    function encodeEnableCollateral(address account, Market calldata market)
```
zhongeric: The naming here seems odd to me given that "enabling" is a concept local to here. What about `afterSupplyCollateral`

src/MarginAccount.sol
```diff
@@ -0,0 +1,210 @@
...
+        (address target, uint256 value, bytes memory callData) =
+            adapter.encodeSupplyCollateral(address(this), market, amount);
+        _setApproval(market.collateral, target, amount);
+        _execCall(target, value, callData);
+        _setApproval(market.collateral, target, 0);
+        // Run the venue's post-supply collateral-enable (Aave needs an explicit, account-scoped
+        // enable; Morpho/Compound return empty and this is skipped). Moves no tokens, so no approval.
+        (address enableTarget, uint256 enableValue, bytes memory enableData) =
+            adapter.encodeEnableCollateral(address(this), market);
+        if (enableData.length != 0) _execCall(enableTarget, enableValue, enableData);
```
zhongeric: Consider the scope of pre/post call actions required by future lending protocols. Is it worth more explcitily defining a pattern here and moving this conditional logic into the adapters instead of here?

One clear limitation of the current adapter design is that they are purely _encoders_, where it's clear that each lending protocol may require a chain of encoded actions and/or intermediate state.

src/MarginAccount.sol
```diff
@@ -0,0 +1,210 @@
...
+    ///      borrowed asset is always delivered to the account because lending protocols differ on
+    ///      whether `borrow` exposes a receiver; forwarding here keeps the recipient under the
+    ///      account's control regardless. `borrowed` is measured as the debt token balance increase
+    ///      (mirroring `repay`), which is accurate across protocols rather than trusting `amount`.
+    function borrow(ILendingAdapter adapter, Market calldata market, uint256 amount, address to)
+        external
+        returns (uint256 borrowed)
+    {
+        (address ownerAddr, address managerAddr) = _authCaller();
+        _requireReceiver(to, ownerAddr, managerAddr);
```
zhongeric: Kind of a nit so feel free to ignore, but something like `to.requireIsReceiver()` is much easier for me to read.

src/MarginAccount.sol
```diff
@@ -0,0 +1,210 @@
...
+        uint256 balanceBefore = market.debt.balanceOfSelf();
```
zhongeric: Flagging Arc ERC20/native USDC weirdness may be in play here

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        address accountImplementation_,
+        address governance_
+    )
+        V4Router(poolManager_)
+        Permit2Forwarder(permit2_)
+        NativeWrapper(weth9_)
+        MarginAccountFactory(accountImplementation_)
+    {
+        // governance is set explicitly so CREATE2 deployment names the intended owner, not the
+        // CREATE2 factory; hand off to a timelock or multisig after setup
```
zhongeric: rm this comment

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+    {
+        uint256 debtBefore;
+        (account, debtBefore) = _increase(params);
+        // Best-effort rich event. describePosition consults the venue oracle, which can revert (e.g.
+        // oracle downtime); the increase is already complete and the PositionUpdated snapshots emitted
+        // inside the unlock carry the resulting state, so a failing read must not roll back the
+        // mutation. Skip the delta-carrying event rather than revert. The try/catch guards only the
+        // external call, so nothing in this success block may revert either: the debt delta saturates
+        // to zero because venues accept permissionless onBehalf repays, and one landing inside the
+        // transaction (from code in the route path) can leave the debt BELOW its pre-increase level.
```
zhongeric: simplify/shorten this please

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+    /// @param params The increase parameters; see `IncreaseParams`.
+    /// @return account The caller's MarginAccount address.
+    /// @return debtBefore The account's debt before the increase; the caller derives debtDrawn as the
+    ///         post-increase debt minus this, correct for both a fresh open and an increase.
+    function _increase(IncreaseParams calldata params) private returns (address account, uint256 debtBefore) {
+        // a zero buy would feed a zero amount into the exact-output swap, which the PoolManager rejects
+        if (params.collateralToBuy == 0) revert ZeroAmount();
+        if (params.maxDebtIn == 0) revert SlippageBoundRequired();
+        // maxLtvAfter stays optional here (zero skips the check), but a supplied bound must be able to
+        // bind: reject >= 100%, which would otherwise read as "set" yet disable ASSERT_HEALTH
```
zhongeric: remove or simplfiy this comment

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        _setActiveAccount(account);
+
+        // snapshot debt before drawing leverage so debtDrawn is correct whether this opens a fresh
+        // position or increases an existing one
+        (, debtBefore) = params.adapter.positionOf(account, params.market);
+
+        // provide equity: native ETH (wrapped to WETH) when sent, else ERC20 pulled via Permit2
+        if (msg.value > 0) {
+            if (Currency.unwrap(params.market.collateral) != address(WETH9)) revert NativeCollateralMismatch();
+            _wrap(msg.value);
```
zhongeric: This will break for Arc and any other chain where WETH9 is not supported.

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+            Currency.wrap(address(WETH9)).transfer(account, msg.value);
+        } else if (params.equity > 0) {
+            permit2.transferFrom(
+                msgSender(), account, params.equity.toUint160(), Currency.unwrap(params.market.collateral)
+            );
+        }
```
zhongeric: Defense in depth - revert if no value AND equity is 0?

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+
+        bytes memory actions = abi.encodePacked(
+            uint8(MarginActions.ROUTE_SWAP),
+            uint8(MarginActions.ASSERT_ACCOUNT_BALANCE),
+            uint8(MarginActions.ACCOUNT_SUPPLY_COLLATERAL),
+            uint8(MarginActions.ACCOUNT_BORROW),
+            uint8(Actions.SETTLE),
+            uint8(MarginActions.ASSERT_HEALTH)
+        );
+        bytes[] memory actionParams = new bytes[](6);
```
zhongeric: nit: usually call this `params`

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        // route the debt->collateral swap through the Universal Router with the caller-supplied route,
+        // behind a flash-take of up to maxDebtIn debt. The route buys collateralToBuy exact-output and
+        // delivers it to the account, and ROUTE_SWAP settles the unspent take, so the router's remaining
+        // debt equals exactly what the swap spent (the same negative delta a native v4 swap would leave).
```
zhongeric: would be great to make these comments concisely: 
```
// ACTION abi
// Brief description...
```

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        // they never reach the NoActiveAccount revert.
+        if (action > MarginActions.ASSERT_ACCOUNT_BALANCE) revert UnsupportedAction(action);
+        address account = _activeAccount();
+        if (account == address(0)) revert NoActiveAccount();
+        _handleAccountAction(action, params, account);
+    }
+
+    /// @notice Handles an opcode below the margin range. V4Router handles swaps and settle/take; this
+    ///         router additionally intercepts the contiguous `SWEEP`/`WRAP`/`UNWRAP` trio (asserted
+    ///         contiguous in MarginCalldataDecoder.t.sol), which V4Router would reject, with
```
zhongeric: clean up this comment

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+            if (action == Actions.SWEEP) {
+                (Currency currency, address to) = params.decodeCurrencyAndAddress();
+                _sweep(currency, _mapRecipient(to));
+            } else if (action == Actions.WRAP) {
+                _wrap(
+                    _mapWrapUnwrapAmount(
+                        CurrencyLibrary.ADDRESS_ZERO, params.decodeUint256(), Currency.wrap(address(WETH9))
+                    )
+                );
+            } else {
```
zhongeric: Defense in depth: `else if (action == Actions.UNWRAP)` and revert in the default else branch

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        IMarginAccount(account).supplyCollateral(params.adapter, params.market, amount);
+        // Best-effort events. addCollateral runs no unlock, so both the PositionUpdated snapshot and
+        // the curated CollateralAdded come from this describePosition read, which consults the venue
+        // oracle and can revert (e.g. oracle downtime; a debt-free Morpho position still triggers an
+        // oracle read). The supply has already completed, and this is a risk-REDUCING top-up, so a
+        // failing read must not roll it back: emit both on success, skip both on failure.
+        // PositionUpdated carries the full pair and maxLtv, which CollateralAdded does not, so
+        // snapshot-only consumers can attribute the supply without pair-resolution heuristics.
+        try params.adapter.describePosition(account, params.market) returns (PositionData memory position) {
+            emit PositionUpdated(
```
zhongeric: could use `_emitPosition` here somehow?

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        _emitPosition(adapter, market, account);
+    }
+
+    /// @notice Repays debt to the lending protocol. Not allowlist-gated. `type(uint256).max` repays
+    ///         the full debt, resolved by the adapter in the venue's own terms (see
+    ///         `ILendingAdapter.encodeRepay`).
+    function _repay(bytes calldata params, address account) private {
+        (ILendingAdapter adapter, Market memory market, uint256 amount) = params.decodeAdapterMarketAmount();
+        uint256 repaid = IMarginAccount(account).repay(adapter, market, amount);
+        // recorded transiently so the curated partial decrease can assert the repay executed once
```
zhongeric: Don't know what `Curated partial decrease` means here. Rewrite to plain english, ex. `Cache the repaid amount in transient storage which is checked to be non zero at the end of the call`

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        external
+        isNotLocked
+        checkDeadline(params.deadline)
+        returns (address account)
+    {
+        bool fullClose = params.debtToRepay == type(uint256).max;
+
+        // a partial decrease's bounds don't depend on the position, so validate them before any
+        // external read: a non-zero repay (a zero would feed a zero into the exact-output swap, which
+        // the PoolManager rejects) and a resulting-health bound so it cannot worsen the LTV
```
zhongeric: Rewrite to clearly state that the repayment has to be nonzero, and the maxLTV after must not be worse.

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        if (!fullClose) {
+            if (params.debtToRepay == 0) revert ZeroAmount();
+            if (Ltv.unwrap(params.maxLtvAfter) == 0) revert SlippageBoundRequired();
+            // reject a bound that cannot bind (>= 100%): it satisfies the require above yet leaves
+            // ASSERT_HEALTH a no-op, so the "mandatory" partial-decrease health bound stays real
+            if (Ltv.unwrap(params.maxLtvAfter) >= WAD) revert IneffectiveLtvBound(params.maxLtvAfter);
```
zhongeric: This seems unecessary

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        (uint256 collateralBefore, uint256 debt) = params.adapter.positionOf(account, params.market);
+
+        // a full close of a debt-free position (funded only via addCollateral, repaid out of band, or
+        // fully liquidated) needs no swap: withdraw the collateral straight to the caller and finish.
+        if (fullClose && debt == 0) {
+            if (collateralBefore > 0) {
+                IMarginAccount(account).withdrawCollateral(params.adapter, params.market, collateralBefore, msgSender());
+            }
+            // this path bypasses the unlock interpreter, so no action handler fires the snapshot;
+            // emit the terminal PositionUpdated here so snapshot-only consumers see the close
```
zhongeric: rewrite this comment or remove

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+                type(uint256).max
+            );
+            return account;
+        }
+
+        // a partial decrease of a debt-free position has nothing to repay: the swap would buy debt
+        // tokens no repay consumes, stranding them in the account (only a full close sweeps the
+        // surplus) while the event reported a repay that never happened. Fail loudly instead; this
+        // also keeps a stale partial decrease against a re-pointed pair, whose reads see zero debt,
+        // a revert rather than a silent value leak.
```
zhongeric: `Revert on a partial decrease of a position without debt as there is nothing to repay to prevent swapping into unecessary debt tokens`

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        if (debt == 0) revert NoDebtToRepay();
+
+        // a swap runs from here (the debt-free full close returned above), so the input cap is mandatory
+        if (params.maxCollateralIn == 0) revert SlippageBoundRequired();
```
zhongeric: nit: this error is identical to the one above for maxLTV

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+            uint8(MarginActions.ASSERT_HEALTH)
+        );
+        // the mode-dependent amounts are inlined (kept out of locals to stay under the stack limit).
+        // full close: require the route delivered AT LEAST the current debt (asserted below), repay
+        // ALL (venue-resolved, see `ILendingAdapter.encodeRepay`) so the borrow clears with no dust,
+        // withdraw all collateral, and
+        // pass a zero health bound that ASSERT_HEALTH skips; any debt the route over-bought is swept
+        // back after the unlock. A partial requires and repays exactly `debtToRepay`, withdraws only
+        // the collateral the swap consumed (OPEN_DELTA == the router's remaining collateral debt), and
+        // enforces `maxLtvAfter`.
```
zhongeric: see above comment about these comments for params

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+            params.routeCommands,
+            params.routeInputs
+        );
+        // require the route to deliver the debt the repay needs ON TOP of any debt-token balance the
+        // account already holds going into the unlock. Setting the ASSERT_ACCOUNT_BALANCE threshold to
+        // `balance + required` makes the absolute check enforce the SWAP DELTA, so a pre-existing or
+        // donated balance cannot mask an exact-output under-fill
+        uint256 debtHeldBefore = params.market.debt.balanceOf(account);
+        actionParams[1] = abi.encode(params.market.debt, debtHeldBefore + (fullClose ? debt : params.debtToRepay));
+        actionParams[2] = abi.encode(params.adapter, params.market, fullClose ? type(uint256).max : params.debtToRepay);
```
zhongeric: nit: would appreciate again a note here with the ABI of the action, and especially saying that uint256.max is mapped in MarginAccount to the debt token balance of the account.

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        );
+        actionParams[4] = abi.encode(params.market.collateral, uint256(ActionConstants.OPEN_DELTA), false);
+        actionParams[5] = abi.encode(params.adapter, params.market, fullClose ? Ltv.wrap(0) : params.maxLtvAfter);
+
+        // measure the router's own collateral gain across the unlock: zero for a partial decrease (it
+        // withdraws exactly the swap cost), the realized PnL for a full close
+        uint256 balanceBefore = params.market.collateral.balanceOfSelf();
+        _setLastRepaid(0); // fail-closed: the plan's ACCOUNT_REPAY overwrites this with what it repaid
+        poolManager.unlock(abi.encode(actions, actionParams));
+        _setActiveAccount(address(0));
```
zhongeric: can be moved lower, to make it more clear that the lastRepaid calls are guarding the PM unlock, and the active Account guard must be set before calling the account.

src/MarginRouter.sol
```diff
@@ -0,0 +1,914 @@
...
+        // permissionless onBehalf repays, and route-path code can execute one): with the debt cleared
+        // mid-transaction, the repay leg no-ops, the bought debt tokens strand in the account, and the
+        // event would report a repay that never happened. Gate the partial branch on the MEASURED
+        // repay rather than the pre-read; a full close tolerates the no-op (it sweeps the over-bought
+        // debt back below) and execute plans rely on it, so only this curated branch asserts.
+        if (!fullClose && _lastRepaid() == 0) revert NoDebtToRepay();
+
+        // a full close whose route over-bought the debt (buffered for accrual) leaves the excess in the
+        // account after repay-all; return it to the caller so nothing is stranded
+        if (fullClose) {
```
zhongeric: we have multiple conditional branches checking `fullClose` here. Would be cleaner to just branch into two distinct flows for full / partial. Generally pro-linearization

PR #586, opened 2026-08-10
fix(perm-pools): run UNSUBSCRIBE in its own unlock; add burn bounds and hookData

SocksNFlops:
## Related Issue
N/A

## Description of changes
- unwindPosition performs UNSUBSCRIBE in a separate unlock, then re-applies the caller's approval immediately before BURN_POSITION, so the approval the burn depends on is set right before it runs.
- Adds amount0Min / amount1Min / hookData parameters, forwarded to BURN_POSITION (bounds → validateMinOut, hookData → the pool hook on removal). Zero bounds and empty hookData preserve current behavior.
- Adds coverage for both approval-lifecycle paths and for the new parameters.

### DIFF
--- a/src/hooks/permissionedPools/PermissionedPositionManager.sol
+++ b/src/hooks/permissionedPools/PermissionedPositionManager.sol
@@ -76,10 +76,17 @@ contract PermissionedPositionManager is PositionManager {
 
     /// @notice Force-exit the LP from a position. Burns the NFT, unwinds liquidity, and routes each currency.
     /// @dev Either PA admin may call. Per-currency fallback is derived on-chain via `_getOwner` (see
-    ///      `_unwindWithFallback`). The 6909 fallback never reverts, so the call is atomic. Emits one
-    ///      `CurrencyUnwound` event per leg.
+    ///      `_unwindWithFallback`). The 6909 fallback never reverts, so routing cannot strand a currency.
+    ///      Emits one `CurrencyUnwound` event per leg.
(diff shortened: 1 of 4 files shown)


src/hooks/permissionedPools/PermissionedPositionManager.sol
```diff
@@ -76,33 +76,48 @@ contract PermissionedPositionManager is PositionManager {
...
+        // Approve so UNSUBSCRIBE below passes its own _isApprovedOrOwner check.
+        getApproved[tokenId] = msg.sender;
+
+        // Unsubscribe in a separate unlock. A hostile subscriber can overwrite getApproved during
+        // notifyUnsubscribe, but that write is discarded by the re-approval below, once this unlock
+        // has closed and no further subscriber callback can run.
+        bytes memory actions = abi.encodePacked(uint8(Actions.UNSUBSCRIBE));
+        bytes[] memory params = new bytes[](1);
+        params[0] = abi.encode(tokenId);
+        poolManager.unlock(abi.encode(actions, params));
```
ericneil-sanc: Just realized this now: You could skip unsubscribe when the position has no subscriber

ericneil-sanc (dismissed):

src/hooks/permissionedPools/PermissionedPositionManager.sol
SocksNFlops: Fixed

ericneil-sanc (approved):

PR #591, opened 2026-08-25
feat: SwapAndAdd — (add / rebalance / increase / compound)

mgretzke:
## Summary

`SwapAndAdd` lets a caller supply the two pool tokens in any ratio — including only one, or an unrelated token routed through the Universal Router — and end up with a standard POSM ERC-721 position in a single transaction. Four operations share one core:

- **`add`** — fresh capital → new position
- **`rebalance`** — burn an existing position → new position in a new range, with signed per-token add/cash-out deltas
- **`increase`** — fresh capital + the position's accrued fees → grow an existing tokenId in place
- **`compound`** — accrued fees only (an increase with a zero pulled budget)

## Design

**Route first, then size from reality.** The naive "compute the exact swap for the perfect ratio" is a circular equation (the swap moves the price that defines the ratio). Instead: run the caller's verbatim UR route first (a black box — it may even trade the target pool), size the position from the actual post-route balances at the live price (fee-aware: the side the same-pool reconcile will sell is discounted by that direction's total swap fee), mint optimistically by flash-taking the shortfall from the PoolManager, then one same-pool swap plus a bounded trim make the books balance. **`minLiquidity` on the final, post-trim position is the single slippage knob** — no swap-rate inputs, no per-amount minimums.

**One main unlock per operation** (`_run` → `_unlockCallback` → `_swapAndAdd`); rebalance first burns the old position through POSM's own vanilla burn path so every op enters the funnel with resolved budgets. The unlock payload is the `CoreParams` struct itself.

**No funds at rest.** Every operation pulls, deploys/settles in full, and sweeps the remainder (pool-token dust — possibly in both tokens, never re-denominated — and unconsumed route funding) within the same operation. Standing max Permit2 allowances to POSM and the UR are safe under this invariant: those spenders only ever pull from their direct caller, i.e. only mid-operation, under the reentrancy lock, with `minLiquidity` bounding the outcome.

**Multicall** (POSM-style) enables `[permitBatch, op]` — approve by signature and operate in one transaction — and keeper batches over several positions. `msg.value` keeps exactly one meaning per batch: ops assert exact equality plus a balance guard, and native spending is balance-funded, so the classic double-spend is unrepresentable (test-pinned).

**Auth model.** Rebalance/increase/compound require owner or ERC-721-approved operator; an operator's entire output (NFT, cash-out, dust, unconsumed funding) is forced to the owner, so a standing approval can never redirect position value. Op-level events (`Added`/`Rebalanced`/`Increased`/`Compounded`) index recipient and tokenIds — `Rebalanced` carries the old→new lineage nothing onchain otherwise links.

## Known limits (documented in `ISwapAndAdd`)

- Hooks with `*_RETURNS_DELTA` permissions are unsupported (they break the reconcile's conservation identity); failure is an atomic revert, funds safe. Dynamic LP-fee hooks are supported and tested.
- Wei-scale budgets can't fully settle the pool's mint/burn rounding toll; a non-zero floor surfaces this as `InsufficientLiquidity`.
- Fee-on-transfer and rebasing tokens are unsupported (as pool currencies — v4 itself does not support them — and as route funding); failures are atomic reverts or a larger trim.
- POSM subscribers are notified from inside the operation; a reverting or pool-trading subscriber affects only its own position's operations, gated by the floor (test-pinned).

## Testing

815 tests green under `--isolate` with gas-snapshot check:

- unit suites per surface (core ops, trim, route funding, same-pool routes, out-of-range increases, subscribers, constructor, events)
- pool-config property fuzz (price ±300k ticks, fee, spacing, depth, ranges, budgets): success + no-funds-at-rest or a whitelisted clean revert
- math fuzzes held at 5000 runs: trim-inverse never under-frees across the full sqrt-price domain; fee-aware sizing accuracy via a dust-tightness bound proven to fail if the discount breaks; extreme-tick full-deployment in both numeraires
- adversarial multicall: batch-vs-sequential observational equivalence (fuzzed), atomic rollback, cross-owner permit mixing, operator batches, `msg.value` discipline
- a far-edge probe pinning that the reconcile sell cannot reach the range's far side on supported pools (grid + fuzz)
- fork suites (Sepolia live-bytecode + mainnet Trading-API fixture) live in `test-integration/` under the `integration` profile

## Deployment note

The routed path requires the Universal Router feature that runs `V4_SWAP` inside an existing PoolManager unlock (submodule pinned at `cf27fb66e5`; upstream candidate for the next UR release). Until that ships in the canonical router, a deployment pairs the zap with its own UR build — the deploy script supports reusing one via `ROUTER=`. Contract compiles under the POSM optimizer profile (via_ir, 500 runs) at 18,233 bytes runtime.


script/fetch-tapi-route.sh
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env bash
```
zhongeric, line 1: should this be commited?

src/interfaces/ISwapAndAdd.sol
```diff
@@ -0,0 +1,389 @@
...
+///         same-pool reconcile will swap by that direction's total swap fee: lp fee compounded with the
+///         directional protocol fee) — and mint it to THIS contract (POSM),
+///      3. one same-pool reconcile swap funds whichever side the mint is short of (either direction: top up if
+///         the route under-converted, sell back if it over-converted), then a DECREASE ("trim") lands the
+///         position exactly on what the holdings support,
+///      4. enforce `minLiquidity`, sweep the small remainder (dust, possibly in both pool tokens) to
+///         `recipient`, and transfer the
+///         NFT to `recipient` after the unlock closes.
+///
+///      DESIGN NOTE — route first, then size:
```
zhongeric: rm all of this content

src/interfaces/ISwapAndAdd.sol
```diff
@@ -0,0 +1,389 @@
...
+    ///                      from arbitrary tokens: e.g. supply USDC, route it into the pool's two tokens). Each
+    ///                      entry is Permit2-pulled up front; whatever of it the route leaves unconsumed is
+    ///                      swept to `recipient` after the operation. Entries must not be pool currencies
+    ///                      (those are `amount0In`/`amount1In`) and require a non-empty `route`. Native ETH may
+    ///                      be an entry (token = address(0)) only when currency0 is not native, and msg.value
+    ///                      must equal that entry's amount — msg.value always has exactly one meaning. A
+    ///                      zero-amount entry pulls nothing but is still swept: it claims any donated balance
+    ///                      of that token (consistent with the donation doctrine in the contract INVARIANT).
+    ///                      Fee-on-transfer tokens are unsupported (the route's declared input would exceed the
+    ///                      delivered balance), matching the pool-token policy.
```
zhongeric: too verbose

src/interfaces/ISwapAndAdd.sol
```diff
@@ -0,0 +1,389 @@
...
+        int24 tickLower;
+        int24 tickUpper;
+        uint256 amount0In;
+        uint256 amount1In;
+        bytes route;
+        TokenAmount[] routeFunding;
+        uint256 minLiquidity;
+        address recipient;
+        bytes hookData;
+        uint256 deadline;
```
zhongeric, line 138: Rearrange so statically sized preceed dynamic

zhongeric: Also there are a lot of params here that seem to be related to two diff entities (position and route)

src/interfaces/ISwapAndAdd.sol
```diff
@@ -0,0 +1,389 @@
...
+        uint256 tokenId;
+        uint256 amount0In;
+        uint256 amount1In;
+        bytes route;
+        TokenAmount[] routeFunding;
+        uint256 minLiquidityAdded;
+        address recipient;
+        bytes hookData;
+        uint256 deadline;
+    }
```
zhongeric, line 175: same here with the prev struct reordering

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+        // pool (a supported routeFunding flow that pushes zero value), or a pre-existing donation — UR
+        // balances are permissionlessly sweepable, so not a wei may be left there. The SWEEP takes the
+        // router's WHOLE balance; on a native pool it joins the caller's budget, otherwise it rests here
+        // until claimed (donation doctrine / a zero-amount address(0) routeFunding entry).
+        if (address(universalRouter).balance > 0) {
+            bytes[] memory sweepInputs = new bytes[](1);
+            // token ETH (address(0)), recipient MSG_SENDER (UR maps it back to this contract), no minimum.
+            sweepInputs[0] = abi.encode(address(0), ActionConstants.MSG_SENDER, 0);
+            universalRouter.execute(abi.encodePacked(UR_SWEEP_COMMAND), sweepInputs);
+        }
```
zhongeric, line 476: Could we just check that the last input / action is a sweep? Why make two calls?

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+
+    /// @dev Run the caller's verbatim Universal Router route. The route encodes its own input token and (fixed)
+    ///      input amount, so BOTH sides are funded up front — the standing Permit2 allowances cover the ERC-20
+    ///      sides and the native balance is forwarded as value — and the route pulls whichever side it declares.
+    ///      (Funding one side chosen by an on-chain surplus classification could flip vs the off-chain route
+    ///      direction on price drift and leave the route's pull unfunded.) UR consumes exactly what the route
+    ///      commands specify; any native value the route did not consume is reclaimed via a UR SWEEP right
+    ///      after — UR's balance is permissionlessly sweepable, so not a wei may be left there. Everything
+    ///      unconsumed thus stays in this contract for the same-pool reconcile.
+    function _runRoute(CoreParams memory cp) internal {
```
zhongeric: nit: don't like the naming as its not descriptive

src/interfaces/ISwapAndAdd.sol
mgretzke: yeah, docs and natspec are all claude rn, will trim them down heavily

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+        // dust can sit in BOTH tokens (the trim frees both sides); it is not re-denominated.
+        (amount0, amount1) = _positionAmounts(cp, liquidity);
+        _sweep(cp.key.currency0, cp.recipient);
+        _sweep(cp.key.currency1, cp.recipient);
+    }
+
+    /// @dev Size the position from the current holdings at the live price. Sizing is fee-aware: the side that
+    ///      will be swapped same-pool in `_reconcile` is discounted by that direction's total swap fee, so the
+    ///      optimistic deploy isn't over-sized by a fee the trim would otherwise claw back.
+    function _planLiquidity(CoreParams memory cp)
```
zhongeric: You only use a subset of the params in CoreParams here. Much cleaner to only take those, and consider writing a library for this logic to externalize it.

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+        internal
+        view
+        returns (uint128 liquidity, uint256 amount0, uint256 amount1)
+    {
+        (uint160 sqrtPriceX96,, uint24 protocolFee, uint24 lpFee) = poolManager.getSlot0(cp.key.toId());
+        // the sizing math divides by the price; only initialized pools guarantee sp >= MIN_SQRT_PRICE.
+        if (sqrtPriceX96 == 0) revert IPoolManager.PoolNotInitialized();
```
zhongeric: Feel like this check should be moved outside of this helper func

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+    /// @dev Flash-`take` from the pool whatever the optimistic deploy needs beyond the held budget, on each side,
+    ///      so the subsequent mint/increase is fully funded; `_reconcile` later settles what the take actually owes.
+    ///      EXACT to the wei — no rounding buffer: `_getAmountsForLiquidity` rounds up with the pool's own math
+    ///      on the pool's own inputs, so the deploy can never pull more than `amount0/1` (the tick/price boundary
+    ///      states, the one place the formulas could diverge, are pinned by tests). Requires the PoolManager to
+    ///      hold the taken amount across ALL pools; if it is globally drained of that token the take reverts
+    ///      inside the token transfer — the same state leaves the reconcile swap nothing to source the deficit
+    ///      from, so the operation is unviable regardless and no pre-check is spent on it (the deficit token is
+    ///      usually a cold SLOAD, i.e. real gas on every call to buy a prettier error in an already-doomed state).
+    function _flashTakeDeficit(CoreParams memory cp, uint256 amount0, uint256 amount1) internal {
```
zhongeric, line 349: no need for an internal func here IMO

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+    ///      Each currency is settled with CLOSE_CURRENCY rather than one SETTLE_PAIR: POSM applies
+    ///      `liquidityDelta + feesAccrued` to its own delta, so on a grow op a fee credit larger than that
+    ///      side's principal makes the delta POSITIVE, and SETTLE_PAIR (which asserts a debt) would revert
+    ///      DeltaNotNegative. The callback collects the position's fees before the route, so the only source
+    ///      left is fees accrued DURING the operation — a route leg trading this pool, or a hook doing the
+    ///      same — but tolerating either sign removes the whole failure class instead of proving each source
+    ///      is handled. A taken credit lands in this contract's balance, where the reconcile spends it on the
+    ///      flash debt (a smaller trim, i.e. more retained liquidity) and the sweep returns any remainder.
+    ///      On a MINT the delta is always a pure debt, so CLOSE_CURRENCY is exactly SETTLE_PAIR there.
+    function _deploy(CoreParams memory cp, uint128 liquidity, uint256 amount0) internal returns (uint256 tokenId) {
```
zhongeric: nit: rename to something more descriptive

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+        if (refValue == 0) return 0;
+        return FullMath.mulDiv(REFERENCE_LIQUIDITY, budgetValue, refValue).toUint128();
+    }
+
+    /// @dev token amounts required for `liquidity` at the given price/range, via SqrtPriceMath. Rounded UP to
+    ///      mirror POSM's MINT_POSITION (which rounds the pull up, in the pool's favor): so the flash-take
+    ///      decision in `_flashTakeDeficit` sees the true required amount and the contract is never a wei short
+    ///      of what POSM pulls. For sizing, rounding the reference up only shrinks L — the safe direction. The
+    ///      <=1-wei over-estimate is swept as dust.
+    function _getAmountsForLiquidity(uint160 sqrtPriceX96, uint160 sqrtA, uint160 sqrtB, uint128 liquidity)
```
zhongeric: I'm pretty sure i wrote something like this for LiquidityLauncher - maybe take a look?

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+    /// @dev Size L by value conservation: the budget's value must equal the value of a position of size L at
+    ///      the current price, so L = REFERENCE_LIQUIDITY * value(budget) / value(reference position). Values
+    ///      are expressed in the CHEAPER token (token1 when price >= 1, token0 otherwise) so the conversion
+    ///      rate is always >= Q96: the expensive side is multiplied up, never truncated toward zero, keeping
+    ///      the sizing exact (<= 2^-95 rate error) across the full tick domain. `pips0`/`pips1` weight each
+    ///      side's value (1e6 = 100%); the fee-aware caller discounts exactly one — the side the reconcile
+    ///      will sell.
+    ///      Precision: the final division truncates at 1 part in value(reference) and only ever under-sizes —
+    ///      the safe direction (the leftover is swept back as dust).
+    function _sizeLiquidityWeighted(
```
zhongeric: nit: _getLiquidityForAmountsWeighted

src/interfaces/ISwapAndAdd.sol
```diff
@@ -0,0 +1,389 @@
...
+        int128 additional1;
+        int24 newTickLower;
+        int24 newTickUpper;
+        bytes route;
+        TokenAmount[] routeFunding;
+        uint256 minLiquidity;
+        address recipient;
+        bytes hookData;
+        uint256 deadline;
+    }
```
zhongeric, line 214: I'd like to see more composition between the struct types here. I feel like we have Position entities duplicated / flattened / renamed across the diff structs.

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+        if (address(universalRouter).balance > 0) {
+            bytes[] memory sweepInputs = new bytes[](1);
+            // token ETH (address(0)), recipient MSG_SENDER (UR maps it back to this contract), no minimum.
+            sweepInputs[0] = abi.encode(address(0), ActionConstants.MSG_SENDER, 0);
+            universalRouter.execute(abi.encodePacked(UR_SWEEP_COMMAND), sweepInputs);
+        }
+    }
+
+    /// @dev Burn the whole position and TAKE both tokens (+ fees) to this contract.
+    function _withdraw(PoolKey memory key, uint256 tokenId, bytes calldata hookData) internal {
```
zhongeric: rename to include burn in the func name

src/SwapAndAdd.sol
zhongeric: like mintAndAddLiquidity

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+        } else if (sqrtPriceX96 < sqrtB) {
+            amount0 = SqrtPriceMath.getAmount0Delta(sqrtPriceX96, sqrtB, liquidity, true);
+            amount1 = SqrtPriceMath.getAmount1Delta(sqrtA, sqrtPriceX96, liquidity, true);
+        } else {
+            amount1 = SqrtPriceMath.getAmount1Delta(sqrtA, sqrtB, liquidity, true);
+        }
+    }
+
+    /// @dev The final position composition at the live (post-reconcile) price, for the return values.
+    function _positionAmounts(CoreParams memory cp, uint128 liquidity)
```
zhongeric: No need for an internal func here

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+    ///      holds the owner's pre-existing principal, which the trim must never consume.
+    ///      The token->liquidity inverse ROUNDS UP on every division, over `amountOut + 1`, so the DECREASE's
+    ///      rounded-DOWN return always covers `amountOut`: the token0 inverse floors an intermediate value
+    ///      whose truncation is scaled by the amount, so its error in liquidity units is unbounded — a
+    ///      "+1 liquidity unit" style compensation cannot bound a shortfall measured in token wei.
+    ///      The cap is reachable only when the deficit exceeds what unwinding ALL of `lopt` returns, i.e. the
+    ///      budget is within the pool's ~1-wei mint/burn rounding toll: the trim then removes all of `lopt`,
+    ///      the added liquidity is 0 and any non-zero `minLiquidity` floor reverts InsufficientLiquidity (a zero
+    ///      floor lets the unsettled toll surface as v4's CurrencyNotSettled at the end of the unlock).
+    function _trim(CoreParams memory cp, uint256 tokenId, uint128 lopt, bool deficitIs1, uint256 amountOut)
```
zhongeric: nit: `deficitIsCurrency0` to align with existing nomenclature

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+        params[1] = abi.encode(key.currency0, key.currency1, ActionConstants.MSG_SENDER);
+        // POSM's SELF-unlocking surface: the burn is a vanilla POSM operation in its own unlock, run before
+        // the main unlock opens. Its mandatory deadline is passed as max: staleness is checked exactly once
+        // per operation, by the entrypoint's checkDeadline (same transaction, same timestamp — a real value
+        // here could never fire differently).
+        positionManager.modifyLiquidities(abi.encode(actions, params), type(uint256).max);
+    }
+
+    /// @dev DECREASE `dl` liquidity (0 collects just the accrued fees) and TAKE both tokens to this contract.
+    function _decrease(PoolKey memory key, uint256 tokenId, uint128 dl, bytes memory hookData) internal {
```
ccashwell, line 490: calling this with `dl = 0` on a position that has 0 liquidity units will revert with `CannotUpdateEmptyPosition()`, suggest adding a simple guard:

```solidity
if (positionManager.getPositionLiquidity(cp.deployTokenId) == 0) return;
```

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+    ///      commands specify; any native value the route did not consume is reclaimed via a UR SWEEP right
+    ///      after — UR's balance is permissionlessly sweepable, so not a wei may be left there. Everything
+    ///      unconsumed thus stays in this contract for the same-pool reconcile.
+    function _runRoute(CoreParams memory cp) internal {
+        (bytes memory commands, bytes[] memory inputs) = abi.decode(cp.route, (bytes, bytes[]));
+        // forward the WHOLE native balance: under no-funds-at-rest it is exactly the operation's own native
+        // budget (native pool) or native route funding (non-native pool) — zero when neither applies.
+        uint256 value = address(this).balance;
+
+        universalRouter.execute{value: value}(commands, inputs);
```
ccashwell, line 469: because you're forwarding the entire native balance into the caller-supplied route here, any balance-relative UR command can overspend. for example, `PAY_PORTION` or `SWEEP` with a bps argument will size against the whole route's budget rather than just the swap output. suggest broadening the documented route construction constraint to note that all balance-relative commands (not just `WRAP_ETH`) are potential footguns.

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+            uint160 hi = sqrtPriceX96 < sqrtUpper ? sqrtPriceX96 : sqrtUpper;
+            dlUp = FullMath.mulDivRoundingUp(amountOut + 1, FixedPoint96.Q96, hi - sqrtLower);
+        } else {
+            // token0 occupies [max(price, sqrtLower), sqrtUpper]: amount0 = L * Q96 * (hi - lo) / (hi * lo)
+            uint160 lo = sqrtPriceX96 > sqrtLower ? sqrtPriceX96 : sqrtLower;
+            uint256 intermediate = FullMath.mulDivRoundingUp(lo, sqrtUpper, FixedPoint96.Q96);
+            // KNOWN (informational): at extreme prices, with the post-swap price parked within sqrt-units of
+            // `sqrtUpper` while a large deficit remains, this quotient can exceed uint256 and revert (blank
+            // EvmError) before the `lopt` cap below would have answered — self-inflicted, atomic, transient.
+            dlUp = FullMath.mulDivRoundingUp(amountOut + 1, intermediate, sqrtUpper - lo);
```
ccashwell: you can fix this overflow edge case by comparing `amountOut` against `getAmount0Delta(lo, sqrtUpper, lopt, false)` first

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.24;
```
ccashwell: use pinned pragma on deployable contracts

src/interfaces/ISwapAndAdd.sol
```diff
@@ -0,0 +1,389 @@
...
+///        - `AFTER_REMOVE_LIQUIDITY_RETURNS_DELTA` — skims what the trim, the fee collect or a rebalance burn
+///          frees, so the trim frees less than computed (or the delta turns negative). UNSUPPORTED.
+///        - `AFTER_ADD_LIQUIDITY_RETURNS_DELTA` — charges the deploy instead, which this contract settles from
+///          the operation's budget: it behaves as an extra cost that leaves less for the position, gated by
+///          `minLiquidity` rather than breaking settlement. Expected to degrade gracefully; NOT verified by
+///          tests, so treat as unsupported until it is.
+///      On an unsupported pool the call reverts atomically inside the unlock (CurrencyNotSettled,
+///      DeltaNotPositive from a removal whose delta went negative, or a low-level arithmetic revert from the
+///      trim math when a skimmed swap leaves the price past the range's far edge) — funds are always safe,
+///      but the operation can simply be unusable there.
```
ccashwell: this policy is enforced only by documentation, and the "reverts atomically" claim is imprecise. the policy could be enforced once, cheaply, via `Hooks.hasPermission` (which is a pure bitmask operation, no SLOAD overhead) with a typed error, instead of leaving every integrator to re-implement the screen and users to hit opaque deep-in-flow reverts. 

also the "reverts atomically" claim is not necessarily universally true. for example, a mildly-skimming `AFTER_SWAP_RETURNS_DELTA` hook is absorbed by a larger trim and the op succeeds with a worse outcome (still bound by `minLiquidity`). and the `AFTER_ADD_LIQUIDITY_RETURNS_DELTA` "degrades gracefully" claim is not covered by any tests as far as i could tell. 

suggest `revert UnsupportedHookPermissions(key.hooks)` in `_swapAndAdd` when any of the three flags is set, and add a test that proves the graceful degradation claim.

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+    ///      budget is within the pool's ~1-wei mint/burn rounding toll: the trim then removes all of `lopt`,
+    ///      the added liquidity is 0 and any non-zero `minLiquidity` floor reverts InsufficientLiquidity (a zero
+    ///      floor lets the unsettled toll surface as v4's CurrencyNotSettled at the end of the unlock).
+    function _trim(CoreParams memory cp, uint256 tokenId, uint128 lopt, bool deficitIs1, uint256 amountOut)
+        internal
+        returns (uint128 dl)
+    {
+        (uint160 sqrtPriceX96,,,) = poolManager.getSlot0(cp.key.toId());
+        uint160 sqrtLower = TickMath.getSqrtPriceAtTick(cp.tickLower);
+        uint160 sqrtUpper = TickMath.getSqrtPriceAtTick(cp.tickUpper);
```
ccashwell: these three are recomputed in `_planLiquidity`, `_trim` and `_positionAmounts`, adding a few thousand extra gas per operation. consider threading them (or using transient if you don't want to thread).

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,897 @@
...
+        // funding token from the collected fees, so funding leftovers are output like any other).
+        _sweepFunding(params.routeFunding, recipient);
+    }
+
+    /// @inheritdoc ISwapAndAdd
+    function compound(CompoundParams calldata params)
+        external
+        isNotLocked
+        checkDeadline(params.deadline)
+        returns (uint128 liquidityAdded, uint256 amount0, uint256 amount1)
```
ccashwell, line 206: not declared as payable but can consume a batch's msg.value under a multicall scenario. not exploitable afaict but confusing.

ccashwell (commented):
few small comments

src/SwapAndAdd.sol
mgretzke: `sqrtPriceX96` needs to be recalculated at each step (because the reconcile swap, or a hook could affect it), but `sqrtLower` and `sqrtUpper` can be reused!
Resolved [here](https://github.com/Uniswap/v4-periphery/pull/591/commits/cb2b69e0abc65f4d423a2e6d3f107838bab19272)

src/SwapAndAdd.sol
mgretzke: https://github.com/Uniswap/v4-periphery/pull/591/commits/5496f9262d9d23ebb1c063d6e79e850c26180bdd

src/interfaces/ISwapAndAdd.sol
mgretzke: fully reworked natspec [here](https://github.com/Uniswap/v4-periphery/pull/591/commits/7752d9a21f4828b12bf990710a1e8bb59fc179c5)

src/interfaces/ISwapAndAdd.sol
mgretzke: Removed all of it [here](https://github.com/Uniswap/v4-periphery/pull/591/commits/7752d9a21f4828b12bf990710a1e8bb59fc179c5)

Replaced it with integration notes

src/SwapAndAdd.sol
mgretzke: yeah, understand your issue with it... Do you see a simple solution though? Making compound payable just because of the multicall opportunity seems to hurt the interface more then help it

src/SwapAndAdd.sol
mgretzke: It only runs if there is still balance inside the universal router, so its already limited to a route design that does not sweep as a last action.

src/SwapAndAdd.sol
mgretzke: https://github.com/Uniswap/v4-periphery/pull/591/commits/0da496e45489dcfc088eeceec0b92e0712fdc50b

src/SwapAndAdd.sol
mgretzke: https://github.com/Uniswap/v4-periphery/pull/591/commits/0da496e45489dcfc088eeceec0b92e0712fdc50b

mgretzke: mintAndAdd is not fully correct, since it either mints or increases

src/SwapAndAdd.sol
mgretzke: https://github.com/Uniswap/v4-periphery/pull/591/commits/0da496e45489dcfc088eeceec0b92e0712fdc50b

src/SwapAndAdd.sol
mgretzke: generally agree, but even after reducing the natspec, this function still does relevant documentation work. Purely because of that aspect, I lean towards leaving as is

src/SwapAndAdd.sol
mgretzke: Essentially, yes:
https://github.com/Uniswap/liquidity-launcher/blob/799e769ef1d826d8a192cf3082d7aae32b6418b3/src/libraries/PositionPlanner.sol#L383-L401

its a private function though and overkill to import all of LL just for that. It is basically the same though 👍

script/fetch-tapi-route.sh
mgretzke: Should be fine. Its to generate real TAPI data for the fork e2e tests. The script is not used at runtime of the fork test, but allows to generate these hardcoded routes.

src/libraries/SwapAndAddMath.sol
```diff
@@ -0,0 +1,172 @@
...
+import {ProtocolFeeLibrary} from "@uniswap/v4-core/src/libraries/ProtocolFeeLibrary.sol";
+
+/// @title SwapAndAddMath
+/// @notice Pure sizing and trim math for SwapAndAdd. All functions are stateless: callers supply the live
+///         sqrt price, range bounds, and fee configuration.
+library SwapAndAddMath {
+    using SafeCast for uint256;
+
+    uint256 internal constant PIPS_DENOMINATOR = 1e6;
+    /// @dev Sizing reference: maximal so the reference-scaling quantization is negligible.
```
zhongeric: nit: rewrite this in more understandable terms

src/interfaces/ISwapAndAdd.sol
mgretzke: addressed in [here](https://github.com/Uniswap/v4-periphery/pull/591/commits/5abe322570eb003016a2d2cf7709cb7f30d44669)

src/interfaces/ISwapAndAdd.sol
mgretzke: addressed [here](https://github.com/Uniswap/v4-periphery/pull/591/commits/5abe322570eb003016a2d2cf7709cb7f30d44669)

src/interfaces/ISwapAndAdd.sol
mgretzke: added a hook permission gate here:
https://github.com/Uniswap/v4-periphery/pull/591/commits/5abe322570eb003016a2d2cf7709cb7f30d44669

src/SwapAndAdd.sol
mgretzke: fixed here:
https://github.com/Uniswap/v4-periphery/pull/591/commits/5abe322570eb003016a2d2cf7709cb7f30d44669

src/SwapAndAdd.sol
mgretzke: I moved everything that was actually pure math into its own library. It feels a bit strange to add functions with external calls to the library. If you feel strongly about it, let me know.

src/libraries/SwapAndAddMath.sol
```diff
@@ -0,0 +1,172 @@
...
+        uint160 sqrtPriceLowerX96,
+        uint160 sqrtPriceUpperX96,
+        uint256 amount0,
+        uint256 amount1,
+        uint256 pipsWeight0,
+        uint256 pipsWeight1
+    ) internal pure returns (uint128) {
+        // Price >= 1 values both sides in token1 terms via rateX96 = sqrtPrice^2 / Q96 (token1 per token0);
+        // price < 1 values in token0 terms via rateX96 = Q96^3 / sqrtPrice^2 (split to prevent intermediate
+        // overflow). Valuing in the cheaper token keeps the rate multiplier >= Q96 across extreme ticks.
```
zhongeric: Basically what you're doing here is inversing the price ratio depending on which token is cheaper right? Hard to follow these comments

src/libraries/SwapAndAddMath.sol
```diff
@@ -0,0 +1,172 @@
...
+        bool isToken1Cheaper = sqrtPriceX96 >= FixedPoint96.Q96;
```
zhongeric, line 84: nit: cheaper or equal?

src/libraries/SwapAndAddMath.sol
```diff
@@ -0,0 +1,172 @@
...
+        uint256 rateX96 = isToken1Cheaper
+            ? FullMath.mulDiv(sqrtPriceX96, sqrtPriceX96, FixedPoint96.Q96)
+            : FullMath.mulDiv(
+                FullMath.mulDiv(FixedPoint96.Q96, FixedPoint96.Q96, sqrtPriceX96), FixedPoint96.Q96, sqrtPriceX96
+            );
+
+        uint256 refValue;
```
zhongeric, line 91: nit: weightedRefLiquidity

src/libraries/SwapAndAddMath.sol
```diff
@@ -0,0 +1,172 @@
...
+            uint256 intermediate = FullMath.mulDivRoundingUp(clampedLower, sqrtPriceUpperX96, FixedPoint96.Q96);
+            // Informational: At extreme prices where post-swap price is within sqrt-units of sqrtUpper with large
+            // deficit remaining, this intermediate quotient can overflow uint256 and revert (self-inflicted, safe).
+            liquidityToFree =
+                FullMath.mulDivRoundingUp(amountToCover + 1, intermediate, sqrtPriceUpperX96 - clampedLower);
+        }
+    }
+
+    /// @dev Values a token pair in the cheaper-token numeraire, weighting each side by its pips factor.
+    function _weightedValue(
```
zhongeric: So this actually does more than just apply a weight, which is unclear from the function name. Would be nice to break this up and/or have the calling function do the conditional logic

src/libraries/SwapAndAddMath.sol
```diff
@@ -0,0 +1,172 @@
...
+            );
+
+        uint256 refValue;
+        {
+            (uint256 ref0, uint256 ref1) =
+                getAmountsForLiquidity(sqrtPriceX96, sqrtPriceLowerX96, sqrtPriceUpperX96, REFERENCE_LIQUIDITY);
+            refValue = _weightedValue(ref0, ref1, rateX96, isToken1Cheaper, pipsWeight0, pipsWeight1);
+        }
+        if (refValue == 0) return 0;
+        // Scale reference liquidity: L = REFERENCE_LIQUIDITY * budgetValue / refValue.
```
zhongeric: nit: this is more like `budget * REF / weightedLiquidityValue` right? easier to read that we're applying a multiplicative factor like this

src/libraries/SwapAndAddMath.sol
```diff
@@ -0,0 +1,172 @@
...
+        // Truncating towards zero here is the safe direction (any leftover budget is swept as dust).
+        uint256 budgetValue = _weightedValue(amount0, amount1, rateX96, isToken1Cheaper, pipsWeight0, pipsWeight1);
+        return FullMath.mulDiv(REFERENCE_LIQUIDITY, budgetValue, refValue).toUint128();
+    }
+
+    /// @notice Calculates token amounts required for a given liquidity amount at current price/range.
+    /// @dev Rounds UP to mirror POSM's MINT_POSITION — callers funding a mint from these amounts are never
+    ///      a wei short of POSM's pull. NOT interchangeable with LiquidityAmounts (which rounds down).
+    ///      The A/B bounds follow LiquidityAmounts' convention: they may be passed in either order.
+    function getAmountsForLiquidity(
```
zhongeric: nit: I know I have this named `getAmountsForLiquidity` in LiquidityLauncher - but maybe worth calling this `getAmountsForLiquidityRoundingUp`

src/libraries/SwapAndAddMath.sol
```diff
@@ -0,0 +1,172 @@
...
+        }
+    }
+
+    /// @notice Computes the liquidity to burn so that v4's rounded-DOWN return covers `amountToCover` of the
+    ///         deficit token.
+    /// @dev Exact ceil inverse over `amountToCover + 1`: the nested-floor bound guarantees freed >= amountToCover.
+    ///      Assumes the price is not past the range's far side for the deficit token (SwapAndAdd's reconcile
+    ///      flow guarantees this); a price outside the near side clamps to the boundary.
+    /// @return liquidityToFree The round-up liquidity to burn, uncapped — callers cap against the liquidity they added.
+    function getLiquidityToFree(
```
zhongeric: this name doesn't make sense. additionally, this is kind of one subset of use case of `getAmountsForLiquidity`

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+        }
+
+        // Validate msg.value: exactly one native contribution (pool budget OR route funding) is allowed.
+        if (msg.value != expectedValue) revert InvalidEthValue();
+        // In multicall batches, verify balance was not already spent by an earlier subcall.
+        if (address(this).balance < expectedValue) revert InvalidEthValue();
+    }
+
+    /// @dev Sweeps unconsumed route funding tokens to the resolved recipient.
+    function _sweepFunding(TokenAmount[] calldata funding, address to) internal {
```
zhongeric, line 573: nit: no need to bind this to "funding", or the TokenAmount struct.

src/interfaces/ISwapAndAdd.sol
```diff
@@ -0,0 +1,264 @@
...
+    /// @param minLiquidity Minimum liquidity required for the newly minted position (slippage floor).
+    /// @param recipient Destination for the new position NFT, returned cash-out tokens, and dust. Forced to owner if caller is an operator.
+    /// @param deadline Timestamp after which the transaction will revert.
+    /// @param route Encoded Universal Router commands and inputs (can be empty).
+    /// @param routeFunding Optional non-pool tokens pulled to fund the route. Unused amounts are swept to `recipient`.
+    /// @param hookData Arbitrary data passed to pool hooks.
+    struct RebalanceParams {
+        uint256 tokenId;
+        int128 additional0;
+        int128 additional1;
```
zhongeric, line 205: would reconsider naming here if these can be negative / represent cashout

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+        if (d > 0) _take(currency, address(this), uint256(d));
+    }
+
+    /// @dev Pulls caller budgets and route funding via Permit2/msg.value, validating single native ETH accounting.
+    function _pull(
+        PoolKey memory key,
+        uint256 amount0In,
+        uint256 amount1In,
+        TokenAmount[] calldata funding,
+        bytes calldata route
```
zhongeric, line 534: why do you have to pass in route here? not used except for this funding carve out below but it all seems bit clunky

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+    ) internal {
+        uint256 expectedValue;
+
+        // Pull route funding tokens (non-pool tokens dedicated to the off-chain route).
+        if (funding.length != 0) {
+            if (route.length == 0) revert RouteFundingRequiresRoute();
+            for (uint256 i = 0; i < funding.length; i++) {
+                Currency token = funding[i].token;
+                if (token == key.currency0 || token == key.currency1) revert InvalidFundingToken(token);
```
zhongeric, line 542: Maybe I'm misremembering but I thought we discussed being allowed to pull tokens related to the Pool key?

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+
+        // Pull pool-token budget: currency0 can be native or ERC-20; currency1 is always ERC-20.
+        Currency c0 = key.currency0;
+        if (c0.isAddressZero()) {
+            expectedValue += amount0In;
+        } else if (amount0In > 0) {
+            permit2.transferFrom(msg.sender, address(this), amount0In.toUint160(), Currency.unwrap(c0));
+        }
+        if (amount1In > 0) {
+            permit2.transferFrom(msg.sender, address(this), amount1In.toUint160(), Currency.unwrap(key.currency1));
```
zhongeric, line 564: probably easier to use the V4 currency library here

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+                sqrtPriceLimitX96: zeroForOne ? TickMath.MIN_SQRT_PRICE + 1 : TickMath.MAX_SQRT_PRICE - 1
+            }),
+            hookData
+        );
+    }
+
+    // ───────────────────────────────────────────── delta / token helpers ─────────────────────────────────────────────
+
+    /// @dev Settles outstanding debt for a currency using held balance.
+    function _settleToward(Currency currency) internal {
```
zhongeric: this is just `_settle`? what does toward mean

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+    function _decrease(PoolKey memory key, uint256 tokenId, uint128 dl, bytes memory hookData) internal {
+        bytes memory actions = abi.encodePacked(uint8(Actions.DECREASE_LIQUIDITY), uint8(Actions.TAKE_PAIR));
+        bytes[] memory params = new bytes[](2);
+        params[0] = abi.encode(tokenId, uint256(dl), uint128(0), uint128(0), hookData);
+        params[1] = abi.encode(key.currency0, key.currency1, ActionConstants.MSG_SENDER);
+        positionManager.modifyLiquiditiesWithoutUnlock(actions, params);
+    }
+
+    /// @dev Executes a same-pool swap to convert surplus tokens to deficit without price limits.
+    ///      Unbounded because `minLiquidity` is the slippage gate and input amount is strictly bounded by held holdings.
```
zhongeric: improve this natspec: `fine to use max slippage here since we check elsewhere that minimum liquidity was added` or seomthing.

zhongeric: and/or say that the calling function MUST check min amounts before using

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+    function _swap(PoolKey memory key, bool zeroForOne, int256 amountSpecified, bytes memory hookData) internal {
```
zhongeric, line 500: Also don't really see the benefit of having htis be its own function - if anything, it's misleading because it applies defaults to the swap that affect its behavior with no indication

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+            permit2.transferFrom(msg.sender, address(this), amount0In.toUint160(), Currency.unwrap(c0));
+        }
+        if (amount1In > 0) {
+            permit2.transferFrom(msg.sender, address(this), amount1In.toUint160(), Currency.unwrap(key.currency1));
+        }
+
+        // Validate msg.value: exactly one native contribution (pool budget OR route funding) is allowed.
+        if (msg.value != expectedValue) revert InvalidEthValue();
+        // In multicall batches, verify balance was not already spent by an earlier subcall.
+        if (address(this).balance < expectedValue) revert InvalidEthValue();
```
zhongeric: combine with the prev if statement with `||`?

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+        uint256 value = address(this).balance;
+
+        universalRouter.execute{value: value}(commands, inputs);
+
+        // Reclaim any unspent native ETH left in Universal Router (UR balances are permissionlessly sweepable).
+        if (address(universalRouter).balance > 0) {
+            bytes[] memory sweepInputs = new bytes[](1);
+            sweepInputs[0] = abi.encode(address(0), ActionConstants.MSG_SENDER, 0);
+            universalRouter.execute(abi.encodePacked(UR_SWEEP_COMMAND), sweepInputs);
+        }
```
zhongeric: Hm still not convinced we need this. Why can't we just rely on the caller to do this?

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+        // Native pools carry a trailing SWEEP to return unconsumed wei of the forwarded ETH.
+        Currency c0 = cp.key.currency0;
+        bool isNative = c0.isAddressZero();
+        bytes memory actions = isNative
+            ? abi.encodePacked(
+                deployAction, uint8(Actions.CLOSE_CURRENCY), uint8(Actions.CLOSE_CURRENCY), uint8(Actions.SWEEP)
+            )
+            : abi.encodePacked(deployAction, uint8(Actions.CLOSE_CURRENCY), uint8(Actions.CLOSE_CURRENCY));
+
+        bytes[] memory params = new bytes[](isNative ? 4 : 3);
```
zhongeric, line 454: quite a bit cleaner to just build this in assembly and overwrite the length pointer imo

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+                type(uint128).max,
+                address(this),
+                cp.hookData
+            )
+            : abi.encode(tokenId, uint256(liquidity), type(uint128).max, type(uint128).max, cp.hookData);
+
+        // Native pools carry a trailing SWEEP to return unconsumed wei of the forwarded ETH.
+        Currency c0 = cp.key.currency0;
+        bool isNative = c0.isAddressZero();
+        bytes memory actions = isNative
```
zhongeric, line 448: generally want to try to reduce the number of conditional branches and linearize as much as possible

zhongeric (commented):
Can we add unit tests for the many internal functions? Namely, `deployLiquidity` and `reconcile`. These functions have lots of biz logic and conditional branching that isn't really tested.

src/SwapAndAdd.sol
```diff
@@ -0,0 +1,647 @@
...
+    }
+
+    /// @dev Burns an existing position and takes both tokens to this contract.
+    ///      POSM `modifyLiquidities` is passed `type(uint256).max` deadline because staleness is already checked by entrypoint modifier.
+    function _burnAndWithdraw(PoolKey memory key, uint256 tokenId, bytes calldata hookData) internal {
+        bytes memory actions = abi.encodePacked(uint8(Actions.BURN_POSITION), uint8(Actions.TAKE_PAIR));
+        bytes[] memory params = new bytes[](2);
+        params[0] = abi.encode(tokenId, uint128(0), uint128(0), hookData);
+        params[1] = abi.encode(key.currency0, key.currency1, ActionConstants.MSG_SENDER);
+        positionManager.modifyLiquidities(abi.encode(actions, params), type(uint256).max);
```
zhongeric, line 486: consider making the deadline block.timestamp since it doesn't matter either

src/SwapAndAdd.sol
mgretzke: pool tokens are not allowed in the funding, they get handled differently. I fully implemented a version where they were part of the funding, but it got really messy, cause pool tokens have way deeper commitments and use cases then funding tokens

src/SwapAndAdd.sol
mgretzke: oh, that makes sense. So an increase operation for a fully emptied out, but not burned position is bricked right now without this? 
Implementing the fix

mgretzke: https://github.com/Uniswap/v4-periphery/pull/591/commits/18338b83c54649fd5d352843760c9c89b00ccc06

src/SwapAndAdd.sol
mgretzke: This is for cases where you want to only swap parts of it because you use the other half for the pool right?
This lead me into looking at a broader inconsistency: 
Multicall technically allows to batch operations (even though thats not the reason its implemented) for non native funds, but it fails for native funds. because all funds are swept to the recipient, while non native funds get pulled just in time for when they are actually needed. Difficult to handle this differently though, without increasing the risk of stranding funds, and an operator can still just batch call, instead of using the multicall to batch.
So added documentation to that additionally to your suggestion, see here:
https://github.com/Uniswap/v4-periphery/pull/591/commits/8bd20a10a4c93e1408d72d5e123bb29c57cc54d2

src/SwapAndAdd.sol
mgretzke: I looked into it, the change seemed bigger then that to me? Definitely feel free to drop a PR or just add a commit if you see a way for this to be a small change with very limited area of affect

test/SwapAndAddConstructor.t.sol
```diff
@@ -0,0 +1,65 @@
...
+    function positionManager() external view returns (address);
+    function universalRouter() external view returns (address);
+}
+
+/// @notice Constructor wiring: every immutable is zero-checked (a mistyped deploy-script env var must kill the
+///         deployment with `ZeroAddress`, not produce a dead zap), and a valid deploy stores all four args.
+///         Deployed via raw `create` on the compiled artifact — `deployCode` swallows creation revert data, so
+///         the specific error could not be asserted through it (and the test profile cannot compile the
+///         contract directly: via_ir-only).
+contract SwapAndAddConstructorTest is Test {
```
zhongeric: Consider adopting the test structure we have in CCA or LL.

test/
    swapAndAdd/
        constructor.t.sol
        events.t.sol

Currently the test suite is quite hard to read

test/SwapAndAddFuzz.t.sol
```diff
@@ -0,0 +1,384 @@
...
+    ISwapAndAdd zap;
+    MockSwapRoute route;
+
+    bytes4 constant CURRENCY_NOT_SETTLED = 0x5212cba1;
+
+    // budget cap (uint88 ~ 3e26 raw) and price-domain cap (|tick| <= 300_000, price ~ 1e±13): together they
+    // keep the flash-take below the reserve pool's holdings, so the separate, documented PoolManager-drained limit (take
+    // exceeding the PoolManager's global balance) cannot masquerade as a property violation here. The trim
+    // bug's operational region (price ~1e-9, token1 from ~6.4e18) sits comfortably inside these bounds.
+    int24 constant MAX_CENTRE = 300_000;
```
zhongeric: don't know what this means

test/SwapAndAddFuzz.t.sol
```diff
@@ -0,0 +1,384 @@
...
+///         fuzz only supplied token0-side budgets — the vulnerable configuration (low price x single-sided
+///         token1) was never generated. This suite's premise: fixed pool configs are what create false
+///         confidence, so NOTHING here is fixed — initial price, fee, tick spacing, depth, range and both
+///         budget sides are all fuzzed, across all four operations.
+///
+///         THE PROPERTY: an operation either succeeds — leaving no funds at rest — or reverts with an error
+///         the design explicitly allows (the whitelist in `_allowed`). v4's CurrencyNotSettled, panics and
+///         wrapped low-level failures are always property violations: the contract's own errors and the
+///         pool's documented limits are the only acceptable ways to fail with a non-zero minLiquidity floor.
+contract SwapAndAddFuzzTest is PosmTestSetup {
```
zhongeric: clean up the AI comment here - what is the trim bug?

test/SwapAndAddFarEdge.t.sol
```diff
@@ -0,0 +1,173 @@
...
+///         traversed sqrt-prices — always worse than spot in the sell direction — while the fee-aware sizing
+///         hands the sell a surplus whose NET value equals the debt at SPOT. Worse-than-spot execution on a
+///         spot-sized budget can never afford the full-range extraction that reaching the far edge requires,
+///         so on hookless pools the post-sell price stays STRICTLY inside the far edge whenever a trim runs.
+///
+///         If that argument is wrong anywhere in the user-reachable input space, this suite goes red with a
+///         raw revert (FullMath's bare require or Panic 0x11) — which is then the regression for the fix.
+///         fee = 0 is included deliberately: no fee cushion, the surplus is exactly spot-sized, the tightest
+///         possible approach to the edge.
+contract SwapAndAddFarEdgeTest is PosmTestSetup {
```
zhongeric: Let's remove these AI regression / probe tests. If you need to test the edge case behavior add it as a focused unit test in another suite.

test/SwapAndAddIncreaseOutOfRange.t.sol
```diff
@@ -0,0 +1,228 @@
...
+import {PosmTestSetup} from "./shared/PosmTestSetup.sol";
+import {MockSwapRoute} from "./mocks/MockSwapRoute.sol";
+
+/// @notice Regression suite for `increase` on OUT-OF-RANGE positions across a sweep of realistic-to-extreme
+///         price drifts. Before the collect-fees-first fix, POSM's accrued-fee credit made `increase` revert
+///         `DeltaNotNegative` unconditionally for any out-of-range position with fees on its empty side. The
+///         fee collect removes that; these tests additionally pin that the collected single-sided fees entering
+///         the budget are valued sanely at spot by `SwapAndAddMath.getLiquidityForAmountsWeighted` (no funds at rest, no operator
+///         redirect, stranger rejected).
+contract SwapAndAddIncreaseOutOfRangeTest is PosmTestSetup {
```
zhongeric: Remove this suite, add it as a test

test/SwapAndAddSamePoolRoute.t.sol
```diff
@@ -0,0 +1,223 @@
...
+///         `DeltaNotNegative`, bricking a route shape the contract advertises as supported. `_deployLiquidity` now
+///         closes each currency with `CLOSE_CURRENCY`, which settles a debt or takes a credit.
+///
+///         Where the credited fees END UP: taken into this contract's balance AFTER sizing, so they cannot
+///         grow the position beyond the already-fixed optimistic size. The reconcile spends them on the flash
+///         debt first (avoiding a trim), and the remainder is swept to the resolved recipient. So fees accrued
+///         DURING an operation are refunded, not compounded — measured below. Fees accrued BEFORE the
+///         operation are unaffected: the prologue collect puts them in the budget ahead of sizing, so they
+///         are still fully reinvested, which is what `compound` promises.
+contract SwapAndAddSamePoolRouteTest is PosmTestSetup {
```
zhongeric: Remove this test

src/libraries/SwapAndAddMath.sol
zhongeric: Rewritten in plain terms in #594.

src/libraries/SwapAndAddMath.sol
zhongeric: Yes, exactly: both sides are valued in whichever token is cheaper so the rate stays >= 1. Comment rewritten to say that in #594.

src/libraries/SwapAndAddMath.sol
zhongeric: Fixed to "cheaper or equal" in #594.

src/libraries/SwapAndAddMath.sol
zhongeric: Reordered to mulDiv(budgetValue, REFERENCE_LIQUIDITY, refValue) in #594 (mulDiv is commutative, no behavior change).

src/SwapAndAdd.sol
zhongeric: NatSpec rewritten in #594: max slippage is fine because callers enforce minLiquidity on the final position, and callers MUST check minimum amounts.

test/SwapAndAddFuzz.t.sol
zhongeric: Comment rewritten in #594: the caps keep the flash-take below the reserve pool's holdings so an out-of-scope PoolManager-drained take cannot fail the property.

test/SwapAndAddFuzz.t.sol
zhongeric: The bug-history narrative is removed in #594. The header now just states the fuzz domain and the property.

PR #602, opened 2026-09-17
fix(perm-pools): make unwindPosition force-exit unblockable against subscriber and delivery DoS

SocksNFlops:
Hardens `unwindPosition` (the compliance force-exit) so the LP cannot block it. Three fixes:

- **Subscriber callback**: unsubscribe with the PoolManager locked and force-clear a re-attached subscriber before burn, so `notifyUnsubscribe`/`notifyBurn` cannot strand a delta or revert the exit.
- **Delivery reentrancy**: settle both currencies as ERC-6909 claims to this contract during the burn unlock (no LP token code runs), then deliver each as the real asset in its own isolated unlock, falling back to handing over the claim.
- **Gas-griefing**: cap each delivery attempt with a new immutable `deliveryGasLimit` (mirrors `unsubscribeGasLimit`).

`MINT_6909`/`UNWIND_WITH_FALLBACK` are reachable via the public router; reviewed, no value-extraction path (both only act on this contract's own credit/claims). `test_unwindActions_onPublicRouter_cannotExtractValue` pins this.

Permissioned suite green under CI config (Foundry v1.3.6, `--isolate`, `FOUNDRY_PROFILE=ci`): 377 passed. `forge fmt --check` clean on v1.4.3.


src/hooks/permissionedPools/PermissionedPositionManager.sol
```diff
@@ -23,6 +23,10 @@ contract PermissionedPositionManager is PositionManager {
 
     IPermissionsAdapterFactory public immutable PERMISSIONS_ADAPTER_FACTORY;
 
+    /// @notice Gas forwarded to each real-asset delivery attempt in `_tryDeliverAsset`.
+    ///         Caps how much a malicious recipient can burn per attempt.
+    uint256 public immutable deliveryGasLimit;
```
gretzke: can you put `deliveryGasLimit` in caps? so it matches the `PERMISSIONS_ADAPTER_FACTORY` convention above

test/hooks/permissionedPools/UnwindGasGriefPoC.t.sol
```diff
@@ -0,0 +1,241 @@
...
+
+/// @title Gas-grief PoC — compounding gas grief on unwindPosition's per-currency delivery cascade, and its fix
+/// @dev The attacker is both the LP and the admin of their own PermissionsAdapter(s). Because `_deliverCurrency`
+///      tries `to = lp` then `to = admin` and both resolve to the SAME attacker-controlled address, BOTH
+///      `_tryDeliverAsset` attempts invoke the malicious underlying token's `transfer`. Before the fix each
+///      attempt opened an uncapped `poolManager.unlock`, so EIP-150's 63/64 forwarding let a griefed leg consume
+///      most of the remaining gas; two (or four) such legs could starve the second currency's delivery and the
+///      claim-handover fallback, bricking the force-exit unless the caller supplied ~15M+ gas.
+///
+///      After the fix `_tryDeliverAsset` forwards at most the immutable `deliveryGasLimit` per attempt, and
```
ericneil-sanc: ultra-nit: theres no _deliveryAttemptBudget in the src, I think this is from an earlier iteration. can you update this to just describe the deliveryGasLimit cap? (FWIW I checked and the guard isnt needed, couldnt get a starved-but-succeeded outcome at any gas budget)

test/hooks/permissionedPools/PermissionedPositionManager.t.sol
```diff
@@ -2549,73 +2591,36 @@ contract PermissionedPositionManagerTest is Test, PermissionedPosmTestSetup, Liq
...
-        vm.expectRevert(abi.encodeWithSelector(IPositionManager.NotApproved.selector, unauthorizedUser));
-        lpm.modifyLiquidities(calls, block.timestamp + 1);
-
-        // subscriber stays attached
-        assertEq(address(INotifier(address(lpm)).subscriber(tokenId)), address(sub));
-    }
-
-    /// @dev The position owner can use the UNSUBSCRIBE action directly.
-    function test_handleAction_unsubscribe_succeeds_for_owner() public {
+    /// @dev The unwind-only unlock actions (MINT_6909 settle-as-claim, UNWIND_WITH_FALLBACK deliver) cannot be
```
ericneil-sanc: super nit: this @dev says the actions revert Unauthorized outside the unwind flow but the @notice right below says theyre reachable, and the @notice is whats true. can you drop the @dev paragraph? (might change again depending on the L112 thread)

src/hooks/permissionedPools/PermissionedPositionManager.sol
```diff
@@ -73,30 +83,39 @@ contract PermissionedPositionManager is PositionManager {
...
         bytes[] memory params = new bytes[](3);
         params[0] = abi.encode(tokenId, amount0Min, amount1Min, hookData);
-        // PoolKey is encoded into the unwind params because BURN_POSITION clears positionInfo[tokenId].
-        params[1] = abi.encode(poolKey, poolKey.currency0, lp, tokenId);
-        params[2] = abi.encode(poolKey, poolKey.currency1, lp, tokenId);
+        params[1] = abi.encode(poolKey.currency0);
+        params[2] = abi.encode(poolKey.currency1);
         poolManager.unlock(abi.encode(actions, params));
+
+        // This contract never holds standing 6909 claims, so these balances are exactly what the burn produced.
```
ericneil-sanc: Maybe make clear we can unintentionally have 6909 claims

src/hooks/permissionedPools/PermissionedPositionManager.sol
SocksNFlops: Done, renamed to `DELIVERY_GAS_LIMIT` in d96be5e7.

test/hooks/permissionedPools/UnwindGasGriefPoC.t.sol
SocksNFlops: Fixed in d96be5e7: rewrote the block to describe just the `DELIVERY_GAS_LIMIT` per-attempt cap and dropped the `_deliveryAttemptBudget` reference. Thanks for confirming the extra gate wasn't needed.

test/hooks/permissionedPools/PermissionedPositionManager.t.sol
SocksNFlops: Fixed in d96be5e7: dropped the stale `@dev`, leaving the accurate `@notice`.

src/hooks/permissionedPools/PermissionedPositionManager.sol
SocksNFlops: Fixed in d96be5e7: the comment now notes the balance is the burn output plus any stray `MINT_6909` claims parked in this contract.

ericneil-sanc (approved):

